PR #64, opened 2015-02-06
FFMPEG_PATH can be specified by the user

Zulko:
Hi there ! I will try outsource all of MoviePy's video IO operations oh imageio. Some MoviePy users prefer to use their own FFMPEG binary instead of imageio's default. This commit gives a handle that I can use to easily change the path of the ffmpeg binary in imageio.

### DIFF
--- a/imageio/plugins/ffmpeg.py
+++ b/imageio/plugins/ffmpeg.py
@@ -35,10 +35,23 @@ FNAME_PER_PLATFORM = {
     'linux64': 'ffmpeg.linux64',
 }
 
+# This constant allows the user to use a custom FFMPEG binary instead of
+# the default binary specified by FNAME_PER_PLATFORM
+#
+# Example of use:
+# >>> import imageio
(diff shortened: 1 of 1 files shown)


almarklein:
What about using an environment variable? Seems a bit more flexible even.

Zulko:
Maybe, I'm not really familiar with environment variables, what would be the advantage ? I have no objection to any solution that would allow me to change the settings during a python session.

almarklein:
The advantage would be that users can set it also from outside the Python session, or without needing to import imageio first.

    # in plugins/ffmpeg.py
    ffmpeg_exe = os.getenv('IMAGEIO_FFMPEG_EXE')
    if ffmpeg_exe:
        return fmpeg_exe

    # Somewhere in the code:
    os.environ['IMAGEIO_FFMPEG_EXE'] = 'ffmpeg'

    # ... or at the command line:
    export IMAGEIO_FFMPEG_EXE=ffmpeg

Zulko:
Ok, that would work for me.

almarklein:
Just implemented this.

Issue #80, opened 2015-03-14
Is it possible to pass arbitrary arguments to ffmpeg with get_writer()


Zulko:
I would like to pass arguments like: bitrate, file for soundtrack, etc... and could not find an easy way, from the code I wonder if it's even possible. Thanks in advance !

almarklein:
See http://imageio.readthedocs.org/en/latest/format_ffmpeg.html#parameters-for-saving

You can set the fps, codec and bitrate, but no other parameters are "exposed" at the moment. I'd be happy to add more parameters though. PR's welcome :)

almarklein:
Maybe we just need one `ffmpeg_params` argument to specify *any* ffmpeg argument. The user would have to know about ffmpeg, but it would be very flexible.

Zulko:
Yes, that's what is done in MoviePy at the moment. If you want I can make a PR.

almarklein:
That would be great!

almarklein:
Done in #83

PR #83, opened 2015-06-06
FFMPEG writer enhancements: sane output pixel format & wmv support

### LINKED ISSUE #31: FFMPEG read camera stream on OSX
Reading from a camera stream is supported on Linux and Win, but mot yet on OSX. Mostly because I cannot test it (or at least I expect accessing my camera from a VM will be problematic).

The main changes requires are in `_get_cam_inputname` to specify how to read the camera, and in the beginning of the module to set `CAM_FORMAT`. Would be good if video4linux would just work on osx, but I doubt that it would.

cc @rreilink (if you have time, someday)

### LINKED ISSUE #79: ffmpeg leaves processes behind
Not sure if this is just me doping something stupid, but here goes:

    for f in files:
             try:
                 vol = imageio.mimread(f)
             except Exception as e:
                 print f

This leaves me with this kind of process list:

leftover_processes

As you can see, this is just a stupid loop to check I can actually load every file in my dataset, HMDB51 btw. Once this loop finishes (with the first error) the process list turns into this:

zombies

Now the error:

    dataset/files/hmdb51_org/videos/scaled_256/stand/RETURN_OF_THE_KING_stand_u_cm_np1_fr_med_12.avi [Errno 24] Too many open files

looking at the ffmpeg plugin there *is* a `_terminate` function.
It sends a SIGTERM to the process and then waits for one second for it to terminate.
If I add a SIGKILL after the busy waiting the method looks like this:

    def _terminate(self, timeout=1.0):
        """ Terminate the sub process.
        """
        # Check
        if self._proc is None:  # pragma: no cover
            return  # no process
        if self._proc.poll() is not None:
            return  # process already dead
        # Terminate process
        self._proc.terminate()
        # Wait for it to close (but do not get stuck)
        etime = time.time() + timeout
        while time.time() < etime:
            time.sleep(0.01)
            if self._proc.poll() is not None:
                break
        if self._proc.poll() is None:
            print('killing process')
            self._proc.kill()

And the kill seems to be called every time.
Any idea what's happening?
I am on Fedora 21 btw.

    ~/.imageio/ffmpeg/ffmpeg.linux64 -version
    ffmpeg version 2.4.2-   http://johnvansickle.com/ffmpeg/    Copyright (c) 2000-2014 the FFmpeg developers
    built on Oct  9 2014 07:24:56 with gcc 4.8 (Debian 4.8.3-11)
    configuration: --enable-gpl --enable-version3 --disable-shared --disable-debug --enable-runtime-cpudetect --enable-libmp3lame --enable-libx264 --enable-libx265 --enable-libwebp --enable-libspeex --enable-libvorbis --enable-libvpx --enable-libfreetype --enable-fontconfig --enable-libxvid --enable-libopencore-amrnb --enable-libopencore-amrwb --enable-libtheora --enable-libvo-aacenc --enable-libvo-amrwbenc --enable-gray --enable-libopenjpeg --enable-libopus --disable-ffserver --enable-libass --enable-gnutls --cc=gcc-4.8
    libavutil      54.  7.100 / 54.  7.100
    libavcodec     56.  1.100 / 56.  1.100
    libavformat    56.  4.101 / 56.  4.101
    libavdevice    56.  0.100 / 56.  0.100
    libavfilter     5.  1.100 /  5.  1.100
    libswscale      3.  0.100 /  3.  0.100
    libswresample   1.  1.100 /  1.  1.100
    libpostproc    53.  0.100 / 53.  0.100

### LINKED ISSUE #84: Images to video - Won't play back. Freezes after 725 images.
I've asked about the issue on stackoverflow, the link is here: http://stackoverflow.com/questions/30583873/ffmpeg-image-to-video-images-from-ram-python

INFO: Running Python 2.7.9/10 on Windows 7. Imageio downloaded from pip.
These are the two main problems
1) After creating an .mp4 file from several images(.png), it wont play back. The file does not have a thumbnail either. I've tried with both windows player, and vlc player.
2) My program looks for a path in a custom maze (explained in link), and makes an image for every step it makes. It creates 4840 images, and on the first prototype, I saved every image with PIL in a folder, and then ran FFMPEG commands from the command-line. This did the job, but when I try to add images on the go (from RAM) with "writer.append_data(image)", it just stops at 725 images. I confirmed this by adding "print loopCount" to my program. Nothing happens, no crash, just full stop.

kuchi:
I added the following:

1) pixelformat option to writer: Lets you specify an output video pixel format. Now defaults to a sane  format to be compatible with most players. Otherwise FFMPEG often uses pixel formats that are not playable in common players. Fixes #84 issue with videos not playable in some players.
2) .wmv file support & a sane default codec since h264 is not always available on windows.
3) ffmpeg_log_level option to writer: does not redirect stderr so you can see what is going on for debugging. Sets default level to warning so there is minimal output.
4) ffmpeg_params option to writer: lets you pass additional arguments to ffmpeg if needed. Addresses #80.
5) Fixes #84 where stderr was getting full on windows and hanging.
6) May fix #79: _close function updated
7) Fixes #31, however static build of ffmpeg needs to be updated current one not compiled with camera support on os x. New version http://evermeet.cx/ffmpeg/ffmpeg-2.6.3.7z does work.
~~8) Made find_exe use ffmpeg if available on system path before downloading ffmpeg.~~
9) Added quality optional argument. Seems better than using constant bitrate. This argument ranges 0-10 with 10 being best quality. Then it sets the variable bit rate quality factor based on the codec. This makes for nicer looking videos that are smaller. Seems like a win overall and more conceptual to use. I left the bitrate argument for compatibility but this one is used by default.

I had a test failing for FFMPEG for an image difference. I had to sightly increase a threshold. Maybe due to my platform (OS X) or FFMPEG version??

Thanks for the excellent work on imageio, it's awesome and a great library.

### DIFF
--- a/imageio/plugins/ffmpeg.py
+++ b/imageio/plugins/ffmpeg.py
@@ -528,7 +528,7 @@ class FfmpegFormat(Format):
             # Write
             try:
                 self._proc.stdin.write(im.tostring())
-            except IOError, e:
+            except IOError as e:
                 # Show the command and stderr from pipe
                 msg = '{}\n\nFFMPEG COMMAND:\n{}\n\nFFMPEG STDERR OUTPUT:\n'\
                     .format(e, self._cmd)
--- a/imageio/plugins/ffmpeg.py
+++ b/imageio/plugins/ffmpeg.py
@@ -120,6 +120,10 @@ class FfmpegFormat(Format):
     pixelformat: str
         The output video pixel format. Default is 'yuv420p' which most widely
         supported by video players.
+    ffmpeg_args: list
+        List additional arguments to ffmpeg for output file options.
+        Example ffmpeg arguments to use only intra frames and set aspect ratio:
+        ['-intra', '-aspect', '16:9']
     """
     
     def _can_read(self, request):
@@ -467,7 +471,8 @@ class FfmpegFormat(Format):
     
     class Writer(Format.Writer):
         
-        def _open(self, fps=10, codec='libx264', bitrate=400000):
+        def _open(self, fps=10, codec='libx264', bitrate=400000,
+                  pixelformat='yuv420p', ffmpeg_args=None):
             self._exe = get_exe()
             # Get local filename
             self._filename = self.request.get_local_filename()
@@ -476,6 +481,7 @@ class FfmpegFormat(Format):
             # Initialize parameters
             self._proc = None
             self._size = None
+            self._cmd = None
         
         def _close(self):
             # Close subprocess
@@ -514,7 +520,15 @@ class FfmpegFormat(Format):
             assert self._proc is not None  # Check status
             
             # Write
-            self._proc.stdin.write(im.tostring())
+            try:
+                self._proc.stdin.write(im.tostring())
+            except IOError, e:
+                # Show the command and stderr from pipe
+                stdout =  self._stderr_catcher.get_text(0.1)
+                msg = '{}\n\nFFMPEG COMMAND:\n{}\n\nFFMPEG STDERR OUTPUT:\n{}'\
+                    .format(e, self._cmd, stdout)
+                raise IOError(msg)
+
         
         def set_meta_data(self, meta):
             raise RuntimeError('The ffmpeg format does not support setting '
@@ -536,6 +550,7 @@ class FfmpegFormat(Format):
                 default_codec = 'msmpeg4'
             codec = self.request.kwargs.get('codec', default_codec)
             bitrate = self.request.kwargs.get('bitrate', 400000)
+            extra_ffmpeg_args = self.request.kwargs.get('ffmpeg_args', [])
             # You may need to use -pix_fmt yuv420p for your output to work in
             # QuickTime and most other players. These players only supports
             # the YUV planar color space with 4:2:0 chroma subsampling for
@@ -556,12 +571,16 @@ class FfmpegFormat(Format):
                    '-vcodec', codec,
                    '-pix_fmt', pixelformat,
                    ]
-            cmd += ['-b', str(bitrate)] if (bitrate is not None) else [] 
-            cmd += ['-r', "%d" % fps, self._filename]
+            cmd += ['-b:v', str(bitrate)] if (bitrate is not None) else []
+            cmd += ['-r', "%d" % fps]
+            cmd += extra_ffmpeg_args
+            cmd.append(self._filename)
+            self._cmd = " ".join(cmd) # For showing command if needed
             
             # Launch process
             self._proc = sp.Popen(cmd, stdin=sp.PIPE,
                                   stdout=sp.PIPE, stderr=sp.PIPE)
+            self._stderr_catcher = StreamCatcher(self._proc.stderr)
 
 
 def cvsecs(*args):
--- a/imageio/plugins/ffmpeg.py
+++ b/imageio/plugins/ffmpeg.py
@@ -19,6 +19,7 @@ import re
 import time
 import threading
 import subprocess as sp
+import warnings
 
 import numpy as np
 
@@ -523,14 +524,7 @@ class FfmpegFormat(Format):
                 return  # process already dead
             if self._proc.stdin:
                 self._proc.stdin.close()
-            # Wait for process to terminate or force if it doesn't in time.
-            waited = 0.0
-            while self._proc.poll() is not None:
-                time.sleep(0.1)
-                waited += 0.1
-                if waited > timeout:
-                    self._proc.terminate()
-                    break
+            self._proc.wait()
             self._proc = None
 
         def _append_data(self, im, meta):
--- a/imageio/plugins/ffmpeg.py
+++ b/imageio/plugins/ffmpeg.py
@@ -516,7 +516,7 @@ class FfmpegFormat(Format):
             self._size = None
             self._cmd = None
 
-        def _close(self, timeout=10.0):
+        def _close(self):
             if self._proc is None:  # pragma: no cover
                 return  # no process
             if self._proc.poll() is not None:
--- a/imageio/plugins/ffmpeg.py
+++ b/imageio/plugins/ffmpeg.py
@@ -492,7 +492,8 @@ class FfmpegFormat(Format):
             # Close subprocess
             if self._proc is not None:
                 self._proc.stdin.close()
-                self._proc.stderr.close()
+                if not self._verbose:
+                    self._proc.stderr.close()
                 self._proc.wait()
                 self._proc = None
         
--- a/imageio/plugins/ffmpeg.py
+++ b/imageio/plugins/ffmpeg.py
@@ -582,7 +582,7 @@ class FfmpegFormat(Format):
             cmd += ['-r', "%d" % fps]
             cmd += extra_ffmpeg_args
             cmd.append(self._filename)
-            self._cmd = " ".join(cmd) # For showing command if needed
+            self._cmd = " ".join(cmd)  # For showing command if needed
 
             stderr = sp.PIPE
             if self._verbose:
--- a/imageio/plugins/ffmpeg.py
+++ b/imageio/plugins/ffmpeg.py
@@ -198,11 +198,7 @@ class FfmpegFormat(Format):
                 # Appears that newer ffmpeg builds don't support -list-devices
                 # on OS X. But you can directly open the camera by index.
                 name = str(index)
-
-                if sys.platform.startswith('darwin'):
-                    return name
-                else:
-                    return 'video=%s' % name
+                return name
 
             else:  # pragma: no cover
                 return '??'
--- a/imageio/plugins/ffmpeg.py
+++ b/imageio/plugins/ffmpeg.py
@@ -67,7 +67,7 @@ if sys.platform.startswith('win'):
 elif sys.platform.startswith('linux'):
     CAM_FORMAT = 'video4linux2'
 elif sys.platform.startswith('darwin'):
-    CAM_FORMAT = '??'
+    CAM_FORMAT = 'avfoundation'
 else:  # pragma: no cover
     CAM_FORMAT = 'unknown-cam-format'
 
@@ -187,6 +187,38 @@ class FfmpegFormat(Format):
                 except IndexError:
                     raise IndexError('No ffdshow camera at index %i.' % index)
                 return 'video=%s' % name
+
+            elif sys.platform.startswith('darwin'):
+                # Ask ffmpeg for list of dshow device names
+                cmd = [self._exe, '-list_devices', 'true',
+                       '-f', CAM_FORMAT, '-i', 'dummy']
+                proc = sp.Popen(cmd, stdin=sp.PIPE, stdout=sp.PIPE,
+                                stderr=sp.PIPE)
+                proc.stdout.readline()
+                proc.terminate()
+                infos = proc.stderr.read().decode('utf-8')
+                # Parse the result
+                device_names = []
+                for line in infos.splitlines():
+                    if "input device" in line:
+                        print(line)
+                        line = line.split(']', 1)[1].strip()
+                        print (line)
+                        if line.startswith('['):
+                                line = line.split(']', 1)[1].strip()
+                                device_names.append(line)
+                # Return device name at index
+                try:
+                    name = device_names[index]
+                except IndexError:
+                    raise IndexError('No avfoundation camera at index %i.'
+                                     % index)
+
+                if sys.platform.startswith('darwin'):
+                    return name
+                else:
+                    return 'video=%s' % name
+
             else:  # pragma: no cover
                 return '??'
         
@@ -296,6 +328,7 @@ class FfmpegFormat(Format):
             cmd = [self._exe] + iargs + ['-i', self._filename] + oargs + ['-']
             self._proc = sp.Popen(cmd, stdin=sp.PIPE,
                                   stdout=sp.PIPE, stderr=sp.PIPE)
+            print(cmd)
             # Create thread that keeps reading from stderr
             self._stderr_catcher = StreamCatcher(self._proc.stderr)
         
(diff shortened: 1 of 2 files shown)


almarklein:
Looks good to me. All test pass, except for style: there are a few lines longer than 80 chars.

kuchi:
@almarklein you should review how I changed get_exe to make sure you are ok with what I did. Now it looks for an ffmpeg on the path before downloading or using the downloaded ffmpeg. Do you think that makes sense or could it cause some unforeseen problems?

imageio/plugins/ffmpeg.py
```diff
@@ -174,10 +199,21 @@ def _get_cam_inputname(self, index):
...
                     raise IndexError('No ffdshow camera at index %i.' % index)
                 return 'video=%s' % name
+
+            elif sys.platform.startswith('darwin'):
+                # Appears that newer ffmpeg builds don't support -list-devices
+                # on OS X. But you can directly open the camera by index.
+                name = str(index)
+
+                if sys.platform.startswith('darwin'):
+                    return name
```
almarklein: This is always True, isnt it?

almarklein:
Well, as I recall, the standard ffmpeg can be rather outdated (e.g. on older or more conservative Linux flavours), which means some features (like video) would not be available ...

almarklein:
Wait, if FFMPEG_FFMPEG_EXE is "ffmpeg", would it not just work? Why the need for getting the full path? 

Also cc-ing @Zulko.

Zulko:
I think it makes more sense to use the imageio ffmpeg by default. This way every user has the same behavior by default. As @almarklein noted the ffmpeg in the path is deprecated in systems like Ubuntu.

imageio/plugins/ffmpeg.py
kuchi: Thanks for catching that. I fixed it.

kuchi:
@almarklein, what are your thoughts on the quality argument I added to the writer. I think this is a better way to set the video output quality rather than a constant bit rate.

kuchi:
@Zulko, your moviepy ffmpeg writer is pretty similar to the imageio ffmpeg writer. Since you are already using imageio in moviepy, would it make sense if the enhancements to your writer could be merged into the imageio writer? Then you could import the imageio ffmpeg writer and use that. That way both codes could have one really good ffmpeg writer. Just a thought.

imageio/plugins/ffmpeg.py
```diff
@@ -534,13 +614,41 @@ def _initialize(self):
...
-                   '-vcodec', codec] 
-            cmd += ['-b', str(bitrate)] if (bitrate is not None) else [] 
-            cmd += ['-r', "%d" % fps, self._filename]
-            
+                   '-vcodec', codec,
+                   '-pix_fmt', pixelformat,
+                   ]
+            if bitrate is not None:
+                cmd += ['-b:v', str(bitrate)]
+            if quality >= 0:  # If < 0, then we don't add anything
```
almarklein: @kuchi I like the quality argument. But I would ditch the -1 option, and simply use bitrate instead if its given. In other words, make this an `else`.

kuchi: I put the -1 option as a way to remove all variable bit rate and fixed bit rate params to ffmpeg. Some codecs require different parameters for compression so you don't want those. This way if someone uses a different codec that needs a different compression parameter they can manually specify it with ffmpeg_params without getting a conflict. 

But maybe there is a better way to do this that would be more straight forward? Any suggestions?

kuchi: I do see an error in the above code. It should be 

    elif quality >= 0:

imageio/plugins/ffmpeg.py
```diff
@@ -112,45 +114,63 @@ class FfmpegFormat(Format):
         The number of frames per second. Default 10.
     codec : str
         the video codec to use. Default 'libx264', which represents the
-        widely available mpeg4.
+        widely available mpeg4. Except when saving .wmv files, then the
+        defaults is 'msmpeg4' which is more commonly supported for windows
+    quality : float
+        Video output quality. Default is 5. Uses variable bit rate. Highest
+        quality is 10, lowest is 0. Set to -1 to prevent quality flags to
+        FFMPEG so you can manually specify them using ffmpeg_params instead.
```
almarklein: ... or use bitrate

almarklein:
I see the point for having -1. We should decide what happens when bitrate is given. Should it be used if not None, or only be used when quality is -1? The former is more backward compatible, the latter is more consistent with setting other quality params via ffmpeg_params. I think I lean towards the latter.

Also, maybe instead of -1,  setting it to None should just turn it off?

kuchi:
I like your suggestion of using None for quality to disable it. I'll change it.
I have it so that if bitrate is given, it overrides quality and uses a fixed bit rate. That seems backwards compatible and intuitive for the user, if they ask for it they get it. They shouldn't specify both bitrate and quality really, but in that case bitrate takes precedence.

kuchi:
@almarklein, I think all your points have been addressed so far. Let me know if you have any other concerns or thoughts on this pull request. Otherwise I think I have everything working that I wanted to add. Apologies that the pull request went a bit beyond the initial scope. I really just wanted to get the pixelformat to be playable in most players. But then I saw other stuff that could be addressed.

Thanks again for your excellent imageio package! I was rolling my own ffmpeg writer for work I needed to do but when I discovered imagio had a writer I decided it would be more fruitful to use yours and contribute to it.

Zulko:
@kuchi Yes, improving imageio's ffmpeg reader/writer and having moviepy use these was the original intention, I just didn't have time to make that happen. I noticed that for some reason the imageio writer was slower than moviepy's, I don't know if it's still true, I'll look into this when I have time (might be in a month).

imageio/plugins/ffmpeg.py
```diff
@@ -112,45 +114,64 @@ class FfmpegFormat(Format):
...
+        defaults is 'msmpeg4' which is more commonly supported for windows
+    quality : float
+        Video output quality. Default is 5. Uses variable bit rate. Highest
+        quality is 10, lowest is 0. Set to None to prevent variable bitrate
+        flags to FFMPEG so you can manually specify them using ffmpeg_params
+        instead.
     bitrate : int
-        A measure for quality. Default 400000
+        Set a constant bitrate for the video encoding. By default 'quality'
+        is used instead.
```
almarklein: Maybe be a bit more precise here:
* `quality : float | None`
* `bitrate : int | None` and say that the default is None
* Mention that if bitrate is given, it overrides quality.

imageio/plugins/ffmpeg.py
```diff
@@ -471,24 +510,41 @@ def _open(self, fps=10, codec='libx264', bitrate=400000):
...
+            if self._proc is None:  # pragma: no cover
+                return  # no process
+            if self._proc.poll() is not None:
+                return  # process already dead
+            # Terminate process
+            self._proc.stdin.close()
+            self._proc.wait()
+            if self._proc.poll() is not None:
+                return  # process already dead
+            self._proc.terminate()
```
almarklein: I think `proc.wait()` would wait for it to stop, so anything that comes below is not necessary... or maybe we should not call `proc.wait()` at all?

almarklein:
Other than the new comments, this looks good. Thank you for the many improvements! This kind of collaboration --- people working together to create *one* good ffmpeg/png/etc. reader --- is exactly what I was hoping for...

imageio/plugins/ffmpeg.py
kuchi: I updated this. The python docs actually suggest to use .communicate to close streams and not to use .wait() which can deadlock. So this is cleaned up.

imageio/plugins/ffmpeg.py
kuchi: Done.

kuchi:
@Zulko I did a speed test using the below code. Using this branch of Imageio and your stable moviepy both writers are about the same speed. The Imageio is a 10% faster but that is probably due to the difference in default quality of the variable bitrate between the two libraries.

Here are the results:
````
Encoding 5 videos with 100 frames, each (512, 512, 3) pixels
imageio encoding time 23.549 sec
moviepy encoding time 25.7 sec
moviepy encoding time / imageio encoding time: 1.1
````

Here is the code I used:
```python
import numpy as np
import imageio
import moviepy.video.io.ffmpeg_writer as moviepy_ffmpeg
import time

img = imageio.imread("lena.png")
movie_size = img.shape

n_movies = 5
n_frames = 100
fps=30

print "Encoding {} videos with {} frames, each {} pixels".format(n_movies, n_frames, movie_size)
def add_frames(frame_writer_func):
    # Fade in image
    for index in range(n_frames):
        frame_writer_func((img*(1.0*index/n_frames)).astype(np.uint8))
    # Frames with noise
    for index in range(n_frames):
        ar = np.random.randint(0, 255, movie_size).astype(np.uint8)
        frame_writer_func(ar)

start = time.time()
for i in range(n_movies):
    writer = imageio.get_writer('imageio_video.mp4', fps=fps)
    add_frames(writer.append_data)
    writer.close()
imageio_dur = time.time() - start
print "imageio encoding time {:.5} sec".format(imageio_dur)

start = time.time()
for i in range(n_movies):
    writer = moviepy_ffmpeg.FFMPEG_VideoWriter('moviepy_video.mp4', movie_size[:2], fps)
    add_frames(writer.write_frame)
    writer.close()
moviepy_dur = time.time() - start
print "moviepy encoding time {:.5} sec".format(moviepy_dur)
print "moviepy encoding time / imageio encoding time: {:.2}".format(moviepy_dur/imageio_dur)
```

imageio/plugins/ffmpeg.py
```diff
@@ -471,24 +514,27 @@ def _open(self, fps=10, codec='libx264', bitrate=400000):
...
+
+        def _close(self, timeout=1.0):
+            if self._proc is None:  # pragma: no cover
+                return  # no process
+            if self._proc.poll() is not None:
+                return  # process already dead
+            # End process
+            # self._proc.stdin.close() # .communicate will take care of this
+            # The .communicate is preferred over .wait which can deadlock
+            self._proc.communicate()
```
almarklein: I don't think this is right. Are you sure it closes stdin (it does not say so in the docs)? If it does, writing should work, I suppose, but for *reading*, we really need to ask ffmpeg to terminate. I'm also slightly worried that if ffmpeg hangs, this hangs as well, but we can try and see if that's really a problem...

kuchi: @almarklein Looking at the source (shown below) it does appear to close the streams. 

```python
   def communicate(self, input=None):
        """Interact with process: Send data to stdin.  Read data from
        stdout and stderr, until end-of-file is reached.  Wait for
        process to terminate.  The optional input argument should be a
        string to be sent to the child process, or None, if no data
        should be sent to the child.

        communicate() returns a tuple (stdout, stderr)."""

        # Optimization: If we are only using one pipe, or no pipe at
        # all, using select() or threads is unnecessary.
        if [self.stdin, self.stdout, self.stderr].count(None) >= 2:
            stdout = None
            stderr = None
            if self.stdin:
                if input:
                    try:
                        self.stdin.write(input)
                    except IOError as e:
                        if e.errno != errno.EPIPE and e.errno != errno.EINVAL:
                            raise
                self.stdin.close()
            elif self.stdout:
                stdout = _eintr_retry_call(self.stdout.read)
                self.stdout.close()
            elif self.stderr:
                stderr = _eintr_retry_call(self.stderr.read)
                self.stderr.close()
            self.wait()
            return (stdout, stderr)

        return self._communicate(input)
```

kuchi: I guess in looking at what the above .communicate source does, we can essentially go back to stdin.close() and .wait(). Looks no different. Maybe it's clearer that way.

imageio/plugins/ffmpeg.py
```diff
@@ -471,24 +514,26 @@ def _open(self, fps=10, codec='libx264', bitrate=400000):
...
+                return  # no process
+            if self._proc.poll() is not None:
+                return  # process already dead
+            if self._proc.stdin:
                 self._proc.stdin.close()
-                self._proc.stderr.close()
-                self._proc.wait()
-                self._proc = None
-        
+            self._proc.wait()
```
almarklein: dont forget setting `_proc` to None. I suspect we should also use `_proc.terminate()`. At some point this method had some code to wait for the process and kill it eventually. Was that original code or did you add (and later remove) that?

kuchi: I had added the extra stuff to wait and kill but you wade a comment above that you thought the stuff after wait was not needed. I'll add the _proc None at the end and check if we need to terminate.

almarklein: Ah, maybe it should have been the other way around, not calling wait(), but do the "manual" waiting and eventual killing... I'm feeling a bit uncomfortable with the wait; what if it waits forever because ffmpeg refuses to quit?

kuchi: I did try that, but I was getting some failing tests where the video file was not closed correctly. See commit 080c1e2 below and the Travis CI tests that fail. I even increased the wait time before terminating. If you want to guarantee the user gets a good output video I think the wait is better.

almarklein: Fair enough, let's try it. If you add the `_proc = None`, this should be good to go.

imageio/plugins/ffmpeg.py
```diff
@@ -471,24 +514,27 @@ def _open(self, fps=10, codec='libx264', bitrate=400000):
...
+            self._proc = None
```
kuchi: The _proc = None is there.

almarklein: Ah, I did not notice it there :)

almarklein:
Thanks for this!

kuchi:
Thanks @almarklein for all your awesome work. imageio is exceptional. You are very talented.

Zulko:
@kuchi @almarklein But the module would need more advertisement I think. At some point when you are happy with the result it would be worth blogging about it or posting a link to reddit/python for instance.

almarklein:
@Zulko You're right, let's keep that in mind for the next release

PR #85, opened 2015-06-12
Use newly uploaded osx ffmpeg exe

almarklein:
I uploaded a new ffmpeg exe, which should make cameras work on OSX. See #31.

@kuchi do you have time to verify that this now works?

### DIFF
--- a/imageio/plugins/ffmpeg.py
+++ b/imageio/plugins/ffmpeg.py
@@ -27,8 +27,8 @@ from ..core import (Format, get_remote_file, string_types, read_n_bytes,
                     image_as_uint8, get_platform)
 
 FNAME_PER_PLATFORM = {
-    'osx32': 'ffmpeg.osx.snowleopardandabove',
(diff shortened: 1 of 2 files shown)


kuchi:
It works!

However, the test_webcam in test_ffmpeg.py needs to be updated to something like below. You currently check <video2> but on OS X the first camera is at <video0> so unless the person has 3 cameras this test will fail. Using the below code the test passes on my system.

```python
def test_webcam():
    need_internet()

    def check_cam(index):
        try:
            imageio.read('<video{0}>'.format(index))
        except Exception:
            return False
        return True

    # See if there is a camera on the first 3 indexes
    for i in range(3):
        if check_cam(i):
            # Found a camera, it worked.
            return

    # Didn't find a camera
    skip('no web cam')
```

almarklein:
Ah i think changed that at some point to sip the test on purpose.. not sure why. I'm setting it back to 0. Thanks for the catch!

PR #96, opened 2015-07-03
ffmpeg compatibility enhancement: macro_block_size check against frame size

kuchi:
- improved ffmpeg_writer compatibility by adding the macro_block_size option. This ensures video frame sizes are divisible by a standard block size that most codecs and video players expect. This prevents imageio from making videos that are not playable. Also has ability to turn off if needed. But this will hopefully help people get videos that work from arbitrary numpy array sizes without having the frustration of generating a video that won't play in many players or ffmpeg will fail to encode because the codec does not support odd frame sizes.
- updated tests to check new functionality.


kuchi:
cc @Zulko - this is an issue I have run into with videos being written with moviepy as well.

Zulko:
I'm not really for enforcing this by default, but it's true that this is a frustrating problem. I have made a special fx in moviepy called "even_size" which will remove one line or column of pixels if necessary to make the clip's dimensions even:

    master_video.resize(height=400).fx( mpy.vfx.even_size)

kuchi:
What would the reasoning be for having it off by default?  Seems like it would be better to have it on by default and warn the user so they don't have a failure or video that won't play. But perhaps I haven't thought of all cases. 

Anthony

Zulko:
Ah I didn't understand it was just a warning, I thought it raised an exception. I am not against a warning by default. Just looked at your code, I think you should issue warnings using the warning module, this way they appear read in IPython and a user can decide to silence them with warning.ignore() (or something like this, can't remember).

imageio/plugins/ffmpeg.py
```diff
@@ -626,6 +636,32 @@ def _initialize(self):
...
+                    (self._size[1] % macro_block_size)
+                out_h = self._size[0] + macro_block_size - \
+                    (self._size[0] % macro_block_size)
+                cmd += ['-vf', 'scale={0}:{1}'.format(out_w, out_h)]
+                print("IMAGEIO FFMPEG_WRITER WARNING: input image is not "
+                      "divisible by macro_block_size={0}, resizing from {1} "
+                      "to {2} to ensure video compatibility with most codecs "
+                      "and players. To prevent resizing, make your input "
+                      "image divisible by the macro_block_size or set the "
+                      "macro_block_size to 0 (risking compatibility) You "
```
almarklein: risking INcompatibility?

imageio/plugins/ffmpeg.py
```diff
@@ -626,6 +636,32 @@ def _initialize(self):
...
+            macro_block_size = self.request.kwargs.get('macro_block_size', 16)
+            if macro_block_size > 1 and \
+                    (self._size[1] % macro_block_size > 0 or
+                        self._size[0] % macro_block_size > 0):
+                out_w = self._size[1] + macro_block_size - \
+                    (self._size[1] % macro_block_size)
+                out_h = self._size[0] + macro_block_size - \
+                    (self._size[0] % macro_block_size)
+                cmd += ['-vf', 'scale={0}:{1}'.format(out_w, out_h)]
+                print("IMAGEIO FFMPEG_WRITER WARNING: input image is not "
```
almarklein: logging.warning would be better indeed

almarklein: Or `warnings.warn`

imageio/plugins/ffmpeg.py
```diff
@@ -140,6 +140,15 @@ class FfmpegFormat(Format):
         Values can be "quiet", "panic", "fatal", "error", "warning", "info"
         "verbose", or "debug". Also prints the FFMPEG command being used by
         imageio if not "quiet" or "warning".
+    macro_block_size: int
+        Size constraint for video width and height, must be divisible by this
+        number. If not divisible by this number imageio will tell ffmpeg to
```
almarklein: This first sentence seems a bit unclear. Do you mean "Size constraint for video. Width and height, must be divisible by this number."?

almarklein: Also mention the default.

almarklein:
What about when None is specified it uses 16 and gives the warning. If 0 or 1 is specified it does not do any scaling, and when 16 is specified, it uses 16 without the warning.

imageio/plugins/ffmpeg.py
kuchi: It has the default of 16 in the description.

imageio/plugins/ffmpeg.py
kuchi: Ok, I'll change it to warnings.warn. I like logging better but if no logger is setup by the user they won't see the warning.

imageio/plugins/ffmpeg.py
almarklein: Oh, it does. Sorry for the noise :)

kuchi:
I'm changing the description to state that you should set it to None to turn it off. But mathematically, if its 0 or 1 it turns off anyway. You always get the warning if it does any automatic scaling so you know whats going on.

imageio/plugins/ffmpeg.py
almarklein: Actually it is, because there always is a default logger.

almarklein:
No I meant that None, meant not-specified, in which case 16 would be used + a warning. And if a value is specified there would not be a warning. Not sure if its a good idea though, but I can imagine that you might want to suppress the warning. Most users just want pretty and working video files and don't care about the exact size of the output.

imageio/plugins/ffmpeg.py
kuchi: But I tried it in my python and no message showed up? It said not default logger configured and didn't print the warning... I'm not sure why. Usually I have to configure a logger to see messages I thought. But maybe something is weird.

kuchi:
My thought is it's always good to have the warning, as it is you always get a warning from ffmpeg as well about a speed loss due to the input image not being sized to match the macro block size (although they don't tell you that). At least this explains to the user why they are getting the message so they are not confused.

almarklein:
Fair enough :)

imageio/plugins/ffmpeg.py
almarklein: weird. It works for me. But let's play safe and use the warnings module then :)

kuchi: Figured it out. I was using a logger.warning via logger = logging.getlogger(__name__), and that then needs to be configured. But if I just do logging.warning it works. So I'll go with that.

kuchi:
Thnks for the helpful suggestions @almarklein and @Zulko .

almarklein:
@kuchi  is this ready?

kuchi:
Yes. Looks good to me as long as you are happy with it. 

Anthony

almarklein:
Thanks!

PR #100, opened 2015-07-06
Swf delay import

blink1073:
Prevent `_swf.py` from loading unless needed.

### DIFF
--- a/imageio/plugins/swf.py
+++ b/imageio/plugins/swf.py
@@ -20,7 +20,7 @@ from ..core import Format, read_n_bytes, image_as_uint8
 _swf = None  # lazily loaded in lib()
 
 
-def lib():
+def load_lib():
     global _swf
     from . import _swf
     return _swf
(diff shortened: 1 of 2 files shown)


imageio/plugins/swf.py
```diff
@@ -214,6 +220,8 @@ def _get_meta_data(self, index):
     class Writer(Format.Writer):
         
         def _open(self, fps=12, loop=True, html=False, compress=False): 
+            if not swf:
```
almarklein: `_swf`

almarklein:
Can we use either do:
* use `lib()` everywhere; or
* do a check in `open` whether `_swf` exists and call `_load_lib()` if necessary. Use `_swf` everywhere else.

I'm leaning towards the latter. We should make the behavior consisten for dicom, swf and tiffile.

blink1073:
How does that look?

almarklein:
Great!  Do you want to do the same for dicom? If it suits you can can make the change in this PR...

blink1073:
Let's keep them separate.

almarklein:
Ok, thanks!

Issue #33, opened 2014-11-07
Keep imageio light while scaling the number of plugins


almarklein:
I think we should provide much of the IO implementation in one module (e.g. `_swf.py`), while the format itself is defined in another (e.g. `swf.py`). The former should *not* be imported by default, but only as the format is being used.

blink1073:
How about deferring the actual import to within the `_open` methods where possible?  This will work for `tifffile` and `swf`.

blink1073:
See #97 for implementation.

almarklein:
Yes, that makes sense. Setup.py won't need changing, but people who freeze an app that makes use of imageio should include these `_tiffile.py` and the likes. Maybe we can provide some functionality to help such users.

blink1073:
I have no experience with freezing, and I'm on OSX.

blink1073:
Would you like a PR that defers loading of `_swf.py`?

blink1073:
I got `cx_freeze` working and verified that `mimread` and `mimsave` work with a mult-page tif file.

almarklein:
But I suppose that you had to add `_tiffile.py` or `tiffile.py` (the 5k one) to the includes?

blink1073:
The only change I made to `imageiotest_freeze.py` was to remove this [line](https://github.com/imageio/imageio/blob/master/tests/freezing/imageiotest_freeze.py#L63), because that function does not exist in [freeze.py](https://github.com/imageio/imageio/blob/master/imageio/freeze.py).

blink1073:
I assume that function *should* have been needed, in which case it looks like the frozen folder is not truly isolated.

blink1073:
I also noticed that when the app is frozen it cannot pull files from github due to an https error with `urlparse`.  I think freezing is an antipattern for many reasons.

almarklein:
>  I think freezing is an antipattern for many reasons.

haha, very True. Ah well, we'll see what happens. 

> Would you like a PR that defers loading of _swf.py?

Forgot to answer that: Sure, since its already split, it should be easy. To do the same for e.g. DICOM would need some refactoring; we can safe that for later.

blink1073:
Sure, I'll take a stab at both PRs. Are you going to the SciPy conference by chance?



—

almarklein:
Nope, I'm going to Euroscipy in August.

blink1073:
Ah well, perhaps Google Hangout sometime to discuss strategy?



—

[...]

PR #101, opened 2015-07-06
Dicom refactor

blink1073:
Split core dicom functionality into `_dicom.py` and delay import until needed.


almarklein:
Awesome, thanks! This helps towards #33

PR #103, opened 2015-07-06
Fits format plugin

blink1073:
We need to add the FITS files from 
https://github.com/scikit-image/scikit-image/tree/master/skimage/data to
https://github.com/imageio/imageio-binaries/tree/master/images

to facilitate testing.  Shall I submit a PR for that?

### DIFF
--- a/imageio/plugins/fits.py
+++ b/imageio/plugins/fits.py
@@ -24,7 +24,7 @@ class FitsFormat(Format):
     """ Flexible Image Transport System (FITS) is an open standard defining a
     digital file format useful for storage, transmission and processing of
     scientific and other images. FITS is the most commonly used digital
-    file format in astronomy.
+    file format in astronomy.  Requires the ``astropy`` package.
 
     Parameters for reading
     ----------------------
--- a/imageio/plugins/fits.py
+++ b/imageio/plugins/fits.py
@@ -81,7 +81,7 @@ class FitsFormat(Format):
         def _open(self, cache=False, **kwargs):
             if not _fits:
                 load_lib()
-            hdulist = _fits.open(self.get_open_file(),
+            hdulist = _fits.open(self.request.get_file(),
                                  cache=cache, **kwargs)
 
             self._index = []
--- a/imageio/plugins/fits.py
+++ b/imageio/plugins/fits.py
@@ -1,0 +1,119 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2015, imageio contributors
+# imageio is distributed under the terms of the (new) BSD License.
+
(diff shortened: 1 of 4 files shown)


blink1073:
While I'm adding files, it would be nice to add the TIFF files for the same reason.

almarklein:
You mean to be able to test the tiff plugin?

blink1073:
We have more extensive tests in skimage that rely on those files.

almarklein:
I've given you commit rights, so you can more easily push files to that repo. (Naturally, all changes to imageio/imageio should go via PR).

blink1073:
mwahaha!

imageio/plugins/fits.py
```diff
@@ -0,0 +1,119 @@
...
+    from astropy.io import fits as _fits
+    return _fits
+
+
+class FitsFormat(Format):
+
+    """ Flexible Image Transport System (FITS) is an open standard defining a
+    digital file format useful for storage, transmission and processing of
+    scientific and other images. FITS is the most commonly used digital
+    file format in astronomy.
```
almarklein: Should we mention that astropy is needed to use read/write images of this format?

blink1073:
All set.  I only added read support since that's all we had in scikit-image.

almarklein:
The tests should run on the python 2.7 build, correct? It seems they're skipped.

blink1073:
The tests are run without optional dependencies, then with them (that's what we did in scikit-image).

almarklein:
Should it not be enough to run the tests *with* optional dependencies on one build, and without in all the others?

blink1073:
Your call brother. 



—

almarklein:
I prefer one run :)

blink1073:
I'm getting this error all of a sudden:

 InternetNotAllowedError: Will not download resource from the internet because enironment variable IMAGEIO_NO_INTERNET is set.

almarklein:
Ah doh. When the test is run with `installed`, internet access is restricted.

blink1073:
I fixed it within the test.

almarklein:
Euhm, so I suppose we should run the test with astropy in another env (not the 2.7 one).

blink1073:
Ah, I see that it will skip the test entirely, ha!

blink1073:
What to create a specific build for optional dependencies?

blink1073:
Or just target 3.4?

almarklein:
Yes, 3.4 is sort of the "main"

blink1073:
Okay, done.

almarklein:
Great, thanks!

PR #107, opened 2015-07-07
Gdal plugin

blink1073:


imageio/plugins/gdal.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+
+    Parameters for reading
+    ----------------------
+    None
+
+    """
+
+    def _can_read(self, request):
+        return request.filename.lower().endswith(GDAL_FORMATS)
```
almarklein: Can we turn this into something like:

        if request.filename.lower().endswith('.ecw'):  # is that actual gdal?
            return True
        if has_module('osgeo.gdal'):
            return request.filename.lower().endswith(self.extensions)

blink1073: ECW is not a gdal format per se, (just like TIFF), but can have embedded geospatial data much like the others.  Since this is the only plugin providing that format, I say sure.

almarklein:
So this should be ready now?

blink1073:
Yep!

PR #108, opened 2015-07-07
Make wheels

### LINKED ISSUE #91: Use Wheels on PyPI
Right now, `pip install imageio` downloads the Linux tarball and associated `freeimage` library.  Wheels are preferred downloads for pip.

blink1073:
Fixes #91.

```bash
$ ls dist
imageio-1.3-py2.py3-linux_i386.whl
imageio-1.3-py2.py3-linux_x86_64.whl
imageio-1.3-py2.py3-macosx_10_6_intel.macosx_10_9_intel.macosx_10_9_x86_64.macosx_10_10_intel.macosx_10_10_x86_64.whl
imageio-1.3-py2.py3-win32.whl
imageio-1.3-py2.py3-win_amd64.whl
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -49,6 +49,12 @@ except ImportError:
     from distutils.core import setup  # Supports anything else
 
 
+try:
+    from wheel.bdist_wheel import bdist_wheel
+except ImportError:
+    bdist_wheel = object
+
+
 name = 'imageio'
 description = 'Library for reading and writing a wide range of image, video, scientific, and volumetric data formats.'
 
@@ -179,21 +185,23 @@ class build_with_fi(build_py):
         build_py.run(self)
 
 
-class sdist_all(sdist):
+class bdist_wheel_all(bdist_wheel):
     """ Build all platform specific dist files, that contain
     a few images and the freeimage lib of the platform.
     """
 
     def run(self):
-        sdist.run(self)
+        self.universal = True
+        bdist_wheel.run(self)
         
         import imageio
         
-        # Get base tarbal
-        import tarfile
+        # Get base zipfile
+        import zipfile
         distdir = op.join(THIS_DIR, 'dist')
-        tarfilename = op.join(distdir, 'imageio-%s.tar.gz' % __version__)
-        assert op.isfile(tarfilename)
+        basename = 'imageio-%s-py2.py3-none-any.whl' % __version__
+        zipfilename = op.join(distdir, basename)
+        assert op.isfile(zipfilename)
         
         # Create/clean build dir
         build_dir = op.join(distdir, 'temp')
@@ -202,11 +210,10 @@ class sdist_all(sdist):
         os.mkdir(build_dir)
         
         # Extract, get resource dir
-        with tarfile.open(tarfilename, 'r:gz') as tf:
+        with zipfile.ZipFile(zipfilename, 'r') as tf:
             tf.extractall(build_dir)
-        resource_dir = op.join(build_dir, 'imageio-%s' % __version__, 
-                               'imageio', 'resources')
-        assert os.path.isdir(resource_dir)
+        resource_dir = op.join(build_dir, 'imageio', 'resources')
+        assert os.path.isdir(resource_dir), build_dir
         
         # Prepare the libs resource directory with cross-platform
         # resources, so we can copy these for each platform
@@ -214,12 +221,13 @@ class sdist_all(sdist):
         
         # Create archives
         dist_files = self.distribution.dist_files
-        for plat in ['', 'linux64', 'linux32', 'win64', 'win32', 'osx64']:
+        for plat in ['linux64', 'linux32', 'win64', 'win32', 'osx64']:
             fname = self._create_dists_for_platform(resource_dir, plat)
-            dist_files.append(('sdist', 'any', 'dist/'+fname))
+            dist_files.append(('bdist_wheel', 'any', 'dist/'+fname))
         
         # Clean up
         shutil.rmtree(build_dir)
+        os.remove('dist/' + basename)
     
     
     def _create_dists_for_platform(self, resource_dir, plat):
@@ -240,7 +248,15 @@ class sdist_all(sdist):
         build_dir = op.join(distdir, 'temp')
         zipfname = 'imageio-%s.zip' % __version__
         if plat:
-            zipfname = 'imageio-%s-%s.zip' % (__version__, plat)
+            if plat == 'win64':
+                plat = 'win_amd64'
+            elif plat == 'osx64':
+                plat = 'macosx_10_6_intel.macosx_10_9_intel.macosx_10_9_x86_64.macosx_10_10_intel.macosx_10_10_x86_64'
+            elif plat == 'linux32':
+                plat = 'linux_i386'
+            elif plat == 'linux64':
+                plat = 'linux_x86_64'
+            zipfname = 'imageio-%s-py2.py3-%s.whl' % (__version__, plat)
         zipfilename = op.join(distdir, zipfname)
         zf = zipfile.ZipFile(zipfilename, 'w', zipfile.ZIP_DEFLATED)
         for root, dirs, files in os.walk(build_dir):
@@ -264,7 +280,7 @@ if 'bdist_wheel' in sys.argv:
 
 
 setup(
-    cmdclass={'sdist_all': sdist_all, 
+    cmdclass={'bdist_wheel_all': bdist_wheel_all, 
               'build_with_fi': build_with_fi,
               'test': test_command},
     
--- a/setup.py
+++ b/setup.py
@@ -221,16 +221,17 @@ class bdist_wheel_all(bdist_wheel):
         
         # Create archives
         dist_files = self.distribution.dist_files
-        for plat in ['linux64', 'linux32', 'win64', 'win32', 'osx64']:
-            fname = self._create_dists_for_platform(resource_dir, plat)
-            dist_files.append(('bdist_wheel', 'any', 'dist/'+fname))
+        for plat in ['win64', 'osx64']:
+            for pyver in ['27', '34']:
+                fname = self._create_wheels_for_platform(resource_dir,
+                                                         plat, pyver)
+                dist_files.append(('bdist_wheel', 'any', 'dist/'+fname))
         
         # Clean up
         shutil.rmtree(build_dir)
         os.remove('dist/' + basename)
-    
-    
-    def _create_dists_for_platform(self, resource_dir, plat):
+
+    def _create_wheels_for_platform(self, resource_dir, plat, pyver):
         import zipfile
         import imageio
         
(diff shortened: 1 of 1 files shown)


blink1073:
OSX is being irrationally difficult.  My 2 cents would be to include the OSX library in the tarball and provide the other wheels.

blink1073:
I'm going to take a break from this: it is maddening.

blink1073:
Apparently PyPI completely blocks Linux wheels at the moment...

blink1073:
I'm calling this a dead end.

almarklein:
Arg. Why is it so hard? Aren't many packages building and uploading wheels?

almarklein:
"Why is it so hard" not meant as a personal thing, but more like why is setuptools/pypi not making this easy for us...

blink1073:
Two examples:
https://pypi.python.org/pypi/pyzmq <https://pypi.python.org/pypi/pyzmq>
https://pypi.python.org/pypi/Pillow/2.9.0 <https://pypi.python.org/pypi/Pillow/2.9.0>

Both messy, neither one supports osx very well or Linux at all.


Regards,

Steve

* * * * * * *
Steven Silvester
Silvester Consulting, LLC
610 Maine Ave
Fort Walton Beach, FL 32547
steven.silvester@ieee.org <mailto:steven.silvester@ieee.org>
(850) 273-8276

blink1073:
Okay, updated thoughts:
Release wheels for win64 and osx64 for py27 and py34 (I figured out the syntax).
Include the linux64 version in the tarball.
All other combinations would have to download.

blink1073:
Note to self:
imageio-1.3-cp27-none-macosx_10_5_x86_64.whl
imageio-1.3-cp34-none-macosx_10_5_x86_64.whl
imageio-1.3-cp34-none-win_amd64.whl
imageio-1.3-cp27-none-win_amd64.whl

blink1073:
Here's my dist directory after running `python setup.py sdist_all bdist_wheel_all`:

```bash
$ ll dist
total 95336
-rw-r--r--  1 ssilvester  staff  4766656 Jul  9 11:05 imageio-1.3-cp27-none-macosx_10_5_x86_64.whl
-rw-r--r--  1 ssilvester  staff  4434964 Jul  9 11:05 imageio-1.3-cp27-none-win_amd64.whl
-rw-r--r--  1 ssilvester  staff  4766656 Jul  9 11:05 imageio-1.3-cp34-none-macosx_10_5_x86_64.whl
-rw-r--r--  1 ssilvester  staff  4434964 Jul  9 11:05 imageio-1.3-cp34-none-win_amd64.whl
-rw-r--r--  1 ssilvester  staff  5043110 Jul  9 11:06 imageio-1.3-linux32.zip
-rw-r--r--  1 ssilvester  staff  5114168 Jul  9 11:06 imageio-1.3-linux64.zip
-rw-r--r--  1 ssilvester  staff  4820461 Jul  9 11:06 imageio-1.3-osx64.zip
-rw-r--r--  1 ssilvester  staff  3205861 Jul  9 11:04 imageio-1.3-py3-none-any.whl
-rw-r--r--  1 ssilvester  staff  4290741 Jul  9 11:06 imageio-1.3-win32.zip
-rw-r--r--  1 ssilvester  staff  4488877 Jul  9 11:06 imageio-1.3-win64.zip
-rw-r--r--  1 ssilvester  staff  3231393 Jul  9 11:06 imageio-1.3.tar.gz
-rw-r--r--  1 ssilvester  staff   199552 Jul  9 11:06 imageio-1.3.zip
```

almarklein:
No wheels for Linux?

almarklein:
And what about 32bit? If its a matter of adding a few lines to make that work, why not? Or is there a problem I am overlooking?

I'd like to get rid of the sdist_all option if wheels work ok enough.

blink1073:
PyPI does not accept wheels, so the the minimum set you could get is:

win32 wheels
win64 wheels
linux32 tar
osx wheels
tar with linux64
source zip

My intent with the previous list was to only provide eggs for the most common cases, and
provide an escape hatch for everyone else.  For instance, win64 py33 users.

almarklein:
But in our case the cp34 wheels are exactly the same as our cp33 wheels since the binaries are not bound to a specific Python version. Can we just copy+rename in these cases?

blink1073:
Producing the wheels is not the problem, its just thriwing more build options together. Want me to add all possible valid combinations?



—

almarklein:
What do you mean by build options? Can we not just have `bdist_wheel_all` produce wheels for all combinations? What I meant is that from `imageio-1.3-cp27-none-win_amd64.whl` we can easily create a wheel for any Python version. So basically you can create 4 wheels and then copy+rename them for all Python versions that we want to support.

Or is the python version also included as meta info inside the wheel?

blink1073:
Ah, yes, that will work because when it builds we have the following created file:

```bash
Wheel-Version: 1.0
Generator: bdist_wheel (0.24.0)
Root-Is-Purelib: true
Tag: py2-none-any
Tag: py3-none-any
```

So it will grab the correct file based on name and install no matter what the underlying platform is.

blink1073:
Here's what the Pillow files look like:

```
Pillow-2.9.0-cp34-cp34m-macosx_10_6_intel.macosx_10_9_intel.macosx_10_9_x86_64.macosx_10_10_intel.macosx_10_10_x86_64.whl
Pillow-2.9.0-cp34-none-win32.whl
Pillow-2.9.0-cp34-none-win_amd64.whl
```

blink1073:
And for whatever reason the 2.7 Mac wheel is different:

`Pillow-2.9.0-cp27-none-macosx_10_6_intel.macosx_10_9_intel.macosx_10_9_x86_64.macosx_10_10_intel.macosx_10_10_x86_64.whl`

blink1073:
Let's bear in mind that `libfreeimage` is available on Homebrew, MacPorts, Debian, and soon Anaconda.  Are we fighting a battle that does not need to be be fought on Mac and Linux?

matthew-brett:
I'm trying to build a wheel-building repo : 

https://github.com/MacPython/imageio-wheels

I can now build (on travis) a dual arch Freeimage dylib, that should work on OSX 10.6, as OSX wheels require.

But - where should I copy this dylib into the imageio directory tree in order to build the wheel?

I currently copy into the `imageio` directory, but should it go into `resources`?

matthew-brett:
Building OK now, but lots of test errors: https://travis-ci.org/MacPython/imageio-wheels/builds/70906322

blink1073:
Yes, we've been putting it in `resources`.

almarklein:
`resources/freeimage` to be precise.

almarklein:
@matthew-brett The failures are because it cannot download the test images that are needed. Is it an idea to put a subset of images in the wheel as well?  It's these files:
```python
    for fname in ['images/chelsea.png',
                  'images/chelsea.zip',
                  'images/astronaut.png',
                  'images/newtonscradle.gif',
                  'images/cockatoo.mp4',
                  'images/realshort.mp4',
                  'images/stent.npz',
                  ]:
        imageio.core.get_remote_file(fname, resource_dir, 
                                     force_download=True)
```

matthew-brett:
Is that the desired behavior - to ship the images inside the wheel?

I guess you are planning that the user should be able to run the tests on their own installation?  So they should have the test images on their system too?

blink1073:
I think it makes sense to have these included in the install, the same way we do currently with `skimage.data`.

almarklein:
It also helps users to run basic examples without the need for an internet connection.

matthew-brett:
I wonder whether wheels allow you to specify the binary platform - like `macosx-10.6-intel` but no particular version of Python?  As in something like `imageio-1.3-py2.py3-none-macosx-10.6-intel.whl` - with one wheel for each of OSX and Windows?   Then at least you could build and test the OSX and Windows wheels separately.

almarklein:
@matthew-brett that would be convenient, but I've no idea if that's possible ...

almarklein:
What's the status on this?

blink1073:
I think it's waiting on me to finish MacPython/imageio-wheels.
I tried renaming the file, and it does not work:

```bash
$ pip install imageio-1.3-py2.py3-none-macosx_10_5_x86_64.whl
imageio-1.3-py2.py3-none-macosx_10_5_x86_64.whl is not a supported wheel on this platform.
```

matthew-brett:
What do you get for `python -c "import distutils.util;
print(distutils.util.get_platform())"` ?

blink1073:
macosx-10.5-x86_64

matthew-brett:
Yes, I'm seeing the same thing - looks like you can't specify `py2.py3`
with the platform stuff.

I did get pip to use a wheel with suffix
`-cp27.cp33.cp34-none-macosx_10_6_intel.whl ` though (with my python.org
Python, on 2.7 and 3.3).

blink1073:
Yep:

```bash
$ pip install imageio-1.3-cp27.cp34-none-macosx_10_5_x86_64.whl
Requirement already satisfied (use --upgrade to upgrade): imageio==1.3 from file:///Users/ssilvester/workspace/imageio/dist/imageio-1.3-cp27.cp34-none-macosx_10_5_x86_64.whl in /Users/ssilvester/anaconda/lib/python3.4/site-packages
```

blink1073:
@almarklein, I switched to `python make test unit`, but I'm still seeing four errors: https://travis-ci.org/blink1073/imageio-wheels/builds/79024328.

The three failures in `test_avbin.py` are occurring locally on my machine as well.

almarklein:
@blink1073 is that testing with a wheel or something? Not sure what to do with avbin. Maybe mark as known fails for now? Any idea why tempfile is imported? If it is expected, you can modify that test to ignore it.

blink1073:
@almarklein, those tests were on my local install on OSX 10.10.5.

almarklein:
But the Travis run, was it on imageio + modifications? I mean, so far Travis ran through the avbin tests ok. Any idea why not now?

blink1073:
I assumed the MacPython builds were running on OSX, is that right @matthew-brett?

matthew-brett:
Yes, right, they are running on an 10.9 OSX VM.

almarklein:
Ah, I did not realize it was a Travis mac run. I say just disable avbin tests on OSX for now. It's interesting though, because the person who contributed tje avbin plugin is a mac user.

blink1073:
I don't think skipping the tests is the right way to go, if the plugin can't read properly on OSX, it should at least indicate this to the user rather than failing because it can't read the metadata.

almarklein:
So, are you suggesting to catch this error and raise using an exception with a more descriptive message?

blink1073:
Yeah, I think that is a fair compromise.

blink1073:
The problem is in the use of the alpha version of libavbin.  I downloaded ``libavbin.10.dylib` and the tests pass.

blink1073:
@almarklein, would you accept a PR to this repo and imageio-binaries with version 10?

almarklein:
I just did :)

blink1073:
We now have http://a365fff413fe338398b6-1c8a9b3114517dc5fe17b7c3f8c63a43.r19.cf2.rackcdn.com/imageio-1.3-py2-none-any.whl, which should be valid as `imageio-1.3-cp27.cp33.cp34-none-macosx_10_5_x86_64.macosx_10_6_intel.whl`

blink1073:
I verified that the wheel installs for me and functions correctly.

blink1073:
Also works on python 3.5 if I add that to the name :smile:

almarklein:
Sweet! How is it generated? Can we do t from imageio's setup.py?

blink1073:
The wheel is generated by https://github.com/MacPython/imageio-wheels, we just need to update the submodule when we want to build a new version, then grab the wheel and rename it.

almarklein:
and manually update that wheel to pypi?

blink1073:
@matthew-brett, how do you deploy the OSX wheels for scikit-image?

matthew-brett:
I used the `wheel-uploader` script - instructions e.g. here : https://github.com/MacPython/numpy-wheels#uploading-the-built-wheels-to-pypi

blink1073:
Sounds reasonable, I'd need to update the file name in MacPython/imageio-wheels.

blink1073:
Now available: http://a365fff413fe338398b6-1c8a9b3114517dc5fe17b7c3f8c63a43.r19.cf2.rackcdn.com/imageio-1.3-cp26.cp27.cp33.cp34.cp35-none-macosx_10_5_x86_64.macosx_10_6_intel.whl.

almarklein:
Sweet! Does this mean the wheel-building is ready? Maybe we should document somewhere (in setup.py?) how we can build and deploy the wheels.

blink1073:
IMHO, we should just use `setup.py` (with the latest commit), since we are not "compiling" for either platform every time.

Issue #93, opened 2015-07-03
Add support for plugin preference


blink1073:
It would be nice to be able to handle preferences for plugins.
For example, `tifffile.py` should be always used as the default plugin for TIF files, if available, since it handles the widest assortment of TIF foibles.
Also, the user should be able to specify the desired plugin three ways:
- A global default
- A default for a given file extension
- As a keyword argument to `imread` or `imwrite`

almarklein:
The global preference order can be determined by the order in which plugins are registered, or imported in https://github.com/imageio/imageio/blob/master/imageio/plugins/__init__.py.

If we put the tifffile import beforr the freeimage, it should be fine.

How automatic format selection works: when a file should be read or written, imageio asks each format (a plugin provides one or more formats) whether it can handle it, in the order that they're registered. The first format that says it can do it, gets to do it. Formats can look at file extensions but most also inspects the first few bytes.

The user can also provide the format explicitly as the second arg to `imread` or `imwrite`.

So IIUC you propose to allow users to specify the plugins based on extensions too? I suppose we can do that, though I am not sure how necessary it is; for tiff we should just make tiffile the default, and if people *really* want to use freeimage, they can.

blink1073:
But, then they wouldn't benefit from the wrappers you have in place.  



—

almarklein:
Sorry, I don't follow ...

blink1073:
If the user knows that the only way to open their particular files properly is to use a specific plugin, they should be able to select that plugin.  The plugin is guaranteed to have uniform behavior with other plugins in imageio.  If they wanted to use the other library to load it (such as astropy), they could do so, but then they just find ways not to use imageio directly...



—

almarklein:
So, you mean that `imageio.imread('myfile.tif', 'tifffile')` is not sufficient, because the user might be using `astropy.imread('myfile.tif')`, which might use imageio, but now there is no way to specify the plugin?

blink1073:
Let's say `tifffile` were not the first plugin.  If the user had a TIFF file that `freeimage` could not load, they would be sunk, unless they could use `imageio.imread('myfile.tif', use_plugin='tifffile')`.

kuchi:
@blink1073 what you are saying imageio.imread('myfile.tif', 'tifffile') does already work. A user can specify the plugin as the 2nd argument. Right @almarklein?

blink1073:
The docstring is not clear enough in this regard, if that is the case.  



—

[...]

Issue #62, opened 2015-01-30
Pure Python ico/png


almarklein:
I have code in ~zoof~ Flexx that can read/write ICO, BMP, PNG in pure Python. Limited to RGB(A) though. But might be interesting to not need freeimage for png. And also for  #21

almarklein:
Need increases, because the current freeimage-based ico implementation segfaults on Travis on 3.5.

almarklein:
The cases for which this makes sense is either people who have somehow managed to install imageio without pillow, or people who need good ico support, in which case its probably related to some sort of app, which can include other code (e.g. https://github.com/zoofIO/flexx/blob/master/flexx/util/icon.py). Either way, adding such functionality at this point does not seem to benefit much, and costs extra weight to imageio, plus extra maintenance. 

Therefore, wont do this unless the need for it increases for some reason.

BTW: Pillow can export icons but reading only gives largest image.

FirefoxMetzger:
@almarklein This is an old issue flagged as wontfix. Revisiting this after 5 years, do you think anything has changed regarding your decision? I.e., do we still think that not adding this is the best solution?

If so, I suggest we close this issue to stay organized :)

almarklein:
If Pillow stays a hard dependency the need for this is indeed quite low. Let's close.

PR #110, opened 2015-07-09
Improvements to format selection and freeimage lazily loads its lib

almarklein:
* The format manager first selects formats based on the extensions and modes of the formats. Then it asks these whether they can read/write. If none could, it asks the remaining formats if they can read/write. 
* This allows more flexibility (e.g. support more formats than published in the extensions)
* It also avoids calling freeimage's `_can_read` when reading/writing a file that is known to be supported by another format (e.g. video)
* Freeimage lazily loads its library at the moment that it really needs it. Together with the above, the freeimage lib is never used/downloaded if you only use it for video or another non-freeimage format. 

This could use an extra pair of eyes. @blink1073. Also cc @Zulko because he'd like this.

Slightly related to #93. Paving the way for #62. Also helps towards #33.

### DIFF
--- a/imageio/plugins/_freeimage.py
+++ b/imageio/plugins/_freeimage.py
@@ -19,6 +19,7 @@ import os
 import sys
 import ctypes
 import threading
+from logging import warn
 import numpy
 
 from ..core import (get_remote_file, load_lib, Dict, resource_dirs, 
@@ -54,7 +55,7 @@ def get_freeimage_lib():
         except InternetNotAllowedError:
             pass
         except RuntimeError as e:  # pragma: no cover
-            print(str(e))
+            warn(str(e))
 
 
 # Define function to encode a filename to bytes (for the current system)
@@ -399,7 +400,30 @@ class Freeimage(object):
         
         # Make sure to keep a ref to function
         self._error_handler = error_handler
-        
+    
+    @property
+    def lib(self):
+        if self._lib is None:
+            try:
+                self.load_freeimage()
+            except OSError as err:
+                self._lib = 'The freeimage library could not be loaded: '
+                self._lib += str(err)
+        if isinstance(self._lib, str):
+            raise RuntimeError(self._lib)
+        return self._lib
+    
+    def has_lib(self):
+        try:
+            self.lib
+        except Exception:
+            return False
+        return True
+    
+    def load_freeimage(self):
+        """ Try to load the freeimage lib from the system. If not successful,
+        try to download the imageio version and try again.
+        """
         # Load library and register API
         success = False
         try:
@@ -407,7 +431,7 @@ class Freeimage(object):
             # to the imageio-provided lib (if previously downloaded)
             self._load_freeimage()
             self._register_api()
-            if self._lib.FreeImage_GetVersion().decode('utf-8') >= '3.15':
+            if self.lib.FreeImage_GetVersion().decode('utf-8') >= '3.15':
                 success = True
         except OSError:
             pass
@@ -419,8 +443,8 @@ class Freeimage(object):
             self._register_api()
         
         # Wrap up
-        self._lib.FreeImage_SetOutputMessage(self._error_handler)
-        self._lib_version = self._lib.FreeImage_GetVersion().decode('utf-8')
+        self.lib.FreeImage_SetOutputMessage(self._error_handler)
+        self.lib_version = self.lib.FreeImage_GetVersion().decode('utf-8')
     
     def _load_freeimage(self):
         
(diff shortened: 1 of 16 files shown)


almarklein:
@blink1073 Also design question about simpleITK and other plugins that rely on an external module: when asked whether they can read/write, should they perhaps check if the module is available?

Advantage: if not, you get an error if the package was not installed, while perhaps another plugin could have done it.
Disadvantage: you do *not* get an error saying that you need to install X.

blink1073:
How about this:

```python
In [6]: importlib.find_loader('numpy')
Out[6]: <_frozen_importlib.SourceFileLoader at 0x103940828>

In [7]: importlib.find_loader('simpleitk')

In [8]:
```

That way, we check for its existence without loading it, and if the user explicitly asks for the plugin by name, it kicks out the error?

imageio/plugins/_freeimage.py
```diff
@@ -399,7 +400,30 @@ def error_handler(fif, message):
         
         # Make sure to keep a ref to function
         self._error_handler = error_handler
-        
+    
+    @property
```
blink1073: I like this pattern, we could generalize this into a generic property and and `load_library` method.

almarklein: Yes, we could generalize this, but since freeimage and avbin are the only plugins that use a library, I think we can hold that off until we get more such plugins.

almarklein:
@blink1073  I like the `importlib` trick, but that only works on Py3k, so we still need an import for Py2.x. I now made it so that the itk plugin reports True if the request was for an extension that only ITK supports, and otherwise only return True if the package is installed. (see last commit).

blink1073:
The same works for `imp.find_module`.

blink1073:
See https://github.com/scikit-image/scikit-image/blob/master/skimage/__init__.py#L70, which has been tested on Py26-Py34.

almarklein:
Thanks. I kept the importlib for Py3k, since imp is to be deprecated.

blink1073:
What if I made a utility function that just Does the Right Thing?

almarklein:
Like a `has_lib(lib_name)`? Go ahead :)

blink1073:
Might make sense to add it to this PR yourself in https://github.com/imageio/imageio/blob/master/imageio/core/util.py:

```python
def has_lib(lib_name):
    """Check to see if a python library is available.
    """
    if sys.version_info > (3, ):
        import importlib
        return importlib.find_loader(lib_name) is not None
    else:
        import imp
        try:
            imp.find_module(lib_name)
        except ImportError:
            return False
        return True
```

almarklein:
done

imageio/plugins/freeimage.py
```diff
@@ -353,7 +352,71 @@ def _append_data(self, im, meta):
...
+        FormatClass = SPECIAL_CLASSES.get(name.lower(), FreeimageFormat)
+        if FormatClass:
+            # Create Format and add
+            format = FormatClass(name, des, ext, FormatClass._modes)
+            format._fif = i
+            formats.add_format(format) 
+
+
 def create_freeimage_formats():
+    """ By default, imageio registeres a list of predefined formats
```
blink1073: spelling (registers)

imageio/plugins/example.py
```diff
@@ -37,10 +37,16 @@ def _can_read(self, request):
         # can do it.
         #
         # The format manager is aware of the extensions and the modes
-        # that each format can handle. However, the ability to read a
-        # format could be more subtle. Also, the format would ideally
-        # check the request.firstbytes and look for a header of some
-        # kind. Further, the extension might not always be known.
+        # that each format can handle. It will first ask all formats
+        # that seem* to be able to read it whether they can. If none
```
blink1073: should be an asterisk before `seem`

imageio/plugins/example.py
```diff
@@ -49,21 +55,21 @@ def _can_read(self, request):
...
-                    return True
+            if request.filename.lower().endswith(self.extensions):
+                return True
     
     def _can_write(self, request):
         # This method is called when the format manager is searching
-        # for a format to write a certain image. Return True if the
-        # format can do it.
+        # for a format to write a certain image. It will first ask all
+        # formats that seem* to be able to write it whether they can.
```
blink1073: Another missing asterisk

almarklein:
Thanks for the fixes! Fixed (that commit message was not intentional).

blink1073:
*Irony*

blink1073:
Ready to merge?

almarklein:
yeah

blink1073:
Progress!

Issue #29, opened 2014-11-06
Plugin: MHD format for reading ITK image data


almarklein:
Used e.g. by the Elastix registration toolkiit.

http://www.itk.org/Wiki/ITK/MetaIO/Documentation

almarklein:
Reopening; this is covered by the `simpleitk` plugin, but that dependency is not pure Python (I think?) and you cannot simply pip install it.

MHD is a simple format that we can implement in pure Python in imageio.

@blink1073 thoughs?

blink1073:
Seems reasonable to leave this open, but I wouldn't say it is a "simple" format. :smile:

gordon-n-stevenson:
MetaImage is always a pain but super useful for Medical Images. I've suggested a pull request to the SimpleITK plugin - adding formats '.mha' and '.nii' formats These are single file 3D image formats that solve I think the problems of a header and separate raw file.

FirefoxMetzger:
@almarklein As you mentioned here a (long) while ago, MHD is supported via itk/simpleitk. Since then, our policy has become more liberal towards non-python backends, so I think it is okay to have MDH support via simpletik for those that need it.

Since this issue doesn't suggest a specific alternative backend for MHD, should we close it for now? (We can always open a new issue for a dedicated plugin/backend for MHD if this becomes relevant in the future.)

almarklein:
I've worked on a pure Python mhd implementation a year or so ago. It's in a propitiatory codebase right now, but I think ppl may be open to bringing that code here. That said, yes lets close this.

PR #113, opened 2015-07-25
Add Single 3D/4D volume file support (.mha and .nii)

### LINKED ISSUE #14: Format for 3D data (because npz is not so good)
NPZ is not supported on pypy and has issues with certain combinations of Python 2.x and numpy. We better had a good alternative ...

gordon-n-stevenson:
fixes #14 and #29

Using SimpleITK and adding the following file extensions adds NIFTI and MetaImage format that is in a single file format.

### DIFF
--- a/imageio/plugins/simpleitk.py
+++ b/imageio/plugins/simpleitk.py
@@ -28,7 +28,7 @@ def load_lib():
 
 
 # Split up in real ITK and all supported formats.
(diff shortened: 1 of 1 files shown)


almarklein:
Looks good to me, @blink1073 contributed this plugin so I'll let him have a look and merge :)

blink1073:
Thanks @gordon-n-stevenson!

PR #116, opened 2015-08-24
Fix for getting correct number of frames in video from ffmpeg. Needed…

kuchi:
… to change 
    
    int(duration*fps)
 
to

    int(round(duration*fps))

Otherwise could report frame count in error (off by 1).


kuchi:
Looks like it passes all the test. I'm not sure what causes the appveyor to fail in downloading ffmpeg, but I didn't change anything related to that. If you are ok with this change @almarklein it's complete on my end. I couldn't make a unittest that perfectly exposed this bug, but I had a video that encountered this issue and this change correctly fixes this problem.

almarklein:
Thanks for the fix.

PR #119, opened 2015-10-03
Fix OSX failures

blink1073:
Relies on https://github.com/imageio/imageio-binaries/pull/3.

### DIFF
--- a/imageio/plugins/avbin.py
+++ b/imageio/plugins/avbin.py
@@ -21,7 +21,7 @@ from ..core import (Format, get_platform, get_remote_file,
 
 
 FNAME_PER_PLATFORM = {
-    'osx64': 'libavbin-11alpha4-osx.dylib',
+    'osx64': 'libavbin-10-osx.dylib',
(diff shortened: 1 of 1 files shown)


tests/test_core.py
```diff
@@ -838,8 +838,11 @@ def test_functions():
     imageio.mimsave(fname5, [im[:, :, 0] for im in ims])
     imageio.mimsave(fname5, ims)
     assert os.path.isfile(fname5)
-    
-    # Test volread()
+
+    if not sys.platform.startswith('linux'):
+        return
```
almarklein: Why is this? Does this not work on osx? If it does not, maybe `if not sys.platform.startswith('darwin'):` because we do test on windows.

blink1073: Interesting, something must have been wrong with the `stent.npz` file in my `resources` directory.  I deleted it and the test passes, I'll remove this.

almarklein:
Must have gone corrupted during download.

PR #124, opened 2015-10-25
fix leaking file descriptor

zshipko:


almarklein:
I've always understood that doing it this way closes the file right away. Are there cases where it does not?

zshipko:
I'm not entirely sure. I just ran into a bunch of warnings about this last night when I was running some unit tests and figured I should just fix it -- at least, if nothing else, to avoid the errors while testing.

almarklein:
Fair enough

almarklein:
Thanks

PR #131, opened 2016-01-12
Fix bug where the ffmpeg plugin fails on certain video files

jackwalker64:
When trying to open a video that contains a lot of metadata, the ffmpeg plugin will limit the number of lines in the caught header, resulting in some of the required metadata being lost.

To reproduce, download and extract [many_chapters.zip](https://github.com/imageio/imageio/files/87004/many_chapters.zip) and run

    import imageio
    reader = imageio.get_reader("many_chapters.mp4")

Expected behaviour: the ffmpeg plugin opens the file without issue.
Actual behaviour: IndexError in [imageio/plugins/ffmpeg.py:461](https://github.com/imageio/imageio/blob/872a317b6d69a1d65ec07561754e88f797461fae/imageio/plugins/ffmpeg.py#L461) due to the "Duration: " line having been thrown away.

My proposed fix is to just wait until the header has been set before limiting the subsequent lines.

### DIFF
--- a/imageio/plugins/ffmpeg.py
+++ b/imageio/plugins/ffmpeg.py
@@ -819,7 +819,8 @@ class StreamCatcher(threading.Thread):
                     header = b'\n'.join(self._lines)
                     self._header += header.decode('utf-8', 'ignore')
                     self._lines = []
-            self._lines = limit_lines(self._lines)
+            if self._header:
+                self._lines = limit_lines(self._lines)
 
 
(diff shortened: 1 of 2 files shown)


almarklein:
This looks good to me. Travis is segfaulting for some reason, but it seems unrelated.

almarklein:
Can you please rebase/merge with the latest master? The travis segfault has been fixed.

almarklein:
Thanks for this!

PR #134, opened 2016-01-13
Disable travis tests for gdal to prevent segfaults

almarklein:
cc @blink1073 

Even if gdal tests are skipped, the segfault occurs. Maybe it already happens when imported? Or maybe gdal tries to open a file that it should not? This could do with some more research, but for now I just don't install gdal on Travis.


blink1073:
Huh, strange.

PR #138, opened 2016-03-16
Search path for ffmpeg before downloading

tdsmith:
Brings behavior of get_exe in line with FfmpegFormat docstring:

    Otherwise, imageio will look for ffmpeg on the system path and then
    download ffmpeg if not found.

Also cache the result of get_exe on the classes, because shelling out to ffmpeg -version is probably too expensive to do on each instance.


imageio/plugins/ffmpeg.py
```diff
@@ -45,6 +45,12 @@ def get_exe():
     if exe:  # pragma: no cover
         return exe
 
+    try:
+        sp.check_output(["ffmpeg", "-version"])
```
tdsmith: `check_output` doesn't exist on 2.6.

imageio/plugins/ffmpeg.py
```diff
@@ -45,6 +45,12 @@ def get_exe():
...
+        return "ffmpeg"
+    except sp.CalledProcessError:
```
tdsmith: `FileNotFoundError` is raised on 3.x (and doesn't exist on 2.x), which inherits from `OSError`.

imageio/plugins/ffmpeg.py
```diff
@@ -45,6 +45,15 @@ def get_exe():
     if exe:  # pragma: no cover
         return exe
 
+    # Check if ffmpeg is in PATH
+    try:
+        sp.check_call(["ffmpeg", "-version"], stdout=sp.PIPE, stderr=sp.PIPE)
```
tdsmith: Nope, this can deadlock.

almarklein:
LGTM. Will merge in a day or two unless @kuchi has comments.

kuchi:
LGTM, all tests pass on my OS X.

almarklein:
Thanks both!

tdsmith:
Thanks for a great tool and a quick response; very gratifying :)

PR #139, opened 2016-03-16
Don't print ffmpeg invocation for quieter loglevels

tdsmith:
"fatal" and "panic" are quieter loglevels than "warning" so it makes
sense that invocations should only be printed for "info" and noisier.


almarklein:
I'll let @kuchi take a look, since I think he wrote that.

kuchi:
LGTM

Issue #140, opened 2016-03-17
Fails with freeimage_lib(IMAGEIO_FREEIMAGE_LIB) custom path (MS Windows)


nad2000:
1. invoke the package to make it download DDL. It gets installed into the local app location:
```
    >>> im = imageio.imread("IM\\app.png")
    Imageio: 'freeimage-3.15.4-win32.dll' was not found on your computer; downloading it now.
    Try 1. Download from https://github.com/imageio/imageio-       binaries/raw/master/freeimage/FreeImage-3.15.4-win32.dll (2.7 MB)
    Downloading: 2790912/2790912 bytes (100.0%)
      Done
    File saved as C:\Users\nad2000\AppData\Local\imageio\freeimage\freeimage-3.15.4-win32.dll.
    >>>
```
2. move DDL a different directory, eg,  **C:\Users\nad2000\libs** and setup the: `SET IMAGEIO_FREEIMAGE_LIB=C:\Users\nad2000\libs\freeimage-3.15.4-win32.dll`
3. Repeat the 1st step. it doesn't attempt to download DDL (apparently the lib was found) and it throws an exception:
```
>>> import imageio
>>> im = imageio.imread("IM\\app.png")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Python27\lib\site-packages\imageio\core\functions.py", line 181, in imread
    reader = read(uri, format, 'i', **kwargs)
  File "C:\Python27\lib\site-packages\imageio\core\functions.py", line 108, in get_reader
    'in mode %r' % mode)
ValueError: Could not find a format to read the specified file in mode 'i'
```

nad2000:
**get_freeimage_lib** returns the path, but **_load_freeimage** is looking for the lib in the default cached directory (in _freeimage.py). Would it OK, if in **load_freeimage** the returned path gets passed to the loader?

PR #141, opened 2016-03-17
added IMAGEIO_FREEIMAGE_LIB path the the exact_lib_names

nad2000:
a fix for bug #140


almarklein:
This looks good to me. It looks like this was a bug because the result from `get_freeimage_lib()` is never actually used. This seems like a good fix. Thanks!

PR #148, opened 2016-05-06
fix appveyor pip install

almarklein:


almarklein:
appveyor seems still to be having issues. Let's merge for now.

PR #150, opened 2016-05-06
Added support for 16-bit PNGs

### LINKED ISSUE #146: Support writing 16-bit color channels for PNG
I'm trying to transition away from MATLAB and imageio has been incredibly helpful. However, I need to work with 64-bit PNGs (16-bit color channels and 16-bit alpha channel) and although it looks like reading 16-bit channels is supported, writing is not supported.

I created some 64-bit pngs in MATLAB using the following code:

    img_out = zeros(256, 256, 3, 'uint16');
    img_alpha = zeros(256, 256, 'uint16');
    color_grad = uint16(reshape(0:2^16-1, 256, []));
    img_out(:, :, 1) = color_grad;
    img_out(:, :, 2) = rot90(color_grad, 1);
    img_out(:, :, 3) = rot90(color_grad, 2);
    img_out_alpha = rot90(color_grad, 3);

    fprintf('Write unique values: R=%u, G=%u, B=%u, A=%u\n', ...
            length(unique(img_out(:, :, 1))), ...
            length(unique(img_out(:, :, 2))), ...
            length(unique(img_out(:, :, 3))), ...
            length(unique(img_out_alpha)))
    imwrite(img_out, '64bit_MATLAB.png', 'Alpha', img_out_alpha, 'BitDepth', 16)
    
    [img_in, ~, img_in_alpha] = imread('64bit_MATLAB.png');
    fprintf('MATLAB PNG unique values: R=%u, G=%u, B=%u, A=%u\n', ...
            length(unique(img_in(:, :, 1))), ...
            length(unique(img_in(:, :, 2))), ...
            length(unique(img_in(:, :, 3))), ...
            length(unique(img_in_alpha)))

This gives me the following output:

> Write unique values: R=65536, G=65536, B=65536, A=65536
> MATLAB PNG unique values: R=65536, G=65536, B=65536, A=65536

If I try the same with imageio in Python:

    import imageio
    import numpy as np

    img_out = np.zeros((256, 256, 4), dtype=np.uint16)
    color_grad = np.reshape(np.arange(2**16), (256,-1))
    img_out[:, :, 0] = color_grad
    img_out[:, :, 1] = np.rot90(color_grad, 1)
    img_out[:, :, 2] = np.rot90(color_grad, 2)
    img_out[:, :, 3] = np.rot90(color_grad, 3)

    print('Write unique values: R={}, G={}, B={}, A={}'.format(
        len(set(img_out[:, :, 0].flatten().tolist())),
        len(set(img_out[:, :, 1].flatten().tolist())),
        len(set(img_out[:, :, 2].flatten().tolist())),
        len(set(img_out[:, :, 3].flatten().tolist()))))
    imageio.imwrite('64bit_imageio.png', img_out)

    img_in = imageio.imread('64bit_imageio.png')
    print('imageio PNG unique values: R={}, G={}, B={}, A={}'.format(
        len(set(img_in[:, :, 0].flatten().tolist())),
        len(set(img_in[:, :, 1].flatten().tolist())),
        len(set(img_in[:, :, 2].flatten().tolist())),
        len(set(img_in[:, :, 3].flatten().tolist()))))

I get the following:

> Write unique values: R=65536, G=65536, B=65536, A=65536
> imageio PNG unique values: R=256, G=256, B=256, A=256

To confirm, imageio is able to read the 64-bit PNG:

    img_in = imageio.imread('64bit_MATLAB.png')
    print('MATLAB PNG unique values: R={}, G={}, B={}, A={}'.format(
        len(set(img_in[:, :, 0].flatten().tolist())),
        len(set(img_in[:, :, 1].flatten().tolist())),
        len(set(img_in[:, :, 2].flatten().tolist())),
        len(set(img_in[:, :, 3].flatten().tolist()))))

> MATLAB PNG unique values: R=65536, G=65536, B=65536, A=65536

Any idea how difficult this would be to implement?

OrganicIrradiation:
Fixes #146

### DIFF
--- a/imageio/core/util.py
+++ b/imageio/core/util.py
@@ -13,6 +13,7 @@ import os
 import sys
 import time
 import struct
+from warnings import warn
 
 import numpy as np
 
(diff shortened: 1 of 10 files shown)


imageio/core/util.py
```diff
@@ -46,39 +47,69 @@ def urlopen(*args, **kwargs):
...
+        bitdepth = 8
+    if not isinstance(img_in, np.ndarray):
+        raise ValueError('Image must be a numpy array')
+    if bitdepth == 8:
+        out_type = np.uint8
+    elif bitdepth == 16:
+        out_type = np.uint16
+    else:
+        raise ValueError('Bitdepth must be either 8 or 16')
+    im = img_in.copy()
```
almarklein: Why the copy? The docs say we return as-is if no change is required.

almarklein:
Awesome work! So no change was required to the freeimage wrapper to make this work? Other than the image copy, this looks great.

imageio/core/util.py
OrganicIrradiation: That should be addressed... I was having issues with the tests because the input image is mutable (without the copy) and some of the tests fail without the copy. For example, in test_freeimage.py test_png(), the assertion for lines 151-155 would fail without the copy because ```rim``` would be rescaled by image_as_uint during saving. Here are notes *without the .copy()*:

    rim = get_ref_im(colors, crop, is float) # Note: ranges from 0.0 to 1.0
    imageio.imsave(fname, rim) # After this, rim would have been rescaled to 0 to 255
    im = imageio.imread(fname)
    mul = 255 if isfloat else 1 # Assumes that rim's range is 0.0 to 1.0
    assert_close(rim * mul, im, 0.1)  # This fails because 255 * [0,255] >> [0, 255]

imageio/core/util.py
```diff
@@ -46,39 +47,69 @@ def urlopen(*args, **kwargs):
...
+       (im.dtype == np.uint16 and bitdepth == 16)):
+        # Already the correct format? Return as-is
         return im
-    # Handle float
-    mi, ma = np.nanmin(im), np.nanmax(im)
-    if dtype_str.startswith('float'):
-        if mi >= 0 and ma <= 1:
-            mi, ma = 0, 1
-    # Now make float copy before we scale
-    im = im.astype('float32')
```
OrganicIrradiation: This line makes a copy of im (see [numpy.ndarray.astype](http://docs.scipy.org/doc/numpy-1.10.1/reference/generated/numpy.ndarray.astype.html)). When it was removed, I needed to make sure that the input image was not changed in place.

imageio/core/util.py
almarklein: Yes, I think it makes sense to delay the copying until after the returning of the (unchanged) image when dtype is already correct. 

I think that if you change `im *= np.power(2.0, bitdepth)-1` (for the conversion from float) to ` im = im * ...`, we've dropped all inplace operations and can discard the copy alltogether.

OrganicIrradiation: Great catch, the im *= ... is in-place while im = im * ... works on a copy. Fixed.

almarklein:
Awesome. Is this ready?

OrganicIrradiation:
I thought it would be good to go, but it is failing when a NaN is passed in test_swf.py test_types(). This is an odd situation... Why should NaNs be silently allowed to pass through when converting to uint? (original behavior)

I tried to clean up the functions tests to improve the coverage as well.

almarklein:
I think this is because of using `empty`, which is a bad idea, because the values can be anything. I think they happen to contain inf or -inf for the failing test. Your extra tests just made the problem explicit. Better use arrays with actual values that we control. Do you mind fixing that?

OrganicIrradiation:
Sure, no problem. Just to clarify, do you think the SWF test should be updated with int/uint/float arrays for test_types()?

almarklein:
More tests are good. Most important is to get replace that `np.empty` with something else.

OrganicIrradiation:
Looks to be good-to-go? Thanks for your assistance in getting this implemented.

almarklein:
Awesome! Thanks for this great contribution!

PR #162, opened 2016-09-16
Allow saving tiff resolution unit.

### LINKED ISSUE #142: resolution_unit tag for TIFF format
This is a small enhancement request for the TIFF format writer: support writing the resolution unit metadata entry (which can be inches (the current default) or cm, see http://www.awaresystems.be/imaging/tiff/tifftags/resolutionunit.html).

anntzer:
Fixes #142.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,4 +26,4 @@ __pychache__
 *.pyo
 *.dll
 *.dylib
-
+.*.swp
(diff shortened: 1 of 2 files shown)


almarklein:
I think a change like that should be contributed to the tiffille project, since we're basically vendoring that. A change done to the file here will be undone when we update tiffile.

Correct me if I'm wrong @blink1073

anntzer:
Well, I noticed that C. Gohlke recently implemented the same thing, so just updating it would work too :-) (http://www.lfd.uci.edu/~gohlke/code/tifffile.py.html see 2016-05-13 entry).

blink1073:
I'd say it is fine to merge this here, I'll update tifffile as well.

almarklein:
Tests fails are spurious. Let's merge then.

Issue #144, opened 2016-04-02
issue saving exr file


moushkka:
Saving to exr seems to result in a black image on my computer.

``` python
import numpy as np
import imageio
original = imageio.imread('original.exr')
print('original min: {} max: {}'.format(np.min(original), np.max(original)))
imageio.imsave('saved.exr', original)
saved = imageio.imread('saved.exr')
print('saved min: {} max: {}'.format(np.min(saved), np.max(saved)))
```

results in the printout:

```
original min: 0.62353515625 max: 1.0
saved min: 0.0 max: 0.0
```

Here is an exr file for testing.
[original.exr.zip](https://github.com/imageio/imageio/files/200776/original.exr.zip)

I'm using python 3.5.1.
imageio is version 1.5 installed using pip.
I'm running this on osx.

almarklein:
Reproduced on Windows. It could be a bug either in FreeImage, or in our freeimage wrapper. It does not help to write as uint8 (because exr does not support that).

so-rose:
Same exact issue here. Trying to write '.hdr' images also breaks. Python 3.5.1/imageio 1.5 via pip on Debian Linux.

I **really** would like to use imageio to drive my entire current project's IO (https://github.com/so-rose/openlut). EXR is a must, though, alas this bug is very irritating! (I might implement imageio anyway as the default plugin for formats other than EXR and DPX, for now).
<br>
I would love to fix this myself & make a pull request, if I knew where to start. I have no experience with ctypes, and limited experience with C++, however. If you point me in the right direction, I can try, though!

almarklein:
@so-rose I good first test would be to try and see if the same error occurs with freeimage itself. That way we could establish whether its a problem with FreeImage or our wrapper.

Although, if this problem occurs with *every* exr image, it is very likely the wrapper.

so-rose:
Hmm... FreeImage, C++ version, works great (I also do think that EXR and HDR not working would have been noticed!).

I've also tried out loading some different images - 16-bit, 8-bit, hdr depth channel, etc. . So far, nothing - just a black file, even though the array looks fine

A note about the black EXR/.hdr output: It seems to be 16 bit black values, not 32 bit, which is a bit curious - imageio seems to save according to dtype quite well. I also replicated the black EXR and .hdr output in Python 2.7.12.


So, **it's most likely the wrapper**!
<br>
[...]

PR #165, opened 2016-09-24
Partial fix for saving exr images

moushkka:
Just had a quick look at bug #144
It seems to me like the old code shouldn't work at all for big endian systems or images in formats other than uint8 or uint16.
Or am I missing something?

The exr images are now saved upside down though :(
I'm guessing freeimage doesn't like the scanline swap for some formats?

### DIFF
--- a/imageio/plugins/_freeimage.py
+++ b/imageio/plugins/_freeimage.py
@@ -957,24 +957,16 @@ class FIBitmap(FIBaseBitmap):
         wrapped_array = self._wrap_bitmap_bits_in_array(w_shape, dtype, True)
         # swizzle the color components and flip the scanlines to go to
         # FreeImage's BGR[A] and upside-down internal memory format
-        if len(shape) == 3:
+        if len(shape) == 3 and isle and dtype.type == numpy.uint8:
             R = array[:, :, 0]
             G = array[:, :, 1]
             B = array[:, :, 2]
-        
-            if isle:
-                if dtype.type == numpy.uint8:
-                    wrapped_array[0] = n(B)
-                    wrapped_array[1] = n(G)
-                    wrapped_array[2] = n(R)
-                elif dtype.type == numpy.uint16:
-                    wrapped_array[0] = n(R)
-                    wrapped_array[1] = n(G)
-                    wrapped_array[2] = n(B)
-                #
-                if shape[2] == 4:
-                    A = array[:, :, 3]
-                    wrapped_array[3] = n(A)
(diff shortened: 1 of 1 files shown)


moushkka:
Found the upside down issue.
Seems like any image that hit the default freeimage codepath would misbehave.

imageio/plugins/_freeimage.py
```diff
@@ -953,28 +953,20 @@ def set_image_data(self, array):
             n_channels = shape[0]
         
         def n(arr):  # normalise to freeimage's in-memory format
-            return arr.T[:, ::-1]
+            return arr[::-1].T
```
almarklein: I don't see how this would fix it. The net result should be the same. I verfied that it is for a png image. Maybe for exr it does make a difference? Is it not 2D?

moushkka: In the old code the T would reverse the axis order and then flip the second axis (x axis).
The new version flips the first axis (the y axis) and then reverses the axis order.

moushkka: It would have been the same if arr was a grayscale image.

imageio/plugins/_freeimage.py
```diff
@@ -953,28 +953,20 @@ def set_image_data(self, array):
...
-                #
-                if shape[2] == 4:
-                    A = array[:, :, 3]
-                    wrapped_array[3] = n(A)
+            wrapped_array[0] = n(B)
+            wrapped_array[1] = n(G)
+            wrapped_array[2] = n(R)
+            if shape[2] == 4:
+                A = array[:, :, 3]
+                wrapped_array[3] = n(A)
```
almarklein: So if I understand correctly, the problem in your case was that the data was big endian, and the code was broken for big endian?

moushkka: Well, no.
isle says whether the system is big endian right? Which I guess rarely happens anymore. So I don't care much about that.
The problem with the old code was that if the image was a color image (len(shape) == 3) with a float32 dtype nothing would be written to wrapped_array.

almarklein: Ah, you're right. Do you happen to know why the color channels are supposed to be re-ordered for uint8 images, but not for anything else? If so, can you please add a comment?

moushkka: I added a note. The documentation isn't very informative though.
The channel ordering is supposedly to "ease the use of FreeImage with graphics API".

almarklein:
Thanks!

PR #173, opened 2016-10-19
Continue if seek not supported, but also not really needed

### LINKED ISSUE #169: seek needed for get_reader
I am trying to call:

``` python
imageio.get_reader(img_data, format='MP4')
```

where `img_data` is a stream (file like object without `seek`). This cause an error:

```
UnsupportedOperation
seek
imageio/core/request.py in get_file at line 290

        # Is there already a file?
        # Either _uri_type == URI_FILE, or we already opened the file, 
        # e.g. by using firstbytes
        if self._file is not None:
            self._file.seek(0)
            return self._file

        if self._uri_type == URI_BYTES:
            if want_to_write:                          
                self._file = BytesIO()
```

is `seek` really needed, or can we make `get_reader` handle a file like object without seek?

almarklein:
Fixes #169, hopefully.

### DIFF
--- a/imageio/core/request.py
+++ b/imageio/core/request.py
@@ -286,10 +286,6 @@ class Request(object):
         # Either _uri_type == URI_FILE, or we already opened the file, 
         # e.g. by using firstbytes
         if self._file is not None:
-            try:
-                self._file.seek(0)
-            except OSError:  # UnsupportedOperation derives from OSError
-                pass  # ah well, let's assume that the it's in the correct pos
             return self._file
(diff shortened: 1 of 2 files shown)


cancan101:
why not eliminate the seek entirely? seems weird to take a passed in file handle and then seek it to 0 as opposed to expecting the caller to have done that if needed.

almarklein:
Good point. Maybe it was a leftover from how `getfirtstbytes` used to work.

PR #182, opened 2016-12-07
Add FEI format reader

jni:
Obviously the tests will fail until imageio/imageio-binaries#4 is merged. Otherwise I may be doing something stupid but seems to work for me... Guidance appreciated!

### DIFF
--- a/imageio/plugins/feisem.py
+++ b/imageio/plugins/feisem.py
@@ -10,7 +10,21 @@ from .. import formats
 
 
 class FEISEMFormat(TiffFormat):
-    """Provide read support for TIFFs produced by an FEI SEM microscope."""
+    """Provide read support for TIFFs produced by an FEI SEM microscope.
+
+    This format is based on TIFF, and supports the same parameters.
+
+    FEI microscopes append metadata as ASCII text at the end of the file,
+    which this reader correctly extracts.
+
+    Parameters for get_data
+    -----------------------
+    discard_watermark : bool
+        If True (default), discard the bottom rows of the image, which
+        contain no image data, only a watermark with metadata.
+    watermark_height : int
+        The height in pixels of the FEI watermark. The default is 70.
+    """
 
     class Reader(TiffFormat.Reader):
         def _get_data(self, index=0, discard_watermark=True,
--- a/imageio/plugins/feisem.py
+++ b/imageio/plugins/feisem.py
@@ -1,0 +1,50 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2016, imageio contributors
+# imageio is distributed under the terms of the (new) BSD License.
+
+from __future__ import absolute_import, unicode_literals
+
+from .tifffile import TiffFormat
+
+
+class FEISEMFormat(TiffFormat):
+    """Provide read support for TIFFs produced by an FEI SEM microscope."""
+
+    class Reader(TiffFormat.Reader):
+
+        def _get_meta_data(self, index=None):
+            """Read the metadata from an FEI SEM TIFF.
+
+            This metadata is included as ASCII text at the end of the file.
+
+            The index, if provided, is ignored.
+
+            Returns
+            -------
+            metadata : dict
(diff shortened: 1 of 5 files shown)


imageio/plugins/feisem.py
```diff
@@ -0,0 +1,72 @@
...
+
+from __future__ import absolute_import, unicode_literals
+
+from .tifffile import TiffFormat
+
+from .. import formats
+
+
+class FEISEMFormat(TiffFormat):
+    """Provide read support for TIFFs produced by an FEI SEM microscope."""
```
almarklein: This is the docstring that ends up [here](http://imageio.readthedocs.io/en/latest/formats.html). I think it could mention that the format derives from the Tifffile format, and that it supports the same parameters. Further, this would be the place to document the kwargs in `get_data`. See the avbin plugin for an example.

almarklein:
Nice!

I merged your PR that adds the image.

I think this does not support writing, correct? Then it would be good to implement `def _can_write(self, request): return False`.

The plugin module should be imported in `imageio/plugins/__init__.py`. I am actually surprised that the tests run correctly right now :)

almarklein:
You also have a style error. Run `python make test flake` in the main repo to run style tests locally. (Its an extra space on line 19 of your test script.)

jni:
btw, `get_test_dir()` doesn't appear to work for in-place installs. After `pip install -e .` I tried to run pytest on my file and got:

```pytb
OSError: No such file: '/Users/jni/Library/Application Support/imageio/testdir/test_tiff.tiff'
```

when the project/file is actually in ~/projects/imageio...

almarklein:
> get_test_dir() doesn't appear to work for in-place installs

That's odd, the test dir is always in appdata (Application Support on OS X) and should not at all depend on how/where imageio is installed. Tested pip's develop install myself, and tests run fine (though I am not on OS X).

tests/test_fei_tiff.py
```diff
@@ -0,0 +1,40 @@
...
+    image_with_watermark = reader.get_data(0, discard_watermark=False)
+    assert image_with_watermark.shape == (1094, 1536)
+    assert np.round(np.mean(image_with_watermark)) == 137
+
+    assert reader.get_data(0, discard_watermark=False).shape == (1094, 1536)
+
+
+def test_fei_file_fail():
+    normal_tif = os.path.join(get_test_dir(), 'test_tiff.tiff')
+    bad_reader = imageio.get_reader(normal_tif, format='fei')
```
almarklein: You only define the filename here; you need to actually write the file :)

jni:
@almarklein whoops! My bad, I misread the pattern. I thought the test dir was supposed to point to a test data directory in the repo itself, as I have seen other packages do. I hope my latest commit fixes everything finally! =)

almarklein:
Instead of `np.testing.assert_raises` you can use `from pytest import raises` and then either:
```py
with raises(ValueError):
  ...

# or
assert raises(ValueError, ...)
```

jni:
@almarklein Is anything else holding up this PR now?

almarklein:
Everything looks good I think, except perhaps the position of import in `__init__.py`. Since this format by default strips rows from the image (the watermark) and cannot detect beforehand whether its a FEI image or a "regular" TIFF, let's put this one after the other pluigins that can read TIFF, i.e. behind freeimage and Pillow. I think people using this will select it explicitly anyway, so it should not matter in that respect.

almarklein:
I am going to go ahead and merge this. I can make that small change while I tweak for release.

Thanks @jni !!

jni:
Thanks! FYI you could also have pushed directly to my branch — this feature has been available since September: 

https://github.com/blog/2247-improving-collaboration-with-forks 

https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/

almarklein:
Right, I've never tried that yet. But its ok, I need to apply some tweaks anyway, so I can easily fit it in :)

PR #188, opened 2016-12-10
Use install_requires instead of requires

ghisvail:


almarklein:
? I read somewhere that it was the other way around. Ah well.

PR #190, opened 2016-12-10
extra_require -> extras_require

ghisvail:
http://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies


almarklein:
Sorry, my mistake. Thanks for the fix!

PR #191, opened 2016-12-11
Better handling of multiarch paths

ghisvail:
This commit enables support for multiarch paths for all architectures, instead of just `amd64`.

### DIFF
--- a/imageio/core/findlib.py
+++ b/imageio/core/findlib.py
@@ -51,10 +51,13 @@ def generate_candidate_libs(lib_names, lib_dirs=None):
     """
     lib_dirs = lib_dirs or []
     
+    # Multiarch triplet as defined in https://wiki.debian.org/Python/MultiArch
+    triplet = getattr(sys, 'implementation', sys)._multiarch
+    
     # Get system dirs to search
     sys_lib_dirs = ['/lib', 
                     '/usr/lib', 
-                    '/usr/lib/x86_64-linux-gnu',
+                    '/usr/lib/{}'.format(triplet),
                     '/usr/local/lib', 
                     '/opt/local/lib', ]
     


ghisvail:
The CI tests fail so it looks like this solution is Debian specific. So, I might as well just keep this patch downstream.

almarklein:
Still getting the hang of the new GH review system. I meant to just write a comment yesterday evening, apparently, it was not submitted yet because I had actually started a "review".

Anyway, a `try-except` around it would also be ok; then it would also work on Debian if `imageio` were pip-installed.

ghisvail:
The thing is that Debian's `ctype.loadlibrary` already handles multiarch paths so `findlib` will shadow this ability. My understanding is that `generate_candidate_libs` will automatically find a library regardless of its `soversion` (like for the `libpython` test which should work for Python 2.7 and 3.x). Am I correct?

almarklein:
The main purpose of `generate_candidate_libs()` is to be able to detect libs (like FreeImage) by looking in all the common places. Regardless of what was found, `load_lib()` will always first use `ctypes.cdll.LoadLibrary` with the lib name (not path) to let ctypes resolve it.

ghisvail:
Ok, so from a packaging standpoint, this is actually something I would want to bypass since binary packages are supposed to use libraries available in the system only. Would there be an easy to achieve this?

almarklein:
I guess we could add a module global to `findlib.py` that if `True` prevents `load_lib()` from calling `generate_candidate_libs()`. Then you can patch the module to set that global.

However, that would mean that the `avbin` plugin would not work on Debian. And other possible future plugins that cannot rely on system libs.

So maybe that global should be a list of lib names for which `load_lib()` should not call into `generate_candidate_libs()`. By default it would be `None`, but you can patch it to contain freeimage.

almarklein:
Mmm, it seems that the freeimage plugin is the only one that makes use of this functionality. So my argument only applies to potential new plugins. The plugin calls `load_lib` with the `exact_lib_names`  arg being `['FreeImage', 'libfreeimage.dylib', 'libfreeimage.so', 'libfreeimage.so.3']`.

ghisvail:
> And other possible future plugins that cannot rely on system libs.

Then, these dependencies would have to be packaged first. The lookup done in `generate_candidate_libs` is not a desirable feature for Debian because it may override usage of the system libraries in favor of potentially unsafe ones. If a user do insist on using a custom version, he or she can explicitly do so by pointing `LD_LIBRARY_PATH` to the desired location.

 > the freeimage plugin is the only one that makes use of this functionality.

That's one of such "unsafe" libraries. I have already patched a few CVEs for it on Debian (and I expect more to come in the future).

Based on your comments, I guess the only viable patching on my side is to prevent `load_lib` from using `generate_candidate_libs`.

almarklein:
> Based on your comments, I guess the only viable patching on my side is to prevent `load_lib` from using `generate_candidate_libs`.

If you have other suggestions, I am all ears, but making `load_lib` not use `generate_candidate_libs` seems relatively easy to do, and should solve your case, right? If you prefer to have a global flag that you can easily (and safely) patch, then I am open to that.

ghisvail:
The global (with an appropriate comment) would definitely be welcome.

almarklein:
Here you go #193

almarklein:
This can be closed, right? Since you simply don't use `generate_candidate_libs()` in the Debian variant of imageio?

ghisvail:
You are correct. Closing.

Issue #151, opened 2016-05-13
Debian packaging


ghisvail:
Dear imageio community.

I would like to resume my initial effort to package `imageio` for Debian, now that the project has gained some maturity. I intend to centralize all questions, concerns and feedback on the process in this issue, which I'll close once the packaging is achieved or dropped.

So far I have a working build but fail to run the test suite properly. Our builders do not face the internet so fetch remote sources is a no-no. I tried setting the `IMAGEIO_NO_INTERNET=1` envvar, but the testsuite fails with quite a few `InternetNotAllowedError` errors. Perhaps I am doing something wrong here.

Another concern is that the packaged `imageio` should be using the packaged dependencies for freeimage, ffmpeg and other libraries called via ctypes. I guess there is no other way but patch the source code directly, unless I am missing something.

Thanks for your assistance.

Ghis

almarklein:
On no internet: is there a point during the building where internet *is* allowed? There are a few image files which are considered "builtin", which are downloaded during installation: https://github.com/imageio/imageio/blob/master/setup.py#L123-L142  these are used during testing, and assumed to be accessible even if no internet is available.

IIRC, both freeimage and ffmpeg now use the system version if availeble. See e.g. https://github.com/imageio/imageio/blob/master/imageio/plugins/ffmpeg.py#L56-L65.

almarklein:
Noting #114, we might drop freeimage due to license issues and use `Pillow` instead.

ghisvail:
I am all in favour for it :+1:

Thanks for keeping me in the loop.

stefanv:
What is the current status of the Debian packaging?

ghisvail:
On hold, until #114 is addressed.

amarao:
Hi guys.

I've done some simple packaging for the debian: https://github.com/amarao/imageio

Checkout debian branch, update debian/changelog and run gpb buildpackage.

Not sure if I done everything properly or not. Build process goes include tests.

P.S. I can't create pull request to non-existent branch on github.

ghisvail:
I am happy to re-evaluate the submission of an official package to the Debian archive now that the problematic dependencies are gone. The work of @amarao would be a good starting point.

@almarklein any plans to tag / submit a new release sometime soon? With the freeze coming, anything submitted after Christmas would have very little chance to be included in the next release. Same comment for your other projects such as `pyzo`.

almarklein:
I'll release as soon as #182 is merged.

[...]

PR #192, opened 2016-12-12
Fix findlib test on Debian

almarklein:
@ghisvail this should fix the `test_findlib()` mentioned in #151


ghisvail:
I agree. That's what I would have suggested too.

PR #194, opened 2016-12-14
fix findlib test fpr  debian

almarklein:
cc @ghisvail This should fix that `findlib()` test.


ghisvail:
Slightly off-topic but, would testing of `generate_candidate_libs` still be run?

Because it would still fail if the test is run on anything but `amd64` due to the hardcoded multiarch path. I just want to know whether I would have to patch that or the tests just won't run it.

almarklein:
Yes, `test_findlib1()`  would still run, which does `generate_candidate_libs(['libc'])`.

PR #195, opened 2016-12-15
Ffmpeg read & write bug fixes

kuchi:
FFMPEG writer had 2 -r flags setting the frame rate. Removed 2nd one since it was overriding first one and it was outputting decimal rather than float causing incorrect frame rates to be created. Before this bug fix creating videos at frame rates like 29.97 fps would end up being 29 fps.

FFMPEG reader does not need two -ss flags to seek to specific time anymore. Now ffpmeg can do frame accurate seeking more efficiently with one call (see https://trac.ffmpeg.org/wiki/Seeking). I had to also increase the number of significant digits to seek time to ensure it accurately seeks but now we can directly go to the desired frame rather than seek to nearby frame and read forward to desired frame. Before this bug fix reading videos with frame rates like 30 fps and seeking to a specific frame often at times off by one especially when going backwards in index.

### DIFF
--- a/imageio/plugins/ffmpeg.py
+++ b/imageio/plugins/ffmpeg.py
@@ -333,7 +333,7 @@ class FfmpegFormat(Format):
             elif index >= self._meta['nframes']:
                 raise IndexError('Reached end of video')
             else:
-                if (index < self._pos) or (index > self._pos+100):
+                if (index < self._pos) or (index > self._pos+100): # May want to reduce from 100
                     self._reinitialize(index)
                 else:
                     self._skip_frames(index-self._pos-1)
@@ -381,9 +381,10 @@ class FfmpegFormat(Format):
                 offset = min(1, starttime)
 
                 # Create input args -> start time
-                iargs = ['-ss', "%.03f" % (starttime-offset),
+                # Need minimum 6 significant digits for high fps videos
+                iargs = ['-ss', "%.06f" % (starttime),
                          '-i', self._filename,
-                         '-ss', "%.03f" % offset]
+                         ]
 
                 # Output args, for writing to pipe
                 oargs = ['-f', 'image2pipe',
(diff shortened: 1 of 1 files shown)


almarklein:
> Now ffpmeg can do frame accurate seeking more efficiently with one call

What does *now* mean? The page you linked seems to suggest v2.1. Would that mean that the versions that we provide (which are about 2 years old, from the looks of it) work with this feature, or do we need to update them?

kuchi:
Sorry for not being clearer. Yes, from what the FFMPEG documentation page
says, frame accurate seeking with one call to -ss should work starting in
v2.1. On my system it looks like imageio is using 2.6.3, and it works fine
with the single call in my testing. So there is no need for an update from
what you currently have.

There is one issue i've noticed that I haven't been able to track down yet.
If you play a video in reverse, after about ~200 frames it hangs on my
system (even before this pull request). So I'm not sure if there is some
issue of calling _reinitialize many times that something is going wrong.
Don't know if you have encountered this. If I get time I can try and make a
test to illustrate this failure.

One other question, what do you think about having the ability to get
frames by indexing? This might make the returned object more pythonic, as in
   mov = imageio.get_reader("my_video.mp4")
   frame = mov[0]     # equivalent to mov.get_data(0)
   frame = mov[1]

Then you could support:
   for frame in mov:
          pass

Would likely be an easy addition for multi image objects. What are your
thoughts?

almarklein:
Thanks for the info!

> If you play a video in reverse, [...] Don't know if you have encountered this.

Haha, no, I've never tried that :)  I can imagine this is a use case that moviepy users might encounter, but I have heard nothing of it thus far. Nevertheless, would be good to look into at some point, yes.

> what do you think about having the ability to get frames by indexing? 

I think I refrained from adding this thus far, because for each frame there is both data and metadata, plus the `get_data()` method possibly accepts more args. But seeing the code like that looks nice; indexing could be an alias for `get_data(i)`. So I would not be against it.

Readers are already iterable though, so `for frame in mov:` has been working for a while.

almarklein:
Is this PR ready, BTW?

kuchi:
Yes, it's ready. :-)

almarklein:
Thanks!

kuchi:
Thanks to you Alamar for all the awesome work on the library. Can't thank you enough for an excellent engineered set of code. 

Anthony

almarklein:
Its seeing how other people like you build awesome things on top of it that makes it all worthwhile.

PR #197, opened 2016-12-19
sdist also include images

almarklein:


ghisvail:
That would be great indeed. Once this is merged, it will be easier to generate a test tarball for the current Debian packaging.

almarklein:
Do you want me to push a release candidate version to pypi, or can you help yourself with this fix?

ghisvail:
I can help myself, don't worry.

Issue #212, opened 2017-01-24
OSError: [Errno 24] Too many open files When calling get_data in reverse order


ekcheng:
When calling `get_data` in reverse order on a video file, ffmpeg processes fail to close properly, causing an `OSError: [Errno 24] Too many open files` error.

Here is some simple code that causes the error in v 2.1.1:

```
import imageio
mov = imageio.get_reader('movie.mp4')
N = mov.get_length()
for i in range(N):
    im = mov.get_data(N-1-i)
```

If the movie is long enough, this will cause the error. But there are no problems reading forwards, from frame 0 to the end.

Appears to be related to #145.

ekcheng:
I see now that this is maybe because `get_data` reinitializes itself when you are decreasing the position per these lines in ffmpeg.py. It looks like maybe `reinitialize` is not closing out existing ffmpeg processes properly:

```
if index == self._pos:
                return self._lastread, {}
            elif index < 0:
                raise IndexError('Frame index must be > 0')
            elif index >= self._meta['nframes']:
                raise IndexError('Reached end of video')
            else:
                if (index < self._pos) or (index > self._pos+100):
                    self._reinitialize(index)
                else:
                    self._skip_frames(index-self._pos-1)
                result = self._read_frame()
                self._pos = index
                return result, {}
```

ekcheng:
Note I have observed this problem both on Ubuntu 16.04 and macOS Sierra (10.12.2).

almarklein:
Thanks for reporting. This is a bit odd, since `_reinitialize()` calls `_close()`, which calls `_terminate()`. The latter ought to be *the* way to safely close the ffmpeg process. It's been though a number of iterations though ...

Can you make sure that you're using the latest version of imageio (2.1.1)? Perhaps you can try fiddling with `_terminate()` to make it work, like uncommenting that `kill` call. Otherwise, @kuchi or me will look into it when we have some time to spare.

almarklein:
Closed by #213

ekcheng:
Thanks! When is the next release planned?

almarklein:
I think we should do one pretty soon, since we fixed a few important bugs. There's one or two open issues, which I hope we can include, so perhaps next week.

PR #213, opened 2017-01-25
fix for ffmpeg not properly terminating

kuchi:
Fix to completely terminate ffmpeg using proc.kill() since proc.terminate() does not seem to work for ffmpeg. Should address #212 
All ffmpeg tests still pass on my system (OS X) so hopefully this is a fix that works on all platforms.
This fix also substantially improves the performance of the ffmpeg reader when having to reinitialize since not there is no wait for the process to die.

### DIFF
--- a/imageio/plugins/ffmpeg.py
+++ b/imageio/plugins/ffmpeg.py
@@ -425,7 +425,6 @@ class FfmpegFormat(Format):
                 self._frame_catcher.stop_me()
 
 
-
         def _load_infos(self):
             """ reads the FFMPEG info on the file and sets size fps
             duration and nframes. """
--- a/imageio/plugins/ffmpeg.py
+++ b/imageio/plugins/ffmpeg.py
@@ -424,7 +424,6 @@ class FfmpegFormat(Format):
             if self._frame_catcher:
                 self._frame_catcher.stop_me()
 
-
         def _load_infos(self):
             """ reads the FFMPEG info on the file and sets size fps
             duration and nframes. """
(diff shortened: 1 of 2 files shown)


almarklein:
Isn't it a little rough to kill without trying "politely" first?

kuchi:
@almarklein, on the old way it tried politely and then had delays to wait to see if it was killed but that didn't seem to be working and it really slowed things down and seemed to keep holding system resources. I don't think FFMPEG was ever getting killed until the main python process ended. The subprocess documentation says kill is just an alias to terminate on windows, but on POSIX the SIGKILL is sent. So this should fix issues on POSIX systems, but I don't know if there was still an issue on windows.

Try out the new test I added, you'll need to run `py.test -k test_reverse_read -v -s` to see where it hangs if it does on your with the old version of the ffmpeg plugin. When you run it on the new one you can also see that it is much faster.

I don't have any additional changes so if you are good with this fix you can use it. But if you have a nicer way to fix it I'm open to it. I would like to see a fix that doesn't have so many delays which cause _reinitialize to be slowed down.

kuchi:
@almarklein  - I fixed the flake8 issues I had and added back stuff I didn't indent to delete `catcher.stop_me()`. Let me know if you think this is a workable fix or if we can do it a better way. But I'm good with it now if you want to use it.

almarklein:
I confirmed that things were already fine on Windows, and this change does not change that, which makes sense, because `p.kill()` is an alias for `p.terminate()` on Windows.

I am/was slightly worried about using SIG_KILL because it does not give ffmpeg a chance to shut down properly. But since it probably does not use temporary files or anything, its probably ok. Also, it seemed that asking it to terminate did not really work, so we'd be killing it anyway if we wanted to solve #212.

So let's go ahead with this. Thanks!

PR #224, opened 2017-02-09
disable looping in gifs functionality

arashrai:
With this I'm able to create a gif using something like:

```
imageio.mimsave(name, numpys, duration=0.25,loop=-1)
```
And the created gif will now play exactly once.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ branches:
 
 matrix:
     include:
-        - python: "3.5"
+        - python: "3.6"
           env: TEST_UNIT=0 TEST_STYLE=1
         #
         - python: "2.6"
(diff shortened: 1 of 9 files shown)


almarklein:
What's wrong with setting loop to 1, which should loop exactly once (i.e. no looping).

Works for me, at least when I play it in Firefox. When I play it in de default Windows 10 photo viewer, it seems to always repeat no matter what value of loop, or applying your change.

arashrai:
That's strange. Firefox seems to be doing some optimization here since who really wants to play a gif twice. Can you please check on chrome?

almarklein:
Tested in Chrome, and to my surprise, the animation was shown twice. It looks like the looping parameter is seen by some readers (e.g. Firefox) as the number of times that the animation should be shown, and by others (e.g. Chrome) how many times the animation should be restarted. I have not yet been able to find what it *should* be.

My inclination is to leave out the application extentsion when `loop = 1` (instead of -1). What do you think?

arashrai:
That works for me. The reasoning for the -1 implementation was me not wanting to break anyones code that actually wanted the double loop of gifs in chrome. It's probably a non-issue.

almarklein:
We were having problems with the tests (#231). That is now fixed. Can you please rebase your PR (let me know if you need help with that)?

almarklein:
Supersceded by #237

PR #237, opened 2017-03-03
make gifs loop once on chrome

arashrai:
With this I'm able to create a gif using something like:

```imageio.mimsave(name, numpys, duration=0.25,loop=1)```

And the created gif will now play exactly once on chrome. Currently gifs created with the same code run twice on chrome because chrome interrupts the NETSCAPE2.0 loop count as the number of times to loop, rather than the number of times to play.


almarklein:
Thank you!

Issue #242, opened 2017-04-05
Saving empty animated image


PeterQLee:
Saving an animated image with mimwrite or mimsave with an empty list will write a file with a single byte.

For example:

```
import imageio
imageio.mimwrite('a.gif',[])
```

If you check the contents of a.gif after running, it contains a single 0x3b byte.
This doesn't seem like the intended behaviour. Maybe it should throw an error, or at least give a warning to the user?

almarklein:
There could be formats where its "legal" to have a sequence of 0 images. Though I don't a single-byte gif can be a valid GIF file :)

PR #247, opened 2017-04-12
Fix for Issue #242, saving empty animated image

PeterQLee:
Fix for #242,  simple length check on the input sequence. 

I've never heard of a gif file with no data, but I think if someone wanted that, they could make their own file trivially 😏.


almarklein:
Seems reasonable enough :)

almarklein:
Thanks!

PeterQLee:
My pleasure!

PR #255, opened 2017-05-21
Fix icorrect tiff dimensions

### LINKED ISSUE #226: Incorrect image dimensions after reading in 3D+c TIFF
I'm trying to read the following file:

https://www.dropbox.com/s/0ynbscx4cmd5k91/E_z2_512_1um_CONTROL.tif?dl=1

skimage (ie tifffile) reads this correctly, but imageio doesn't — even though I thought it also used tifffile as a backend for tiffs?

```python
In [1]: import imageio
In [2]: from skimage import io
In [3]: im_iio = imageio.imread('E_z2_512_1um_CONTROL.tif')
In [4]: im_iio.shape
Out[4]: (159, 320, 512)

In [5]: im_ski = io.imread('E_z2_512_1um_CONTROL.tif')
In [6]: im_ski.shape
Out[6]: (53, 320, 512, 3)
```

What imageio is doing is reading the file in the array order present in the file (plane, channel, row, column), and squashing pln and ch dimensions together, while skimage is behaving as expected and producing two separate dimensions for pln and ch and (maybe a bit more than expected) moving ch to the last dimension, which is the dimension order expected by most scipy packages.

Any ideas about what is going wrong here?

almarklein:
Fixes #226 (hopefully)

cc @jni @blink1073 

Does this look like a proper solution? What about the meta data?

### DIFF
--- a/imageio/plugins/tifffile.py
+++ b/imageio/plugins/tifffile.py
@@ -177,20 +177,28 @@ class TiffFormat(Format):
 
         def _close(self):
             self._tf.close()
-
+        
         def _get_length(self):
-            return len(self._tf)
-
-        def _get_data(self, index):
-            # Get data
-            if index < 0 or index >= len(self._tf):
-                raise IndexError(
-                    'Index out of range while reading from tiff file')
-            if index == 0 and 'i' in self.request.mode:
-                im = self._tf.asarray()  # request as singleton image
+            if self.request.mode[1] in 'vV':
+                return 1  # or can there be pages in pages or something?
             else:
+                return len(self._tf)
+        
+        def _get_data(self, index):
+            if self.request.mode[1] in 'vV':
+                # Read data as single 3D (+ color channels) array
+                if index != 0:
+                    raise IndexError(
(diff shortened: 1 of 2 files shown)


jni:
@almarklein 

```python
In [1]: import imageio as iio
In [2]: im = iio.imread('E_z2_512_1um_CONTROL.tif')
In [3]: im.shape
Out[3]: (53, 3, 320, 512)
```
🎉 

Thanks for the fix!

jni:
LOL Love the commit message btw. =D

almarklein:
Ok, I changed this a bit by making `imageio.volread()` return the multi-page data as a whole. Although "volume" might not be the right term in the context of TIFF files, the case that `volread()` exists for very closely matches this use-case. Plus it does not break backward compatibility.

So for @jni it should be `im = iio.imread('E_z2_512_1um_CONTROL.tif')`.

jni:
@almarklein I suppose I can live with this. =)

blink1073:
Seems reasonable to me.

almarklein:
Great. I added docs to explain the new behavior.

PR #267, opened 2017-06-22
Fixed #265: Override the default user directory using an IMAGEIO_USERDIR env var

### LINKED ISSUE #265: Allow userDir to be configurable to account for more restrictive environments like AWS Lambda
Its seems that AWS Lambda only allows writes to `/tmp`. When imageio tries to download ffmpeg, it first tries `~`, then `/var/tmp` and finally fails because neither are writeable ([relevant code](https://github.com/imageio/imageio/blob/master/imageio/core/util.py#L441) -- also see https://github.com/imageio/imageio/issues/54).

Instead of trying yet another directory (`/tmp`), it would be handy if we were able to configure this using an environment variable, say `IMAGEIO_USERDIR`.

kappataumu:
Fixes #265


almarklein:
Thanks!

PR #268, opened 2017-06-22
Prevent ffmpeg Window popup in frozen application (Windows)

### LINKED ISSUE #266: Add `shell=True` to `subprocess.Popen` calls (Pyinstaller)
I can confirm that imageio+ffmpeg works with Pyinstaller on Windows and Anaconda Python 2.7 (#32).
However, when using the [ffmpeg](https://github.com/imageio/imageio/blob/master/imageio/plugins/ffmpeg.py) plugin in a frozen application on Windows, and when seeking through an avi video (`imageio.get_reader`) using `get_data`, an empty command Window pops up with the title ("path/to/ffmpeg-win32-v3.2.4.exe"). These command windows are not shown when the keyword argument `shell=True` is set in `sp.Popen`: https://github.com/paulmueller/imageio/commit/b5aeb617fb6b1a74ff30c40ae998bac31b526f39

If this does not break compatibility, I would recommend adding that extra keyword argument everywhere. I can create a pull request if you are too busy yourself.

paulmueller:
fixes #266

### DIFF
--- a/imageio/plugins/ffmpeg.py
+++ b/imageio/plugins/ffmpeg.py
@@ -28,6 +28,9 @@ from ..core import (Format, get_remote_file, string_types, read_n_bytes,
                     image_as_uint, get_platform, CannotReadFrameError,
                     InternetNotAllowedError, NeedDownloadError)
 
+
+ISWIN = sys.platform.startswith('win')
+
 FNAME_PER_PLATFORM = {
     'osx32': 'ffmpeg-osx-v3.2.4',
     'osx64': 'ffmpeg-osx-v3.2.4',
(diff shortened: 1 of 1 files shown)


imageio/plugins/ffmpeg.py
```diff
@@ -250,8 +250,10 @@ def _get_cam_inputname(self, index):
                 # Ask ffmpeg for list of dshow device names
                 cmd = [self._exe, '-list_devices', 'true',
                        '-f', CAM_FORMAT, '-i', 'dummy']
+                # Set `shell=True` in sp.Popen to prevent popup of a command
+                # line window in frozen applications (Windows only).
                 proc = sp.Popen(cmd, stdin=sp.PIPE, stdout=sp.PIPE,
-                                stderr=sp.PIPE)
+                                stderr=sp.PIPE, shell=True)
```
almarklein: Did you omit `shell=is_win` here for a reason?

Also, please define `is_win` just once, near the top of the module.

paulmueller: Yes, because it already is in the case `elif sys.platform.startswith('win')`.

paulmueller: (this is not visible in the diff)

almarklein:
Looking good. Thanks!

Issue #32, opened 2014-11-07
Support for frozen apps


almarklein:
We would like imageio to work well in frozen apps.

For now, frozen apps with imageio should just work, but required binaries are still downloaded at runtime. This requires an internet connection, so it may be good to provide a way to _ship_ the binary dependencies.

For this we need to provide a way for the developer to copy all required binaries to the app directory (so they can be shipped along) and we need to _look_ in that directory too.

paulmueller:
I propose the following workflow:
1. Implement an [entry point](http://python-packaging.readthedocs.io/en/latest/command-line-scripts.html#the-console-scripts-entry-point) named e.g. `imageio_download_binary` for downloading the required binary.
2. Add the option `--package-dir` to the entry point which will download the binary to the `plugins` directory of the installed package. This directory can be determined with [`pkg_resources.recource_filename`](http://setuptools.readthedocs.io/en/latest/pkg_resources.html#resource-extraction) (which also works with eggs), e.g. `pkg_resources.resource_filename("imageio.plugins", imageio.plugins.ffmpeg.FNAME_PER_PLATFORM[imageio.plugins.ffmpeg.get_platform()])`.
3. Give the binary file in the plugins directory highest priority when calling the `get_exe` method.
4. For Pyinstaller, only a simple hooks script is required:
   ```python
   # hook-imageio.py
   from PyInstaller.utils.hooks import collect_data_files
   datas = collect_data_files('imageio')
   ```
   (needs pull request for PyInstaller)

This workflow as these advantages:
1. It does not break the current behaviour
2. Not many changes to the sources of imageio are required
2. It allows to download binaries with a command line script
3. On CIs such as appveyor or travisCI, it allows to easily build PyInstaller distributables with only the required binaries

almarklein:
Sounds like a plan. How do we deal with the fact that there are multiple binaries? Would you do e.g.
`imageio_download_binary ffmpeg` ?

paulmueller:
Yes, that would be the usage of the entry point script. Did I oversee anything? I could start with a PR.

almarklein:
As far as I can see, this should work. Looking forward to a PR!

paulmueller:
I have a question regarding the `_set_platform_resources` method in `setup.py` (used in `build_with_fi`). What is the use-case of this method? Also, there are two lines dealing with ffmpeg commented out in `_set_platform_resources`?

almarklein:
[...]

PR #270, opened 2017-07-24
Add console entry point script for binary download

paulmueller:
Please do not merge yet. There are a few points open for discussion.

Here is a proposal for an entry point script for binary download (related to issue #32)

The new entry point script can be seen in action in this appveyor build (line 246):
https://ci.appveyor.com/project/paulmueller/ShapeOut/build/1.0.688/job/0mo8iu4xxs6on968#L246

There are a few things that need to be clarified before merging:
1. Are you fine with a separate `scripts.py` module or would you like to have this functionality someplace else?
2. To allow the option to download all available binaries, I need a method that gets all plugins with binary files. One possibility is to check all submodules in `plugins` for the `download` method. Is there a better way? 
3. I would like to add `*args` and `**kwargs` to the `download` methods for ffmpeg, avbin, and freeimage  to be able force-download the binary files into the `recources` directory (see #32). Would that be an acceptable change?
4. Could you please point me to the relevant parts of the documentation that require an update due to this PR?


almarklein:
Thanks for this!

> Are you fine with a separate scripts.py module or would you like to have this functionality someplace else?

I like `__main__.py` for this, since it also allows invoking the entry point through `python -m imageio ...`. E.g. for cases where the package is not installed (e.g. dev env), or to explicitly specify the Python interpreter.

> To allow the option to download all available binaries, I need a method that gets all plugins with binary files. One possibility is to check all submodules in plugins for the download method. Is there a better way?

Mmm, I think it should be enough to specify a list of plugin names at the top of the script. Checking plugins for a download method might break when a plugin implements a download method (for whatever reason) that does not do what the script expects it to do. There's not a lot of plugins that need downloads, so it should not be hard to maintain this. If needed, we could look at automating this, but then use a more specific function name.

> I would like to add `*args` and `**kwargs` to the download methods for ffmpeg, avbin, and freeimage to be able force-download the binary files into the recources directory (see #32). Would that be an acceptable change?

I think I'd prefer a keyword argument (e.g. `def download(force_download=False)`) and then pass that to where its needed.

> Could you please point me to the relevant parts of the documentation that require an update due to this PR?

* I think we need a new page, i.e. a subling of https://github.com/imageio/imageio/blob/master/docs/envvariables.rst
* Add it in the index here: https://github.com/imageio/imageio/blob/master/docs/sec_reference.rst, I think right before the section on environment variables.
* The docs in the ffmpeg and freeimage plugin should mention both the CLI  and `imageio.plugins.xx.download()` as a way to download the binaries. They still seem to refer to an automated download, which is no longer the case.

paulmueller:
Ok, I addressed all your points except for the docs.
I had to edit the [ffmpeg.get_exe](https://github.com/imageio/imageio/pull/270/files#diff-d72802e02eea93d5f9fa4aa279eeaef2L61) method so that the imageio binaries get highest priority (this is especially important when shipping a frozen application). Please have a look and let me know if that is ok.

paulmueller:
Update: Creating a distribution with PyInstaller works.
Here is the pull request with the hook script:
https://github.com/pyinstaller/pyinstaller/pull/2696

imageio/core/util.py
```diff
@@ -495,8 +497,10 @@ def resource_dirs():
     in the future.
     """
     dirs = []
-    # Resource dir baked in the package
-    dirs.append(os.path.abspath(os.path.join(THIS_DIR, '..', 'resources')))
+    # Resource dir baked in the package.
+    # (The directory returned by `pkg_resources.resource_filename`
+    # also works with eggs.)
+    dirs.append(pkg_resources.resource_filename("imageio", "resources"))
```
almarklein: `pkg_resources` comes with setuptools, right? Can we please make it optional? I want to avoid cases like [this](https://stackoverflow.com/questions/7446187/no-module-named-pkg-resources).

E..g.
```py
try:
   import pkg_resources
except ImportError:
    pkg_resources = None

...

if pkg_resources:
    use pkg_resources
else:
   use the old line
```

imageio/plugins/ffmpeg.py
```diff
@@ -91,33 +115,39 @@ def get_exe():
         try:
             with open(os.devnull, "w") as null:
                 sp.check_call([exe, "-version"], stdout=null, stderr=sp.STDOUT)
-                return exe
+                exes.append(exe)
```
almarklein: I guess this should be `insert`?

Maybe its easiest to just append to `exes`, in the order of priority, and add `'ffmpeg'` last. 

I think it would be ok to give highest priority to the one provided by imageio. But it would be good to have a way to remove the binary to have more control, e.g. in case its corrupt. Can you add that to the CLI?

paulmueller: Sure, this can be `insert` to give it priority over the  system's ffmpeg.

If we want to append the exes in order of priority, then we would have to refactor the case in lines 122+.

Sure, I can add a `--remove` switch to the entry point.

paulmueller: see https://github.com/imageio/imageio/pull/270/commits/40655463b0c215cd5c4a9d114c2795d3dea0e5e1: I refactored out `resource_package_dir` which is also used in `__main__`.

paulmueller: see https://github.com/imageio/imageio/pull/270/commits/956955ef187aeed6fe3f6839b41c9324e4c105cd: I added the `imageio_remove_bin` entry point.

paulmueller:
Please have a look. If you have nothing to add, this could be merged now.

imageio/__main__.py
```diff
@@ -0,0 +1,137 @@
...
+        # directory will be created by pkg_resources (requires setuptools).
+        # see `imageio.core.util.resource_dirs`
+        # and `imageio.core.utilresource_package_dir`
+        directory = util.resource_package_dir()
+    else:
+        directory = None
+    
+    for plg in plugin_names:
+        msg = "Plugin {} is not registered for binary download!".format(plg)
+        assert plg in PLUGINS_WITH_BINARIES, msg
```
almarklein: Can you make this a test + raise, assert statements are removed from optimized bytecode.

imageio/__main__.py
```diff
@@ -0,0 +1,137 @@
...
+        # Use all plugins
+        plugin_names = PLUGINS_WITH_BINARIES
+
+    print("Removing binaries for: {}.".format(", ".join(plugin_names)))
+
+    rdirs = util.resource_dirs()
+
+    for plg in plugin_names:
+        msg = "Plugin {} is not registered for binary download!".format(plg)
+        assert plg in PLUGINS_WITH_BINARIES, msg
```
almarklein: Dito

imageio/__main__.py
```diff
@@ -0,0 +1,137 @@
...
+    
+    for rd in rdirs:
+        # plugin name is in subdirectories
+        for rsub in os.listdir(rd):
+            if rsub in plugin_names:
+                shutil.rmtree(op.join(rd, rsub))
```
almarklein: Maybe have a try-except around this? When the lib is still loaded, a remove may fail, preventing the removal for other dirs. Perhaps raise error at the end when any of the deletions failed so that the process returns -1.

imageio/plugins/ffmpeg.py
```diff
@@ -66,22 +82,16 @@ def get_exe():
...
+
+    # To give the executable provided by imageio highest priority,
+    # generate a list of executables and allow to add executables 
+    # to the beginning or the end according to priority.
+    # The initial string "ffmpeg" is a fallback in case nothing is found.
+    exes = ["ffmpeg"]
     
     # Check if ffmpeg is installed in Python environment
     # (e.g. via conda install ffmpeg -c conda-forge)
-    exe = None
```
almarklein: I'm not yet happy with this part :) E.g. exes is always nonempty, but at the end we test whether it is.

I think it would be better to forget about that list and just try each option in order of priority, returning the value if it was found to be correct.

* If `IMAGEIO_FFMPEG_EXE` is set, use that (even without testing whether it is a valid path).
* Try our own version using `get_remote_file()`. (if `NeedDownloadError` is raised, just continue to next step)
* Try from conda package
* Try `ffmpeg` command
* Nothing found, raise `NeedDownloadError` with nice message.

imageio/plugins/ffmpeg.py
```diff
@@ -143,11 +158,15 @@ class FfmpegFormat(Format):
     Note that for reading regular video files, the avbin plugin is more
     efficient.
     
+    The ffmpeg plugin requires an `ffmpeg` binary. If this binary is
+    not available on the system, it can be downloaded by either
```
almarklein: Would be nice to document the places where imageio tries to load ffmpeg from.

almarklein: Well, it sort of is already, but maybe making the priority more explicit.

almarklein:
I apologize for the late reply. I was away on holiday and then a conference, and it disappeared from my radar (i.e. I probably accidentally marked the notification email as unread).

This is looking really good! I made a few small comments, and a slightly larger one regarding ffmpeg's download sequence.

A test would be great, also to prevent the coverage to drop too much. It might be a bit hard to fully test it fully, but some unit tests that cover most of the new code would be great.

imageio/__main__.py
```diff
@@ -0,0 +1,137 @@
...
+        epilog=example_text,
+        formatter_class=argparse.RawDescriptionHelpFormatter)
+    parser.add_argument("plugin", type=str, nargs="*", default="all",
+                        help=phelp)
+    args = parser.parse_args()
+    remove_bin(plugin_names=args.plugin)
+
+
+if __name__ == "__main__":
+    download_bin_main()
```
almarklein: For testing and dev, it might be nice to support `python -m imageio download_bin` and `python -m imageio remove_bin`. As it is now, I think one could call `python -m imageio` and it would default to `download_bin_main()`. In short something along these lines should work:
```py
if __name__ == "__main__":  # technically always True for a __main__.py
    if len(sys.argv) > 1 and sys.argv[1] == 'download_bin':
        download_bin_main()
    elif len(sys.argv) > 1 and sys.argv[1] == 'remove_bin':
        remove_bin_main()
    else:
        raise RuntimeError('Invalid use of the imageio CLI')
```

paulmueller:
I covered everything you mentioned and added a few tests.

If you are ok with it, I would not like to implement compatibility for binary download in Python 2.6.
The problem is similar to this one: https://stackoverflow.com/questions/19668395/str-format-for-python-2-6-gives-error-where-2-7-does-not

I think it makes the code unreadable if I added zeros at every use of "format" and I think no one should be using Python2.6 anyway. Do you agree?

see https://github.com/paulmueller/imageio/blob/406a2166a627e81a0b78ceb66be0d7d1e79332ed/tests/test_bin.py

almarklein:
> If you are ok with it, I would not like to implement compatibility for binary download in Python 2.6.

Fair enough.

> I think no one should be using Python2.6 anyway. Do you agree?

Hell yeah :)

This is looking good. Is it ready from your end?

paulmueller:
image

almarklein:
Thank you!

PR #272, opened 2017-07-28
Fixed doc of get_writer.

jusjusjus:


almarklein:
Wow, the fact that this has remained unnoticed is almost scary. Thanks!

PR #282, opened 2017-09-20
Increase the timeout interval to 10 seconds

imriss:
On some VMs with low amount of resources, the timeout of 4 seconds might not be enough, and this could result in unstable operation.


almarklein:
Thanks. Sorry for the late response!

PR #284, opened 2017-09-22
add basic test for issue 283, and modify device-name parser to fix this issue

dennisvang:
The device list parsing code was moved into a separate (static) method to facilitate testing without actually having any webcams. 
The corresponding test requires a string containing an ffmpeg output sample (i.e. the content of the `infos` variable from `_get_cam_inputname()`). For now this is hardcoded into the test, but I would get it from a file instead.

### DIFF
--- a/tests/test_ffmpeg.py
+++ b/tests/test_ffmpeg.py
@@ -371,13 +371,40 @@ def test_webcam():
     except Exception:
         skip('no web cam')
 
+
+# Sample from ffmpeg -device-list (distilled down to two lines per section)
+ffmpeg_output_sample = \
+    u'ffmpeg version 3.2.4 Copyright (c) 2000-2017 the FFmpeg developers\r\n' \
+    u'  built with gcc 6.3.0 (GCC)\r\n' \
+    u'  configuration: --enable-gpl --enable-version3 --enable-d3d11va' \
+    u' --enable-dxva2 --enable-libmfx --enable-nvenc --enable-avisynth' \
+    u'libswresample   2.  3.100 /  2.  3.100\r\n  ' \
+    u'libpostproc    54.  1.100 / 54.  1.100\r\n' \
+    u'[dshow @ 039a7e20] DirectShow video devices (some may be both video ' \
+    u'and audio devices)\r\n' \
+    u'[dshow @ 039a7e20]  "AVerMedia USB Polaris Analog Capture"\r\n' \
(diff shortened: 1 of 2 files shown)


tests/test_ffmpeg.py
```diff
@@ -371,6 +371,22 @@ def test_webcam():
...
...ae1&1&0000#{65e8773d-8f56-11d0-a3b9-00a0c9223196}\\global"\r\n[dshow @ 039a7e20] DirectShow audio devices\r\n[dshow @ 039a7e20]  "Microphone (2- USB Multimedia Audio Device)"\r\n[dshow @ 039a7e20]     Alternative name "@device_cm_{33D9A762-90C8-11D0-BD43-00A0C911CE86}\\wave_{73C17834-AA57-4CA1-847A-6BBEB1E0F2E6}"\r\n[dshow @ 039a7e20]  "SPDIF Interface (2- USB Multimedia Audio Device)"\r\n[dshow @ 039a7e20]     Alternative name "@device_cm_{33D9A762-90C8-11D0-BD43-00A0C911CE86}\\wave_{617B63FB-CFC0-4D10-AE30-42A66CAF6A4E}"\r\n[dshow @ 039a7e20]  "Microphone Array (Realtek High Definition Audio)"\r\n[dshow @ 039a7e20]     Alternative name "@device_cm_{33D9A762-90C8-11D0-BD43-00A0C911CE86}\\wave_{8A2CC322-DC22-4F8B-BC5A-DB8AF5DB844C}"\r\n[dshow @ 039a7e20]  "Analog Audio In (AVerMedia C039 USB Capture Card)"\r\n[dshow @ 039a7e20]     Alternative name "@device_cm_{33D9A762-90C8-11D0-BD43-00A0C911CE86}\\wave_{C160E7BB-863A-4C5E-81D7-C8B68627D6CD}"\r\n[dshow @ 039a7e20]  "Line (2- USB Multimedia Audio Device)"\r\n[dshow @ 039a7e20]     Alternative name "@device_cm_{33D9A762-90C8-11D0-BD43-00A0C911CE86}\\wave_{F80C3538-68FA-4816-B400-44CB19336269}"\r\ndummy: Immediate exit requested\r\n'
```
almarklein: This is ok by me, I'd even say to put this in a multiline string so it is readable. Perhaps move it to just above the function to avoid indentation issues.

dennisvang: Ok, I moved the string outside the function as a multiline. Reduced its size by removing some irrelevant parts, but it's still quite a lot of lines.

almarklein:
Thanks for this!

PR #301, opened 2018-01-12
Feature/lytro lfr plugin

MaxSchambach:
# Description

This PR adds a plugin to read Lytro ``.lfr`` and ``.raw`` files as provided by the [Lytro Illum](https://en.wikipedia.org/wiki/Lytro#Lytro_ILLUM) light field camera.

The ``.lfr`` format is a container format containing raw 10bit image data and metadata.
The ``.raw`` format is a 10bit raw image format.

# Why is this useful?

The Lytro Illum is a light field camera that aimed for consumer market but is now mostly used in the scientific community as a reference implementation of a light field camera. Lytro provides closed-source software to extract and decode the light field images from the camera.

For scientific use, reading the raw ``.lfr`` and ``.raw`` format is necessary, e.g. to test and compare new decoding algorithms.

As of now, there is no public Python module capable of reading the ``.lfr`` and ``.raw`` data provided by the Lytro Illum camera. The most videly spread (open-source) software to read and decode lightfields is the [MATLAB Light Field Toolbox](https://de.mathworks.com/matlabcentral/fileexchange/49683-light-field-toolbox-v0-4) by Donald Dansereau.

# Type of change

- [x] Add Lytro lfr and raw plugin (``plugins/lytro.py``)
- [x] Add plugin loading to ``plugins/__init__.py``

# What about testing?

I'm struggling with writing a test for the plugin and I am unsure how to go about this.
First problem:
The ``.lfr`` and ``.raw`` files of the Lytro Illum camera are about 50mb in size. Adding them to the ``imageio-binaries/image`` project would increase its size drastically.

Second:
How would I go about testing the correct behavior of the plugin as there is no data to compare the read images with? Is there more to test then that there are no errors when reading an example file?

# Checklist:

- [x] Code follows PEP guidellines
- [x] Add unit test for the lytro plugin

### DIFF
--- a/imageio/plugins/__init__.py
+++ b/imageio/plugins/__init__.py
@@ -106,6 +106,8 @@ from . import fits  # depends on astropy
 from . import simpleitk  # depends on SimpleITK
 from . import gdal  # depends on gdal
 
+from . import lytro
+
 from . import example
 
 # Sort
--- a/imageio/plugins/lytro.py
+++ b/imageio/plugins/lytro.py
@@ -391,15 +391,15 @@ class LytroLfrFormat(LytroFormat):
 
 
 # Create the formats
-SPECIAL_CLASSES = {'lytro-raw': LytroRawFormat,
-                   'lytro-lfr': LytroLfrFormat
+SPECIAL_CLASSES = {'lytro-lfr': LytroLfrFormat,
+                    'lytro-raw': LytroRawFormat
                    }
 
 # Supported Formats.
 # Only single image files supported.
 file_formats = [
-    ('LYTRO-RAW', 'Lytro Illum raw image file', 'raw', 'i'),
-    ('LYTRO-LFR', 'Lytro Illum lfr image file', 'lfr', 'i')
+    ('LYTRO-LFR', 'Lytro Illum lfr image file', 'lfr', 'i'),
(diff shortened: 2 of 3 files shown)


almarklein:
Nice addition!

About testing. Is there a way to crop the image to reduce its size (I suspect not, since the size seems hard-coded?)  If not, let's just add it to imageio/binaries. We made that a separate repo so that we can add large binary files. It might become the largest file, but not by far (e.g. ffmpeg executables can be pretty large too).

Having a test that reads a file without errors would be minimal. It would be great if you can also check the correctness by e.g. verifying a few pixel values. The best would be if the test image had a few distinct features that you could test for.

MaxSchambach:
Glad you like it!
Yes, cropping the image is not possible, as the sensor size is hard coded in the format.

OK, I will make a PR in ``imageio-binaries`` to add to test files (one ``.raw``, one ``.lfr``) and test the reader plugin. I could verify pixel values by comparing them to the values from the MATLAB toolbox.
I'll try to get to this in the upcoming days!

MaxSchambach:
OK, I have written the tests for the plugin.
As soon as you merge the test images in the PR https://github.com/imageio/imageio-binaries/pull/6, I can push the tests to this PR.

almarklein:
done

MaxSchambach:
Should be ready from my side.

PR #305, opened 2018-02-06
Emma update numpy

EmmaNguyen:


EmmaNguyen:
Update Numpy function used in a plugin of TIFF file:
- Change `fromstring` to `frombuffer`

PR #308, opened 2018-02-18
proper FileNotFoundError, pathlib support

scivision:
Python's [FileNotFoundError](https://docs.python.org/3/library/exceptions.html#FileNotFoundError) is for non-existent directories and files.

Moreover, as of Python 3.3, IOError has been [merged](https://docs.python.org/3/library/exceptions.html#OSError) into OSError, and may eventually be deprecated.

It was very surprising to me to have IOError raised for a missing file or directory, as every other module I've used uses the appropriate FileNotFoundError.

Additionally, added pathlib support, which is Python standard library since Python 3.5.

### DIFF
--- a/imageio/core/functions.py
+++ b/imageio/core/functions.py
@@ -2,7 +2,7 @@
 # Copyright (c) 2015, imageio contributors
 # imageio is distributed under the terms of the (new) BSD License.
 
-""" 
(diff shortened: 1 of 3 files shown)


almarklein:
Thanks for these improvements!

Issue #238, opened 2017-03-07
No seeking allowed!


insatmc:
I am using imageio inside a AWS Lambda function.

When reading a file with imageio.imread I got an Error : No seeking allowed!

What's the meaning of this error ?

almarklein:
Looks like a duplicate of #202, though a fix for that made it into the last release. Are you using the latest version?

insatmc:
I don't think that's the same error as I have:

this is the complete trace of the error


No seeking allowed!: RuntimeError
Traceback (most recent call last):
File "/var/task/func.py", line 32, in handler
images.append(imageio.imread("https://s3.eu-west-2.amazonaws.com/run-records/"+filename))
File "/var/task/imageio/core/functions.py", line 185, in imread
reader = read(uri, format, 'i', **kwargs)
File "/var/task/imageio/core/functions.py", line 111, in get_reader
return format.get_reader(request)
File "/var/task/imageio/core/format.py", line 169, in get_reader
return self.Reader(self, request)
File "/var/task/imageio/core/format.py", line 218, in __init__
self._open(**self.request.kwargs.copy())
File "/var/task/imageio/plugins/pillow.py", line 299, in _open
return PillowFormat.Reader._open(self)
File "/var/task/imageio/plugins/pillow.py", line 85, in _open
pil_try_read(self._im)
File "/var/task/imageio/plugins/pillow.py", line 375, in pil_try_read
im.getdata()[0]
File "/var/task/PIL/Image.py", line 1153, in getdata
self.load()
File "/var/task/PIL/ImageFile.py", line 195, in load
seek(offset)
File "/var/task/imageio/core/request.py", line 481, in seek
ori_seek(i, mode)
File "/var/task/imageio/core/request.py", line 484, in fail_seek
raise RuntimeError('No seeking allowed!')
RuntimeError: No seeking allowed!

almarklein:
Do you have a minimal example to reproduce this? I tried the code below, and it works fine:
```py
im = imageio.imread('https://github.com/imageio/imageio-binaries/raw/master/images/camera.png')
```

What version of Pillow are you using (I tried on 4.0.0)?

almarklein:
ping @insatmc. I'd like to do a release soon. Would be nice to fix this this is still a problem.

almarklein:
I'm about to do a release. As a workaround you can download the image first, and then provide the raw bytes to imageio:

```py
import imageio
from urllib.request import urlopen

[...]

PR #311, opened 2018-02-20
Getting ready for Scipy users

### LINKED ISSUE #289: [Feature Request] Partial Conformance with the SciPy Image I/O API
The long-heralded release of [SciPy 1.0.0](https://scipy.github.io/devdocs/release.1.0.0.html) has landed.

Sixteen years was a hard wait, but good things come to those who wait... *for sixteen years.*

## The Inevitable Bad News

Well, mostly good things. SciPy version:

* 1.0.0 [officially deprecates](https://scipy.github.io/devdocs/release.1.0.0.html#backwards-incompatible-changes) *all* image I/O functionality residing in the `scipy.ndimage` and `scipy.misc` packages – including the commonly called `scipy.ndimage.imread()` and `scipy.ndimage.imwrite()` functions.
* 1.2.0 plans to **permanently remove** this functionality. *Boom!*

Since [our downstream biophysics simulator](https://gitlab.com/betse/betse) heavily depends upon this functionality, SciPy 1.0.0 is more of a bad than a good thing for us. <sup>Yet again, progress only brings pain.</sup>

[SciPy documentation](https://docs.scipy.org/doc/scipy/reference/generated/scipy.misc.imread.html) explicitly recommends `imageio` as a suitable replacement. Since the `imageio` API is *considerably* more bare bones than the equivalent SciPy API, however, `imageio` may prove to be a less-than-suitable replacement for some end users – mostly, us.

## The Overly Optimistic Hope

Ideally, the `imageio` API could be marginally improved to provide at least *partial* support for several of the most commonly used features of SciPy's image I/O API. This includes:

* **Pillow-based mode conversion.** The [`scipy.ndimage.imread()`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.misc.imread.html) function, for example, provides an optional `mode` string parameter permitting external callers to specify a desired target colorspace (defaulting to `None`). If non-`None` and different from the image's current `mode`, this parameter is passed as is to the `PIL.Image.convert()` method. This is two lines of Python. So, this is good.
* **Grayscale flattening.** Again, the [`scipy.ndimage.imread()`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.misc.imread.html) function provides an optional `flatten` boolean parameter permitting external callers to reduce non-grayscale images to grayscale (defaulting to `False`). If `True`, the `mode` parameter is coerced to `F`. Why not `L`? I have no idea and doubt SciPy developers do either. I also resent the overly ambiguous name of this boolean, when something resembling `grayscale` or `as_gray` would have sown less confusion. This also is two lines of Python. So, at least that is good.

This isn't simply syntactic sugar.

This is critical functionality of general interest. Well, O.K... `flatten` is blatantly syntactic sugar.

But `mode` isn't. External callers have *no* explicit access to the `PIL.Image` instance internally instantiated by this function. These optional parameters are the only means callers have of intervening in the conversion process. Doing so after the fact is non-trivial (or perhaps even infeasible).

I accept that `imageio` is not, will never be, and *should* never be an image processor. `imageio` is a pure image I/O framework – just as Code Jesus intended. Still... the thin line between image I/O and image processing is an easy Rubicon to cross, especially when Pillow already performs all of the heavy lifting on our behalf.

## Bold Details for a Bright Future

SciPy's image I/O is unconditionally implemented in terms of Pillow.

`imageio`, on the other hand, is conditionally implemented in terms of Pillow and whatever else it happens to scrounge from the local filesystem (e.g., FreeImage). Presumably, this conditionality prevents the optional `mode` and `flatten` parameters from being directly handled by the `imageio.imread()` function.

I grok that. Nonetheless, would it be feasible to permit these parameters to be:

* Passed by callers as variadic keyword arguments accepted by `imageio.imread()`.
* Passed by Pillow plugins invoked by `imageio.imread()` to the `PIL.Image.convert()` method, munging a `True` value for the `flatten` key into a `mode` value of either `L` or `F`.
* Processed in some fashion by the non-Pillow plugins. Hopefully, these parameters won't simply be silently ignored – but I'll take what I can get.

The `imageio.plugins.pillow.pil_get_frame()` function appears to offer similar functionality, albeit in a manner *not* exposed to external callers. Or is it? My modest (read: *awful*) `grep` skills failed me here.

Alternately, would exposing the `PIL.Image` instance internally instantiated by Pillow plugins via some public variable or getter method of the `imageio.core.format.Reader` class be feasible? I half-suspect this class *already* exposes `PIL.Image` instances, but (again) shamefully failed to `grep` the exact variable or method for doing so.

## The End of a Dream

Thus ends this excrutiating feature request.

Thanks for the long years of gruelling work. May there be many more to come! 🌞

### LINKED ISSUE #296: Fair rounding when converting floats to ints
When converting from floats to ints during saving an image, imageio always rounds down to the next-smallest integer value, instead of rounding to the closest one. e.g. a value of 245.99 will get rounded down to 245. Arguably, rounding up would lead to a better uint8 representation, and I'd suggest changing imageio's rounding behavior in [this function](https://github.com/imageio/imageio/blob/master/imageio/core/util.py#L57) accordingly.

Note that this is likely related to #289 , as `scipy.misc.imsave` (which is now deprecated and suggests using `imageio.imwrite`) rounds to the closest pixel value, so that `imsave` and `imwrite` will often produce different images: many pixels in the output differ by 1 intensity value. While not visually noticeable, this produced weird artefacts in my image analysis pipeline (which is how I found out about this in the first place).

### LINKED ISSUE #300: imageio.imread with jpg image with Spyder 3.2.5 and Python 3.6 debug message Error closing: 'NoneType' object has no attribute 'close'
when running imageio.imread to import a jpg image with DEBUG logging enabled I get the following on Spyder 3.2.5 and Python 3.6

```
import imageio
import logging

logging.basicConfig(level=logging.DEBUG)
a = imageio.imread('picture.jpg')
```
`DEBUG:PIL.Image:Error closing: 'NoneType' object has no attribute 'close'`

the picture imports fine.

### LINKED ISSUE #302: Reading incorrect values from PNG into float64 array
I'm trying to read the following file from the standard Kodak24 dataset (http://r0k.us/graphics/kodak/kodim03.html):

    im = imageio.imread('http://r0k.us/graphics/kodak/kodak/kodim03.png')

However, this results in float64 array containing incorrect values, instead of the expected uint8 array. I did not find a way to configure this. Any idea why this PNG fails? It also fails on all other images from the Kodak dataset.

Doing

    im = imageio.imread('imageio:chelsea.png')

works as expected, however, i.e., it produces a uint8 array.

almarklein:
* [x] Pillow plugin has `pilmode` argument (can't name it `mode` because we already use that name in `get_reader()` (closes #289).
* [x] Pillow plugin has `as_gray` argument.
* [x] Raise an informative error when `mode` is used in `imread`.
* [x] Add a small doc for transitioning users.
* [x] Get rid of debug message when Pillow image fails to close (closes #300) .
* [x] Update autogenerated Pillow info.
* [x] Fix gamma correction dtype (closes #302).
* [x] Fix rounding of float data (closes #296).
* [x] Fix that jpg cannot read from http due it pillow using eek (cloese #238 ).

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -5,12 +5,8 @@ environment:
 
   matrix:
 
-    - PYTHON: "C:\\Python35-conda64"
-      PYTHON_VERSION: "3.5"
-      PYTHON_ARCH: "64"
-
-    - PYTHON: "C:\\Python27-conda64"
-      PYTHON_VERSION: "2.7"
+    - PYTHON: "C:\\Python36-conda64"
+      PYTHON_VERSION: "3.6"
       PYTHON_ARCH: "64"
 
 install:
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -14,5 +14,8 @@ To install imageio, use one of the following methods:
 * Good old ``python setup.py install``
 
 For developers, we provide a simple mechanism to allow importing 
-imageio from the cloned repository. See the file ``imageio.proxy.io`` for
+imageio from the cloned repository. See the file ``imageio.proxy.py`` for
 details.
+
+After installation, checkout the
+:doc:`examples  <examples>` and :doc:`user api <userapi>`. 
--- a/docs/scipy.rst
+++ b/docs/scipy.rst
@@ -1,6 +1,9 @@
 Transitioning from Scipy's imread
 =================================
 
+Scipy is `deprecating <https://scipy.github.io/devdocs/release.1.0.0.html#backwards-incompatible-changes>`_
+their image I/O functionality.
+
 This document is intended to help people coming from
 `Scipy <https://docs.scipy.org/doc/scipy/reference/generated/scipy.misc.imread.html>`_
 to adapt to Imageio's :func:`imread <imageio.imread>` function.
--- a/docs/scipy.rst
+++ b/docs/scipy.rst
@@ -1,0 +1,19 @@
+Transitioning from Scipy's imread
+=================================
+
+This document is intended to help people coming from
+`Scipy <https://docs.scipy.org/doc/scipy/reference/generated/scipy.misc.imread.html>`_
+to adapt to Imageio's :func:`imread <imageio.imread>` function.
+We recommend reading the :doc:`user api <userapi>` and checkout some
+:doc:`examples <examples>` to get a feel of imageio.
+
+Imageio makes use of variety of plugins to support reading images (and volumes/movies)
+from many different formats. Fortunately, Pillow is the main plugin for common images,
+which is the same library that Scipy's ``imread`` used. When the Pillow
+plugin is used, imageio provides the same functionality as Scipy. But
+keep in mind:
+
+    * Instead of ``mode``, use the ``pilmode`` keyword argument.
+    * Instead of ``flatten``, use the ``as_gray`` keyword argument.
+    * The documentation for the above arguments is not on :func:`imread <imageio.imread>`,
+      but on the docs of the individual formats, e.g. :doc:`PNG <format_png-pil>`.
--- a/docs/scipy.rst
+++ b/docs/scipy.rst
@@ -9,11 +9,17 @@ We recommend reading the :doc:`user api <userapi>` and checkout some
 
 Imageio makes use of variety of plugins to support reading images (and volumes/movies)
 from many different formats. Fortunately, Pillow is the main plugin for common images,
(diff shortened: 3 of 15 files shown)


almarklein:
Ping @leycec @untom, it would be great if you could take a look.

PR #316, opened 2018-03-12
New plugin based on BSDF

### LINKED ISSUE #14: Format for 3D data (because npz is not so good)

almarklein:
This adds a plugin to store image data in BSDF files. [BSDF](http://bsdf.io) is a generic data format capable of storing ndarrays. We make use of an extension to add semantics to such an nd array (whether its 2D or 3D to be precise) and add meta data. In this way, we can store singleton images/volumes, as well as series thereof.

BSDF is a simple format and implemented in pure Python. It allows lazy loading of data and streaming. For imageio this means that you can stream a series of images from a remote url, or open a single file containing multiple images and read only a few, without the whole file being read.

BSDF also has implementation in Matlab and JS, so this could be a good way to e.g. send images to JS over a websocket.

Closes #14

### DIFF
--- a/imageio/core/fetching.py
+++ b/imageio/core/fetching.py
@@ -100,7 +100,7 @@ def get_remote_file(fname, directory=None, force_download=False, auto=True):
     # If we get here, we're going to try to download the file
     if os.getenv('IMAGEIO_NO_INTERNET', '').lower() in ('1', 'true', 'yes'):
         raise InternetNotAllowedError('Will not download resource from the '
-                                      'internet because enironment variable '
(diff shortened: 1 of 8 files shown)


jni:
@almarklein sorry that I haven't contributed here for a very long time, but just wanted to say 👏👏👏! This looks very nice! =) The proliferation of formats is annoying but it is heavily mitigated if they come with corresponding parsers that make it into "standard" IO libraries. =)

Issue #299, opened 2018-01-07
update tifffile


ischg:
[Version 2017.09.29](https://www.lfd.uci.edu/%7Egohlke/code/tifffile.py.html) brought significant updates for `tifffile.py`, e.g. consolidated access to metadata. At the moment, imageio holds at version 2017.01.12; it would be great if this could be updated.

almarklein:
@blink1073 do you have time to look into this sometime?

almarklein:
@blink1073  do you have time to look into this?

blink1073:
I'm limiting myself Jupyter-related projects, lest I burn out again.

almarklein:
@blink1073 No problem, good to set boundaries. Sorry to hear you burned out ...

almarklein:
@ischg @scivision It would be awesome if one (or the two) of you could keep an eye out for Tifffile, since I don't use it myself. A ping when an update is needed is already great. PR's even better ;)

scivision:
Probably the adoption rate of imageio will jump, as SciPy is emitting DeprecationWarning. It should be an infrequent occurance in any case.

PR #325, opened 2018-03-27
adding tests (currently failing) for standard TIF metadata write

scivision:
This is a continuation of #299.
Upon testing, I found that imageio 2.3.0 has the exact same problem as before. Perhaps this is because metadata write wasn't part of the test suite.

These standard tags are a core part of tifffile usage for science/medical purposes, as you can automatically record arbitrary data about the experiment etc.

So I created a pull request that shows the failure on write, that has worked for years with tiffffile

### DIFF
--- a/tests/test_tifffile.py
+++ b/tests/test_tifffile.py
@@ -23,9 +23,9 @@ def test_tifffile_format():
 
 def test_tifffile_reading_writing():
     """ Test reading and saving tiff """
-    
+
     need_internet()  # We keep a test image in the imageio-binary repo
-    
+
     im2 = np.ones((10, 10, 3), np.uint8) * 2
 
     filename1 = os.path.join(test_dir, 'test_tiff.tiff')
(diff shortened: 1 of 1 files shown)


almarklein:
Awesome! Can I assume more commits that will make the tests pass? :P

scivision:
The TIF metadata writing case may be simpler, and I would be happy if just write worked for now.

I just need to pass through kwargs `description` and `extraargs` unchanged to tifffile.py. Do you know offhand a simple way to do that? I think you do that for other arguments and to other libraries?

almarklein:
Oh, I think that there may be a misunderstanding of how to pass these metadata. I think this should be the way:

```py
with imageio.get_writer(...) as w:
    w.append_data(image, meta)
```

Alternatively:
```py
im = imageio.core.Image(im)
im.meta.update({...})
imageio.imwrite(im)
```

almarklein:
Maybe we should consider supporting something like this:
```py
imageio.imwrite(filename, im, meta=...)
```

We could add it, but we'd have to verify that "meta" is not used as a kwarg in any of the plugins. Please open an issue for this if you think that this makes sense.

scivision:
I thought about this, and I see I was using the mimwrite API incorrectly.  However, I also realized with the others that using NetCDF4/HDF5 is a much better way to store metadata with video/image data, so we're going to do that instead of hanging onto a legacy format (TIFF)

PR #336, opened 2018-04-18
Use more standard __version__ rather than PILLOW_VERSION

hugovk:
They're both the same thing:

https://github.com/python-pillow/Pillow/blob/master/src/PIL/__init__.py


almarklein:
Fair enough. I verified that this has indeed always ben the case, and we can expect it to remain so.

almarklein:
Thanks!

PR #330, opened 2018-04-03
Better error when tiff is used and enum34 and futures are not present on py27

almarklein:
Closes #328

PR #337, opened 2018-04-18
Setup: include enum and futures backports

clbarnes:
tifffile.py requires enum and concurrent.futures, which are missing from
old python. Now the setup script installs them as necessary.

Minor change: 'full' extra requirements are generated dynamically.


almarklein:
Related to #330. There I improved the error, but did not bother updating `setup.py`. Thanks to you for fixing this as well!

Issue #343, opened 2018-05-18
After upgrade to v2.3.0, Ffmpeg process for webcam does not terminate (Python 2.7)


dennisvang:
Description
-----------

After upgrading from imageio v2.2.0 to v2.3.0 (on a Windows system with Python 2.7 environment), it appears that the ffmpeg process is no longer terminated properly after closing a webcam Reader instance.

How to reproduce issue
------------------------

Here's a minimal test:

```python
import psutil
import imageio

def ffmpeg_alive():
    """ enumerate ffmpeg processes, then wait for them to terminate """
    ffmpeg_processes = []
    for process in psutil.process_iter():
        try:
            if 'ffmpeg' in process.name().lower():
                ffmpeg_processes.append(process)
        except psutil.NoSuchProcess as e:
            print e.message
    still_alive = False
    if ffmpeg_processes:
        __, still_alive = psutil.wait_procs(ffmpeg_processes, timeout=1)
    return still_alive


def test_webcam_ffmpeg_process_termination():
    """ open reader for a webcam, using context manager, then check ffmpeg process status """
    with imageio.get_reader('<video0>') as reader:
        assert reader._proc is not None
        assert ffmpeg_alive()
    assert reader._proc is None
    assert not ffmpeg_alive()


# Run test
print 'imageio version {}'.format(imageio.__version__)
test_webcam_ffmpeg_process_termination()
```

On my system this test *passes* when using a clean conda env running python 2.7 and imageio v2.2.0, but it *fails* when running an (otherwise identical) conda env with imageio v2.3.0 (environment details below). 

Probable cause
---------------

The culprit appears to be the `shell=ISWIN` argument in the `sp.Popen` call in `ffmpeg.Reader._initialize`. 

Proposed solution
-------------------
Quick fix for my specific case would be to remove the `shell` argument (i.e. revert that line to v2.2.0). However, the argument was probably added with a reason? 

Otherwise, a better option would probably be to modify the `ffmpeg.Reader._terminate` method, replacing `self._proc.kill()` by something like the following:

```python
if self.request._video:
    self._proc.communicate('q')
else:
    self._proc.kill()
```

At least this works on my system.

Environment details
--------------------

v2.2.0 environment:

```
certifi                   2016.2.28                py27_0
imageio                   2.2.0                     <pip>
numpy                     1.14.3                    <pip>
Pillow                    5.1.0                     <pip>
pip                       9.0.1                    py27_1
psutil                    5.2.2                    py27_0
[...]

PR #346, opened 2018-05-23
Update the `FfmpegFormat.Reader._terminate()` method to deal with webcam

dennisvang:
This should fix issue #343.

@almarklein Also added a rudimentary test for this issue. This test requires the `psutil` package, but I'm not sure where to specify this requirement (other than the `import` statement).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -74,6 +74,7 @@ install:
         source activate testenv > ${REDIRECT_TO};
         conda install --yes --quiet numpy > ${REDIRECT_TO};
         conda install --yes --quiet pillow -c conda-forge > ${REDIRECT_TO};
+        conda install --yes --quiet psutil > ${REDIRECT_TO};
       fi;
     # Install dependencies (also for testing)
     - pip install invoke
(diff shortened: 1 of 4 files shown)


dennisvang:
Added `psutil` to the Travis CI and AppVeyor yaml files so the CI checks pass. 

However, as the actual webcam tests are skipped during CI testing, I suppose it is still necessary to test this on different systems with a webcam in the loop.

almarklein:
Thanks!

Great to have a specific test for this (even if it only runs locally when a webcam is present).

Do you know whether the not-terminating of ffmpeg is specific to webcam-mode? In other words, do you think that it can also apply to reading of video files with ffmpeg?

dennisvang:
@almarklein The issue does not arise when I read a video file (on a windows 7 machine), e.g. using the example from the docs: [Iterate over frames in a movie](https://imageio.readthedocs.io/en/stable/examples.html#iterate-over-frames-in-a-movie)

almarklein:
Thanks for this!

PR #349, opened 2018-05-29
Added functionality to sort slices based on their position

krishnaanandk:
Added functionality to sort the slices of a dicom series based on their position (distance from the origin) in the patient co-ordinate system. Previously, the unreliable InstanceNumber attribute was being used to sort the slices. Now, all the slices are sorted based on their positions.


almarklein:
One case to keep in mind is where there is one series that has multiple images at the same position, such as in temporal scans (e.g. ECG-gated CT).

almarklein:
(I just solved a conflict introduced by #373)

almarklein:
Closing this. This is a problem that needs more thought and tests ...

PR #359, opened 2018-07-06
Issue 357 add proxy support

eli-s-goldberg:
Added ```proxy_address``` as a passed variable. 
If ```False```, business as usual. If ```proxy_address``` is a proxy address, it's handled through urllib in ```fetching.py```. 

It seemed to work for myself and otherse, but let me know if there are problems. 

Proxy is passed through the ```get_remote_file``` method. 

Example:
```imageio.plugins.ffmpeg.download(proxy_address='https:corporate.zombie.net:2011')```

### DIFF
--- a/imageio/core/fetching.py
+++ b/imageio/core/fetching.py
@@ -129,7 +129,7 @@ def get_remote_file(fname, directory=None, force_download=False, auto=True, prox
         return filename
 
 
-def _fetch_file(url, file_name, print_destination=True, proxy_address=False):
+def _fetch_file(url, file_name, proxy_address=False, print_destination=True):
     """Load requested file, downloading it if needed or requested
 
     Parameters
--- a/imageio/core/fetching.py
+++ b/imageio/core/fetching.py
@@ -13,6 +13,7 @@ from os import path as op
 import sys
 import shutil
 import time
+import urllib
 
 from . import appdata_dir, resource_dirs
 from . import StdoutProgressIndicator, string_types, urlopen
@@ -31,7 +32,7 @@ class NeedDownloadError(IOError):
     """
 
 
-def get_remote_file(fname, directory=None, force_download=False, auto=True):
+def get_remote_file(fname, directory=None, force_download=False, auto=True, proxy_address=False):
     """ Get a the filename for the local version of a file from the web
 
     Parameters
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: imageio/core/fetching.py [hunk 1]
```diff
@@ -140,26 +141,36 @@ def _fetch_file(url, file_name, print_destination=True):
     print_destination: bool, optional
         If true, destination of where file was saved will be printed after
         download finishes.
+    proxy_address: bool, optional
+        If false, no proxy employed; if ,
```
almarklein: Unfinished sentense

imageio/core/fetching.py
```diff
@@ -140,26 +141,36 @@ def _fetch_file(url, file_name, print_destination=True):
...
+
     temp_file_name = file_name + ".part"
     local_file = None
     initial_size = 0
     errors = []
     for tries in range(4):
         try:
-            # Checking file size and displaying it alongside the download url
-            remote_file = urlopen(url, timeout=5.)
+            if proxy_address:
```
almarklein: This bit only works with python 3 (because of the urllib module). We can either try to also make it work with Python 2, or put just a check here and raise an error on Python 2. I'd prefer the latter, because it keeps the code simpler, and end of life for Python 2 is nearing.

imageio/core/fetching.py
```diff
@@ -31,7 +32,7 @@ class NeedDownloadError(IOError):
     """
 
 
-def get_remote_file(fname, directory=None, force_download=False, auto=True):
+def get_remote_file(fname, directory=None, force_download=False, auto=True, proxy_address=False):
```
almarklein: The argument is of string type, and is optional, so I think a default value of None makes more sense.

eli-s-goldberg:
Oh man. Just getting to this. I got sucked into corporate zombie land.

imageio/core/fetching.py
eli-s-goldberg: @almarklein - You've convinced me. It'll change the logic in fetching to use None, but it's a trivial change.

imageio/core/fetching.py
eli-s-goldberg: What's your preferred method for raising Python2 flags in this repo?

```raise RuntimeError('Python2+ proxy unsupported') from error```

or is there a class of errors I can draw from?

imageio/core/fetching.py
eli-s-goldberg: Lol. Will be changed to:
``` If None, no proxy employed; If proxy address, employ address as proxy. ```

imageio/core/fetching.py
almarklein: RuntimeError with a clear message should do :)

almarklein:
Ok, I just changed merged #373 which makes this PR (most PR's actually) unmergable. Sorry about that!

To get beyond this, in this case its probably easiest to create a new PR against the latest master and copying the two changed files over (not the testing.py).

From now on, you can code without worrying about style, and run `invoke black` in the root of the repository once before pushing. You may need to run `pip install black invoke` once.

almarklein:
Closing this for now. It's probably easiest to open a new PR with the intended changes.

Issue #380, opened 2018-09-06
test_lytro_lfr_format takes really long


almarklein:
Like several minutes. Taking up a significant part of the time. Can we scale this down?

almarklein:
cc @MaxSchambach

It's about this test: https://github.com/imageio/imageio/blob/master/tests/test_lytro.py

Seems to be simply the downloading of the data. The 4 largest files in imageio-binaries are used in this test:
![afbeelding](https://user-images.githubusercontent.com/3015475/45349968-073cd200-b5b3-11e8-8596-5483ac19306e.png)

When I delete these files, and run the test, it takes a long time. The second run is really fast. Unfortunately, CI (either, both?) need download the data on each run.

Possible things we can do:
* Accept it :) (good enough for the time being)
* Reduce the image sizes
* Make CI cache the images
* Make the downloading go faster (I just tried to play with chunksize in `core/fetching.py` and it matters a lot.)

MaxSchambach:
Yes, the images are rather large in size. Unfortunately, down-scaling is not possible since they all are saved in a proprietary binary format of which there is no image writer available. I don't see another way of reducing the image sizes, since e.g. compression also does not seem to make any sense.

I think, the most straightforward solution is to make the downloading faster. Downloading <140 MB should not take several minutes.

almarklein:
Made downloading faster and enabled CI caching (I think) in #381. That should do it. Thanks for the reply :)

PR #381, opened 2018-09-11
Dont slow down downloads with too much sleeping

almarklein:
This should help a lot for #380 and should get a better user experience in all cases were the download mechanics is used.


almarklein:
For reference, `time.sleep(0)` does not actually sleep, but *does* imply a thread-switch; it's a good trick for heavy tasks to play nice with other threads. Might not actually be needed here, since the task is io-bound, but should not hurt.

PR #390, opened 2018-10-04
itk package support support

thewtex:
Support using the itk Python package as a backend. If not available, SimpleITK can also be used.

Extend the list of ITK supported formats.


imageio/plugins/simpleitk.py
```diff
@@ -13,23 +13,50 @@
...
+
+            _read_function = _itk.ReadImage
+            _write_function = _itk.WriteImage
+        except ImportError:
+            raise ImportError(
+                "itk could not be found. "
+                "Please try "
+                "  python -m pip install itk "
+                "or refer to "
+                "  https://itkpythonpackage.readthedocs.io/ "
```
almarklein: I guess we want to refer to both `itk` and `simpleitk` here? 

BTW: hooray for binary wheels on pypi!

pytest.ini
```diff
@@ -0,0 +1,2 @@
+[pytest]
+addargs = --assert=plain
```
almarklein: This should not be needed if you run the tests via `invoke tests --unit`. But I suppose it's nice to be able to just run `pytest tests` too :)

setup.py
```diff
@@ -193,7 +193,7 @@ def run(self):
         sdist.run(self)
 
 
-extras_require = {"fits": ["astropy"], "gdal": ["gdal"], "simpleitk": ["SimpleITK"]}
+extras_require = {"fits": ["astropy"], "gdal": ["gdal"], "itk": ["itk"]}
```
almarklein: Out of curiousity, do you have some motivation for why itk is preferred over simpleitk here?

tests/test_simpleitk.py
```diff
@@ -14,10 +14,16 @@
...
-    import SimpleITK as itk
+    import itk
 except ImportError:
-    itk = None
+    pass
+try:
+    if not itk:
+        import SimpleITK as itk
+except ImportError:
+    pass
```
almarklein: How hard would it be to test both itk and simpleitk?
* Can we pip-install both side by side?
* Can we somehow force imageio to use one over the other?

Or is the SimpleItk functionality so thin that we can rely on testing just one?

imageio/plugins/_tifffile.py
```diff
@@ -266,7 +266,7 @@
 *  `python-bioformats <https://github.com/CellProfiler/python-bioformats>`_
 *  `Imread <https://github.com/luispedro/imread>`_
 *  `PyLibTiff <https://github.com/pearu/pylibtiff>`_
-*  `SimpleITK <http://www.simpleitk.org>`_
+*  `ITK <https://www.itk.org>`_
```
almarklein: Note that this file is replaced from time to time with the latest tifffile version.

imageio/plugins/simpleitk.py
thewtex: @almarklein Good idea -- I will also add the updated simpleitk installation instructions.

pytest.ini
thewtex: Ah, nice, I will remove `pytest.ini`. This was precautionary, but it seems to work either way.

setup.py
thewtex: `itk` is more standard and capable.

imageio/plugins/_tifffile.py
thewtex: Ah, ok.

thewtex:
@almarklein thanks for the review :+1: .

tests/test_simpleitk.py
thewtex: I tested locally with `itk`, `simpleitk`, and with both installed. If both are installed, it will prefer `itk`. There is not currently a way to force imageio to use one or the other, but I do not see a big need for it.

I tried a few times to change the TravisCI installation, but I ran into odd failures.

imageio/plugins/gdal.py
```diff
@@ -40,7 +40,7 @@ class GdalFormat(Format):
     def _can_read(self, request):
         if request.extension in (".ecw",):
             return True
-        if has_module("osgeo.gdal"):
+        if has_module("gdal", "osgeo"):
```
almarklein: I changed `has_module` in a different branch, and it can now be used as before:
```suggestion
        has_module("osgeo.gdal"):
```

thewtex: Nice, thanks @almarklein

imageio/core/util.py
```diff
@@ -568,7 +568,7 @@ def has_module(module_name):
         import imp
 
         try:
-            imp.find_module(module_name)
+            imp.find_module(module_name, package)
```
almarklein: The second arg of find_module must be a path. Not sure how to do paths with this approach tbh, so this is kinda broken on 27.
```suggestion
        imp.find_module(module_name)
```

imageio/core/util.py
```diff
@@ -553,7 +553,7 @@ def get_platform():
     return plat % (struct.calcsize("P") * 8)  # 32 or 64 bits
 
 
-def has_module(module_name):
+def has_module(module_name, package=None):
```
almarklein:
```suggestion
def has_module(module_name):
```

imageio/plugins/gdal.py
almarklein: Also see the other uses of has_module ...

tests/test_simpleitk.py
almarklein: Fair enough!

almarklein:
Sorry for not getting back to this earlier. Now it needs a rebase, and a bit of work on `has_module()`.

thewtex:
Fantastic, thanks @almarklein !

Issue #393, opened 2018-10-16
Feature request: use np.asarray in imwrite


hmaarrfk:
Would you consider using `np.asarray` in `imwrite` to coerce things into a numpy array?

https://github.com/imageio/imageio/blob/1e08c70ce38b8607fbc528d91f1b604bec4c7a88/imageio/core/functions.py#L247

Writing dask arrays or xarrays would become more convenient among other things.

almarklein:
Yes, I think this makes sense. Though we should probably catch any exceptions in this step and reraise with a more meaningful error message. A PR would be great 🙏

hmaarrfk:
I'm not even sure what can't be coerced into a numpy array

PR #402, opened 2018-11-01
use asanyarry in high level functions

hmaarrfk:
I couldn't find where the high level API tests went. I added them in core. I need to add a test for the video one too.

I'm really not sure what can't be coerced into a numpy.

Fixes:  #393

### DIFF
--- a/imageio/core/functions.py
+++ b/imageio/core/functions.py
@@ -242,17 +242,16 @@ def imwrite(uri, im, format=None, **kwargs):
         Further keyword arguments are passed to the writer. See :func:`.help`
         to see what arguments are available for a particular format.
     """
+    # Don't coerce subclasses as we need to pass the Image
+    im = np.asanyarray(im)
 
     # Test image
-    if isinstance(im, np.ndarray):
-        if im.ndim == 2:
-            pass
-        elif im.ndim == 3 and im.shape[2] in [1, 3, 4]:
-            pass
-        else:
-            raise ValueError("Image must be 2D (grayscale, RGB, or RGBA).")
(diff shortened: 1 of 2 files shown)


hmaarrfk:
This PR is by no means complete, I’m mostly asking for your thoughts on how you think we can test for objects being coerced into numpy arrays. I’m not too sure it matters if it fails early, or later in the call to img as int.

imageio/core/functions.py
```diff
@@ -242,17 +242,16 @@ def imwrite(uri, im, format=None, **kwargs):
...
-            pass
-        elif im.ndim == 3 and im.shape[2] in [1, 3, 4]:
-            pass
-        else:
-            raise ValueError("Image must be 2D (grayscale, RGB, or RGBA).")
+    if im.ndim == 2:
+        pass
+    elif im.ndim == 3 and im.shape[2] in [1, 3, 4]:
+        pass
     else:
```
almarklein: Would be nice if this can be reduced to a single `if not (...)`

almarklein: And I think I prefer to fail early here if the dtype is not numeric (or some similar measure).

almarklein:
Nice! I think that we should accept "anything that can be coerced into a sensible (numeric and of appropriate shape) array".

almarklein:
Ok, I added a bit to this and merged. The most important bit I added (to me) was to test that the dtype was checked to be numeric, so that when bullocks input is given, the error message is much more descriptive,

Thanks for your contribution!

hmaarrfk:
cool things!

PR #403, opened 2018-11-01
Faster video write

hmaarrfk:
writing to video converts things to string.

This is pretty costly seeing as a string is a full fledged copy of the image.

I'm encoding close to 4k video. I'm still trying to optimize how things are written, but these benchmarks show that at least in the first 30 frames, the system is slowed down, at least by 10-20% by converting the image to a string.

Without the patch:
frame_per_seconds_ffmpeg_no_patch

With the patch:
frame_per_seconds_ffmpeg

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -52,3 +52,6 @@
 
 - Addison Elliott
   Add support for 8-bit paletted BMPs
+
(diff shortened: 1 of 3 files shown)


almarklein:
That's a nice optimization. Very nice to see someone trying to push as much performance out of it :)

PR #407, opened 2018-11-06
Update to latest tiffile, and hide warnings

### LINKED ISSUE #350: Spurious warnings from tiffile
When reading a TIFF file, I get the following warning:
```
[...]/_tifffile.py:7285: UserWarning: module 'imageio.plugins._tifffile' has no attribute 'reverse_bitorder'
  Functionality might be degraded or be slow.

  warnings.warn("%s%s" % (e, warn))
```
This happens on Linux and Windows, with version 2.3.0 from conda as well as with current Git master, commit 1335217992aca7558a3600bd11cd05a22e698e22.

almarklein:
* Copied over the latest original tiffile.py
* Added `return identityfunc  # imageio: just use what's in here` in the `_replace_by()` func

Fixes #350

### DIFF
--- a/imageio/plugins/_tifffile.py
+++ b/imageio/plugins/_tifffile.py
@@ -1,9 +1,11 @@
-#! /usr/bin/python3
+
+
+#! /usr/bin/env python3
 # -*- coding: utf-8 -*-
 # tifffile.py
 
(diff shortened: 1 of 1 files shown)


almarklein:
Good enough, will fix that last bit (installing `pathlib` on Travis for 27) in #405

Issue #410, opened 2018-11-14
Caret (^) symbols need to be escaped when passed to ffmpeg on the command line


cdusold:
This is kind of a rare edge case, but an easy fix. If a path includes the character sequence `^^` it will become a single caret when passed to FFMPEG. This causes the FFMPEG plugin to say that there is no file.

almarklein:
Closed by #411

PR #411, opened 2018-11-14
Fixes imageio/imageio bug #410

cdusold:
There may be other plugins that need this fix, most that I've used don't need it.


almarklein:
Thanks! Could you please add a small comment or a ref to #410, so that the reason for the replacement does not get lost in time? :)

cdusold:
Of course. :)

almarklein:
Thanks!

PR #418, opened 2019-01-02
Support reading images from memory views

orf:
Currently reading an image from a buffer (`memoryview(some_bytes)`) fails. This PR aims to support reading memory views.

### DIFF
--- a/imageio/core/request.py
+++ b/imageio/core/request.py
@@ -183,14 +183,10 @@ class Request(object):
                 elif len(uri) < 256:  # Can go wrong with veeery tiny images
                     self._uri_type = URI_FILENAME
                     self._filename = uri
-                elif isinstance(uri, (binary_type, memoryview)) and is_read_request:
-                    self._uri_type = URI_BYTES
-                    self._filename = "<bytes>"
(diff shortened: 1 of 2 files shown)


imageio/core/request.py
```diff
@@ -183,14 +183,10 @@ def _parse_uri(self, uri):
                 elif len(uri) < 256:  # Can go wrong with veeery tiny images
                     self._uri_type = URI_FILENAME
                     self._filename = uri
-                elif isinstance(uri, binary_type) and is_read_request:
-                    self._uri_type = URI_BYTES
-                    self._filename = "<bytes>"
-                    self._bytes = uri
```
orf: This was a duplicate of the py3k branch below, so I removed it.

almarklein: I don't think it is, it's indented further. Also note that `binary_type` overlaps with `string_types` on Python 2. We can get rid of all this stuff somewhere this year, when we stop supporting Python 2 :)

almarklein:
Nice addition!

almarklein:
Done via #427. Thanks for this!

PR #435, opened 2019-02-19
Add preliminary FreeBSD support and setup.py cleanup

sunpoet:
- Add preliminary FreeBSD support to fix moviepy on FreeBSD 
- Remove python 2.7 stuff in setup.py of imageio 2.5.0 (py3k only)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -188,10 +188,6 @@ extras_require = {
 extras_require["full"] = sorted(set(chain.from_iterable(extras_require.values())))
 
 install_requires = ["numpy", "pillow"]
-if sys.version_info < (3, 4):
-    install_requires.append("enum34")
-if sys.version_info < (3, 2):
-    install_requires.append("futures")
 
 
 setup(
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -214,7 +210,7 @@ def run(self):
     long_description=long_description.replace("__doc__", __doc__),
     platforms="any",
     provides=["imageio"],
-    python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*",
+    python_requires=">=3.4",
```
almarklein: Oh, I would love this change, but let's wait a few months :)

scivision:
This may have been discussed already elsewhere, but it might be worth considering supporting only Python &ge; 3.5 to get the benefits of type hinting and pathlib built-in. This will clean up and help make the project more robust.
Python 3.4 has been EOL'd already and Python 3.5 may be EOL in late 2020.

almarklein:
Let's do this!

almarklein:
Also see #484

Issue #352, opened 2018-06-07
PNG-PIL: low-saturation 16-bit array cast and saved as 8-bit PNG


LandingEllipse:
## What did you do?
Try to save a _low saturation_ grayscale 16-bit PNG.

## What did you expect to happen?
Obtain a 16-bit PNG file.

## What actually happened?
The file was silently saved as an 8-bit PNG, despite the input array dtype being uint16.

## Reproducible example
imageio: 2.3.0
Pillow: 5.1.0
Python: 3.6.5
### Expected behaviour
```python
import numpy as np
import imageio

arr = np.full((32, 32), 256, dtype=np.uint16)
imageio.imwrite("~/16-bit_test.png", arr)

img = imageio.imread("~/16-bit_test.png")
print(img.dtype)  # => "uint16"
```
```
$ file ~/16-bit_test.png
16-bit_test.png: PNG image data, 32 x 32, 16-bit grayscale, non-interlaced
```
### Unexpected behaviour
```python
import numpy as np
import imageio

arr = np.full((32, 32), 255, dtype=np.uint16)
imageio.imwrite("~/16-bit_test.png", arr)

img = imageio.imread("~/16-bit_test.png")
print(img.dtype)  # => "uint8"
```
```
$ file ~/16-bit_test.png
16-bit_test.png: PNG image data, 32 x 32, 8-bit grayscale, non-interlaced
```

## Source of the issue
The behaviour originates in the Pillow plugin's `ndarray_to_pil` function:
https://github.com/imageio/imageio/blob/23ae5c1df0ca0d30931ff42833cf88fc3eb218aa/imageio/plugins/pillow.py#L621-L623
The function checks whether the values in the input ndarray fits in uint8 and recasts the array and sets the mode accordingly if this is the case. 

## Consequences and thoughts
My use case involves capturing calibration images from a scientific imaging system. Some sets are well saturated (e.g. flat fields), while others carry very low signal levels (e.g. bias images). Images can be captured in both 8-bit and 16-bit, and consuming pipeline scripts use the dtype of images for normalisation etc.

Therefore, it is vital that 16-bit data is actually saved as 16-bit PNGs. If for instance a 16-bit bias image peaking at `arr.max() == 255` is saved by `imageio` as an 8-bit PNG, pixel intensities will be completely misrepresented. In the original, the max intensity is just under 0.39% while in the saved 8-bit image the apparent max intensity is 100%. Image viewers will of course also reflect this issue.

[...]

PR #439, opened 2019-04-09
Add keyword parameter to prevent interpreting uint16 arrays as uint8.

LandingEllipse:
Response to issue #352, better late than never...

When set to False, the new keyword parameter `interpret_uint8` accepted by `imwrite` can be set to false to inhibit the default write behaviour of PNGs whereby np.uint16 image arrays are cast to np.uint8 if their elements are compatible with the latter (i.e. all values fall within the range [0, 255]).

I have perhaps not found the most elegant way to implement this change, but it gets the job done. The indirection provided by the format writer inheritance hierarchy means that propagation of the new keyword parameter to `ndarray_to_pil` needs to go via the parent class `PillowFormat.Writer`. Depending on your approach to such things it might have been nice to be able to specify the default behaviour at the level of `PNGFormat`, however this would be complicated by the fact that `ndarray_to_pil` is a general function used indirectly by all children of `PillowFormat.Writer` and thus requires a default behaviour to be defined at that level or by the function itself. I opted for the latter, with the choice for `PillowFormat.Writer` to only present it with an override if the parameter has been passed in as a keyword argument from `imwrite`.

Basically I've tried to disturb as little of the codebase as possible; please let me know if you would prefer a different approach and I'll be happy to accommodate.

### DIFF
--- a/imageio/plugins/pillow.py
+++ b/imageio/plugins/pillow.py
@@ -211,13 +211,9 @@ class PillowFormat(Format):
                 im = im[:, :, 0]
             self._written = True
             self._meta.update(meta)
-            if "prefer_uint8" in self._meta:
-                img = ndarray_to_pil(
-                    im, self.format.plugin_id, self._meta["prefer_uint8"]
-                )
-                del self._meta["prefer_uint8"]
-            else:
-                img = ndarray_to_pil(im, self.format.plugin_id)
+            img = ndarray_to_pil(
+                im, self.format.plugin_id, self._meta.pop("prefer_uint8", True)
+            )
             if "bits" in self._meta:
                 img = img.quantize()  # Make it a P image, so bits arg is used
             img.save(self._fp, format=self.format.plugin_id, **self._meta)
--- a/imageio/plugins/pillow.py
+++ b/imageio/plugins/pillow.py
@@ -211,7 +211,13 @@ class PillowFormat(Format):
                 im = im[:, :, 0]
             self._written = True
             self._meta.update(meta)
-            img = ndarray_to_pil(im, self.format.plugin_id)
+            if "interpret_uint8" in self._meta:
+                img = ndarray_to_pil(
+                    im, self.format.plugin_id, self._meta["interpret_uint8"]
+                )
+                del self._meta["interpret_uint8"]
+            else:
+                img = ndarray_to_pil(im, self.format.plugin_id)
             if "bits" in self._meta:
                 img = img.quantize()  # Make it a P image, so bits arg is used
             img.save(self._fp, format=self.format.plugin_id, **self._meta)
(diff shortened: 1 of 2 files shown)


imageio/plugins/pillow.py
```diff
@@ -286,6 +292,11 @@ class PNGFormat(PillowFormat):
         bits. In this case, given as a number between 1-256.
     dictionary (experimental): dict
         Set the ZLIB encoder dictionary.
+    interpret_uint8: bool
```
almarklein: I think renaming this to `prefer_uint8` would more accurately describe the behavior. What do you think?

imageio/plugins/pillow.py
```diff
@@ -211,7 +211,13 @@ def _append_data(self, im, meta):
                 im = im[:, :, 0]
             self._written = True
             self._meta.update(meta)
-            img = ndarray_to_pil(im, self.format.plugin_id)
+            if "interpret_uint8" in self._meta:
+                img = ndarray_to_pil(
+                    im, self.format.plugin_id, self._meta["interpret_uint8"]
```
almarklein: Maybe use `self._meta.get("interpret_uint8", True)` instead of the `if-else`?

tests/test_pillow.py
```diff
@@ -148,6 +148,12 @@ def test_png():
     im2 = imageio.imread(fnamebase + "1.png")
     assert im2.dtype == np.uint16
 
+    # issue #352 - prevent low-luma uint16 truncation to uint8
+    arr = np.full((32, 32), 255, dtype=np.uint16)  # values within range of uint8
+    imageio.imwrite(fnamebase + ".png", arr, interpret_uint8=False)
+    im = imageio.imread(fnamebase + ".png")
+    assert im.dtype == np.uint16
```
almarklein: Could you please perform the three possible cases here: arg not given, arg set to `False`, arg set to `True`?

almarklein:
Thanks! I made a few small comments.

imageio/plugins/pillow.py
LandingEllipse: Perfect, the inaccuracy of "interpret" has been bugging me but I was stuck trying to come up with a better term.

imageio/plugins/pillow.py
LandingEllipse: Aye, I agree this would be cleaner. Only consequence as noted in the PR is that this shifts specification of default behaviour to `PillowFormat.Writer._append_data`. To be fair, upon further consideration this doesn't really present an issue since the plugin's code base is relatively small; I'll implement the change.

tests/test_pillow.py
LandingEllipse: Will do!

LandingEllipse:
Alright, all comments should have been responded to through commits; let me know if you have any further feedback.

almarklein:
Nice, looking good! Thanks!

PR #442, opened 2019-04-19
user-defined size for memtest

clbarnes:
Allows callers to define the size limit where they want memtest to raise an error. Retains compatibility with `True`/`False`, and interprets either a number of bytes, or a string e.g. "256MB", "10.2Yib".

Includes tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,6 +53,8 @@ matrix:
       sudo: true
       python: "3.8-dev"
       env: TEST_UNIT=1
+  allow_failures:
+    - python: "3.8-dev"
 
 
 before_install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,6 +54,8 @@ matrix:
       python: "3.8-dev"
       env: TEST_UNIT=1
   allow_failures:
+    # future compatibility is nice,
+    # but a failure here shouldn't break the build until released
     - python: "3.8-dev"
 
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,7 +44,14 @@ matrix:
       python: "3.6"
       env: TEST_UNIT=1 TEST_COVER=1
     - os: linux
-      python: "3.7-dev"
+      dist: xenial
+      sudo: true
+      python: "3.7"
+      env: TEST_UNIT=1
+    - os: linux
+      dist: xenial
+      sudo: true
+      python: "3.8-dev"
       env: TEST_UNIT=1
 
 
--- a/README.md
+++ b/README.md
@@ -23,11 +23,11 @@ Python 2.7 and 3.4+, and is easy to install.
 Here's a minimal example of how to use imageio. See the docs for 
 <a href='http://imageio.readthedocs.io/en/latest/examples.html'>more examples</a>.
 <pre>
->>> import imageio
->>> im = imageio.imread('imageio:chelsea.png')  # read a standard image
->>> im.shape  # im is a numpy array
-(300, 451, 3)
->>> imageio.imwrite('~/chelsea-gray.jpg', im[:, :, 0])
+import imageio
+im = imageio.imread('imageio:chelsea.png')  # read a standard image
+im.shape  # im is a numpy array
+>> (300, 451, 3)
+imageio.imwrite('~/chelsea-gray.jpg', im[:, :, 0])
 </pre>
 
 <h2>API in a nutshell</h2>
@@ -127,25 +127,25 @@ one or more plugins. In that way, the burden of each developer is low,
 and together we can make imageio into a really useful library!
 </p>
 
+<h2>Contributing</h2>
 
-## Contributing
+<p>Install a complete development environment:</p>
 
-Install a complete development environment:
-
-```bash
+<pre>
 pip install -r requirements.txt
 pip install -e .
-```
+</pre>
 
-*N.B. this does not include GDAL because it has awkward compiled dependencies*
+<p><i>N.B. this does not include GDAL because it has awkward compiled dependencies</i></p>
 
-You may see failing test(s) due to missing installed components.
-On ubuntu, do `sudo apt install libfreeimage3`
+<p>You may see failing test(s) due to missing installed components.
+On ubuntu, do <code>sudo apt install libfreeimage3</code></p>
 
-Style checks, unit tests and coverage are controlled by `invoke`.
-Before committing, check these with:
+<p>
+Style checks, unit tests and coverage are controlled by <code>invoke</code>.
+Before committing, check these with:</p>
 
-```bash
+<pre>
 # reformat code on python 3.6+
 invoke autoformat
 # check there are no style errors
@@ -154,4 +154,4 @@ invoke test --style
 invoke test --unit
 # check test coverage (re-runs tests)
 invoke test --cover
-```
+</pre>
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ Optional Python packages:
     <li>itk or SimpleITK (for ITK formats)</li>
     <li>astropy (for FITS plugin)</li>
     <li>osgeo (for GDAL plugin)</li>
-</ul>  
+</ul>
 
 
 <h2>Origin and outlook</h2>
@@ -126,3 +126,32 @@ It is our hope to form a group of developers, whom each maintain
 one or more plugins. In that way, the burden of each developer is low,
 and together we can make imageio into a really useful library!
 </p>
+
+
+## Contributing
+
+Install a complete development environment:
+
+```bash
+pip install -r requirements.txt
+pip install -e .
+```
+
+*N.B. this does not include GDAL because it has awkward compiled dependencies*
+
+You may see failing test(s) due to missing installed components.
+On ubuntu, do `sudo apt install libfreeimage3`
+
+Style checks, unit tests and coverage are controlled by `invoke`.
+Before committing, check these with:
+
+```bash
+# reformat code on python 3.6+
+invoke autoformat
+# check there are no style errors
+invoke test --style
+# check the tests pass
+invoke test --unit
+# check test coverage (re-runs tests)
+invoke test --cover
+```
--- a/imageio/core/functions.py
+++ b/imageio/core/functions.py
@@ -76,10 +76,53 @@ like ``imageio.imread(imageio.core.urlopen(url).read(), '.gif')``.
 
 from __future__ import absolute_import, print_function, division
 
+from numbers import Number
+import re
+
 import numpy as np
 
 from . import Request, RETURN_BYTES
 from .. import formats
+
+
+MEMTEST_DEFAULT_MIM = "256MB"
+MEMTEST_DEFAULT_MVOL = "1GB"
+
+
+mem_re = re.compile(r"^([\d.]+)\s*([kKMGTPEZY]*i?[bB])?$")
+sizes = {"": 1, None: 1}
+for i, si in enumerate([""] + list("kMGTPEZY")):
+    for b, divisor in [("b", 8), ("B", 1)]:
+        sizes[si + b] = 1000 ** i / divisor
+        sizes[si.upper() + "i" + b] = 1024 ** i / divisor
+
+
+def to_nbytes(arg, default=None):
+    if arg is True:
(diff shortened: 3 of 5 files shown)


clbarnes:
Also addresses #443 with requirements.txt and addition to README.md

clbarnes:
Also updated the travis config to build against 3.7 (release) and 3.8-dev.

README.md
```diff
@@ -126,3 +126,32 @@ It is our hope to form a group of developers, whom each maintain
...
+```bash
+# reformat code on python 3.6+
+invoke autoformat
+# check there are no style errors
+invoke test --style
+# check the tests pass
+invoke test --unit
+# check test coverage (re-runs tests)
+invoke test --cover
+```
```
almarklein: This is great. Can you please make this compatible with plain html (using `<h2>` and `<pre>`)? The readme is also used for the web-page at imageio.github.io.

imageio/core/functions.py
```diff
@@ -281,19 +324,26 @@ def mimread(uri, format=None, memtest=True, **kwargs):
...
+        list of images consumes greater than the amount of memory specified.
+        This is to protect the system from using so much memory that it needs
+        to resort to swapping, and thereby stall the computer. E.g.
         ``mimread('hunger_games.avi')``.
+        If the argument is a number, that will be used as the threshold number
+        of bytes.
+        If the argument is a string, it will be interpreted as a memory size with
+        units (e.g. '1kB', '250MiB', '80Yb'). This is case-sensitive.
+        If True, resort to the default (for compatibility reasons).
+        Default: '256MB'
```
almarklein: I agree with making this configurable like this. How strongly do you feel about the smart string parsing? I think I would prefer less complexity and accept only bool and int. In the documentation we can make a note that using `40*2**20` matches 40 MiB.

requirements.txt
```diff
@@ -0,0 +1,22 @@
+# core
+numpy
+pillow
+enum34; python_version < "3.4"
+futures; python_version < "3.2"
+pathlib; python_version < "3.0"
+# extras
```
almarklein: Can you please make the "section headers" more clear, e.g. with a newline?

.travis.yml
```diff
@@ -44,7 +44,14 @@ matrix:
...
     - os: linux
-      python: "3.7-dev"
+      dist: xenial
+      sudo: true
+      python: "3.7"
+      env: TEST_UNIT=1
+    - os: linux
+      dist: xenial
+      sudo: true
+      python: "3.8-dev"
```
almarklein: Awesome!

README.md
clbarnes: Ah, got it. I suspected this may be the case but I would have expected an html file to have an html extension rather than md (although it looks like github won't automatically display the readme with an html extension).

imageio/core/functions.py
clbarnes: I think it would be beneficial - people don't often think about their RAM in bytes now that 8+GB are commonplace, and the more readable and writeable it is, the better (too easy to drop or add an extra `*1000`). That said, I accept that too much flexibility adds API surface.

I'd prefer to keep it (I feel the function is well-defined and well-documented enough to avoid pain points, as it goes through a dict rather than truly parsing for the units), but of course defer to you.

tests/test_core.py
```diff
@@ -620,6 +623,40 @@ def test_functions():
...
 
+@pytest.mark.parametrize(
+    "arg,expected",
+    [
+        (1, 1),
+        ("1", 1),
+        ("8b", 1),
+        ("8B", 8),
+        ("1MB", 1000 ** 2),
+        ("1Gib", 1024 ** 3 / 8),
```
clbarnes: Add decimals

imageio/core/functions.py
```diff
@@ -76,12 +76,55 @@
...
 
 from . import Request, RETURN_BYTES
 from .. import formats
 
 
+MEMTEST_DEFAULT_MIM = "256MB"
+MEMTEST_DEFAULT_MVOL = "1GB"
+
+
+mem_re = re.compile(r"^([\d.]+)\s*([kKMGTPEZY]*i?[bB])?$")
```
clbarnes: replace `([\d.]+)` with `(\d+\.?\d*)` to prevent multiple decimal points

clbarnes: Possibly use `?` instead of `*` for first character of unit string (although multiple letters will get stopped by the dict lookup later anyway).

.travis.yml
clbarnes: What do you think about allowing failures on the dev release? Doesn't seem necessary as it's passing right now but might save someone a headache at some point.

imageio/core/functions.py
clbarnes: An alternative would be this:

```python
def mimread(uri, memtest=True, memlimit=None, **kwargs):
    if memtest:
        if memlimit:
            nbyte_limit = to_nbytes(memlimit)
        else:
            nbyte_limit = to_nbytes(MEMTEST_DEFAULT_MIM)
    else:
        nbyte_limit = False

    ...
```

This leaves the existing API untouched (where technically the original proposal is a breaking change), and reduces the number of things one parameter is doing.

imageio/core/functions.py
```diff
@@ -302,7 +352,7 @@ def mimread(uri, format=None, memtest=True, **kwargs):
         ims.append(im)
         # Memory check
         nbytes += im.nbytes
-        if memtest and nbytes > 256 * 1024 * 1024:
+        if nbyte_limit and nbytes > nbyte_limit:
             ims[:] = []  # clear to free the memory
             raise RuntimeError(
                 "imageio.mimread() has read over 256 MiB of "
```
clbarnes: Update exception message

.travis.yml
```diff
@@ -44,8 +44,17 @@ matrix:
...
+      sudo: true
+      python: "3.7"
       env: TEST_UNIT=1
+    - os: linux
+      dist: xenial
+      sudo: true
+      python: "3.8-dev"
+      env: TEST_UNIT=1
+  allow_failures:
+    - python: "3.8-dev"
```
almarklein: Perhaps a small comment why we accept failures from this env? Does it glitch?

imageio/core/functions.py
```diff
@@ -76,12 +76,55 @@
...
 from .. import formats
 
 
+MEMTEST_DEFAULT_MIM = "256MB"
+MEMTEST_DEFAULT_MVOL = "1GB"
+
+
+mem_re = re.compile(r"^(\d+\.?\d*)\s*([kKMGTPEZY]?i?[bB])?$")
+sizes = {"": 1, None: 1}
+for i, si in enumerate([""] + list("kMGTPEZY")):
```
almarklein: You lost me after "P": are "EB", "ZB" and "YB" common memory size units?

imageio/core/functions.py
almarklein: > people don't often think about their RAM in bytes now that 8+GB are commonplace, and the more readable and writeable it is, the better (too easy to drop or add an extra *1000).

I agree. Let's do it this way.

imageio/core/functions.py
```diff
@@ -76,12 +76,55 @@
...
+    for b, divisor in [("b", 8), ("B", 1)]:
+        sizes[si + b] = 1000 ** i / divisor
+        sizes[si.upper() + "i" + b] = 1024 ** i / divisor
```
almarklein: You've made lowercase "b" mean bits, but I think we can assume that we always talk about bytes. I'd prefer to make this mechanism completely case insensitive: I think it would not be uncommon for people to write simply "100mb".

As I understand it (but correct me if I'm wrong!), the "MiB" notation explicitly means `1024` (`2**10`) multipliers, whereas for "MB" its can mean either. In practice almost everybody uses the 1024 multiplier, with the exception of hard drive manufacturers. So I'd prefer to always use the 1024 multiplier, and allow both the "MiB" and "MB" notation.

imageio/core/functions.py
almarklein: Ah, here's an explanation: https://en.wikipedia.org/wiki/Units_of_information

clbarnes: They're not common, but if we can add 3 characters and future proof the code for the next 100 years it seems a shame not to ;) SI units aren't changing.

.travis.yml
clbarnes: I think it passes; I just prefer keeping dev pythons under `allow_failures` as a matter of course, because the feature set isn't stable and if it *were* to fail, we shouldn't lose any sleep over it until it actually gets released. I can add a comment to this effect.

imageio/core/functions.py
clbarnes: My understanding is that there was a period where the SI prefixes meant jumps of 1000, unambiguously. Then it was decided that computers working in binary meant that we may as well use 1024 when talking about computers, which created ambiguity as context-dependent unit prefixes are not a good idea. So in \*hits wikipedia\* 1998, a standard was published giving binary prefixes their own letters, and thereby making SI units Standard again. k unambiguously means 1000, and Ki unambiguously means 1024, according to all major standards organisations and the SI system.

I fully accept that I am more nitpicky than most about this but I personally have strong feelings about maintaining correctness in this regard - in both the prefixes and the b/B. SI units are valuable because they are standardised and unambiguous. If people want to use them incorrectly, they're welcome to: that's the "consenting adults" ethos of python. We don't account for typos in file names either. I don't think the kilo/kibi will make much of a difference to anyone, but I'll add to the docstring to make it clear that proper capitalisation is expected.

For me, this change was about empowering users to write expressive, grok-able code, rather than saving keystrokes.

imageio/core/functions.py
clbarnes: Just to clarify, going with the additional `memlimit` parameter?

imageio/core/functions.py
clbarnes: ... but I'll also stop it from accepting `"i"` as a unit.

imageio/core/functions.py
almarklein: haha, agreed

imageio/core/functions.py
almarklein: No, with you original approach. I rather introduce a small backward-incompatibility for people that did not use `True`/`False` than convolute the API with an extra param.

.travis.yml
almarklein: Thanks, yeah a comment like this in the file would help me if I forget the reason in a few months :)

imageio/core/functions.py
almarklein: I see you already implemented the other, sorry for not being more clear earlier!

imageio/core/functions.py
almarklein: > but I personally have strong feelings about maintaining correctness in this regard 

I'm with you on that :)

As for `b` vs `B`, what about only supporting uppercase `B`? Using lowercase `b` will be invalid, because the argument should be given in bytes, not bits. This way we have correctness without user confusion (a quick line in the docs should sort this out).

As for whether `1MB` should/can mean `2**20` bytes or `10**6`, I'm still not sure. If I right-click a file on Windows, it may say `1 MB (1048576 bytes)`. On Ubuntu `1 MB (1000000 bytes). This page is also not clear https://en.wikipedia.org/wiki/Byte#Unit_multiples. That said, the difference will be low in practice, and I'd be happy to go with your approach.

clbarnes: > On Ubuntu `1 MB (1000000 bytes)`

Oh boy is there drama on that: https://wiki.ubuntu.com/UnitsPolicy From the looks of it, [it varies by program](https://askubuntu.com/a/897397/439260). You're right that there is more ambiguity in the general ecosystem than I optimistically implied, though.

I'm happy to drop `'b'` though, as it'll remove the most likely pain point and as you said before, few people think in bits (I was trying to imagine a situation where someone might be thinking in terms of data transmission speed, but came up short).

almarklein:
Awesome. Ready?

clbarnes:
I think so!

almarklein:
(I missed your reply and then went on holiday; sorry for the late merge :))

almarklein:
Thanks a lot for this!

PR #451, opened 2019-05-30
Fixed bug in image_as_uint

clintg6:
There's a bug in image_as_uint that causes imsave to crash if you try to save an image of all 0's and the datatype isn't unit8 or uint16.

You can easily test this:
`import imageio as io`
`img = np.zeros((256,256), dtype=np.int32)`
`io.imsave("test.png",img)`

### DIFF
--- a/imageio/core/util.py
+++ b/imageio/core/util.py
@@ -119,7 +119,7 @@ def image_as_uint(im, bitdepth=None):
         if not np.isfinite(ma):
             raise ValueError("Maximum image value is not finite")
         if ma == mi:
-             return im.astype(out_type) 
+             return im.astype(out_type)
         _precision_warn(dtype_str1, dtype_str2, "Range [{}, {}].".format(mi, ma))
         # Now make float copy before we scale
         im = im.astype("float64")
--- a/imageio/core/util.py
+++ b/imageio/core/util.py
@@ -91,7 +91,7 @@ def image_as_uint(im, bitdepth=None):
     ):
         # Already the correct format? Return as-is
         return im
-    if im.sum() == 0:
+    if np.all(im == 0):
         return im.astype(out_type)
     if dtype_str1.startswith("float") and np.nanmin(im) >= 0 and np.nanmax(im) <= 1:
         _precision_warn(dtype_str1, dtype_str2, "Range [0, 1].")
--- a/imageio/core/util.py
+++ b/imageio/core/util.py
@@ -119,10 +119,7 @@ def image_as_uint(im, bitdepth=None):
         if not np.isfinite(ma):
             raise ValueError("Maximum image value is not finite")
         if ma == mi:
-            if np.all(im == ma): 
-                return im.astype(out_type)
-            else:
-                raise ValueError("Max value == min value, ambiguous given dtype")
+             return im.astype(out_type) 
         _precision_warn(dtype_str1, dtype_str2, "Range [{}, {}].".format(mi, ma))
         # Now make float copy before we scale
         im = im.astype("float64")
--- a/imageio/core/util.py
+++ b/imageio/core/util.py
@@ -91,8 +91,6 @@ def image_as_uint(im, bitdepth=None):
     ):
         # Already the correct format? Return as-is
         return im
-    if np.all(im == 0):
-        return im.astype(out_type)
     if dtype_str1.startswith("float") and np.nanmin(im) >= 0 and np.nanmax(im) <= 1:
         _precision_warn(dtype_str1, dtype_str2, "Range [0, 1].")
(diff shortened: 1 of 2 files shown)


almarklein:
Nice catch! Can we move that check to the last `else` clause? The error occurs inside `if ma == mi:`, maybe we can put `if mi ==0 and ma == 0` in there. That way we don't have to do this check (which has to go over all pixels) for every image.

clintg6:
I moved it into the lower if block and then got to thinking. It will fix this bug for all zeros but it doesn't solve all possible bugs. Imagine a user wants to write out a solid image whether in grayscale or RGB. In this scenario, this function will still crash so I changed it so that if all the pixels values equal to the max value to return the image as the out type. 

Consequently, I had to update the core test as well because there was a test that raised an error if the array had identical values which is not good because it lead to this bug slipping into imsave where it would crash on any kind of image where the pixel values are all the same.

imageio/core/util.py
```diff
@@ -119,7 +119,10 @@ def image_as_uint(im, bitdepth=None):
         if not np.isfinite(ma):
             raise ValueError("Maximum image value is not finite")
         if ma == mi:
-            raise ValueError("Max value == min value, ambiguous given dtype")
+            if np.all(im == ma):
```
almarklein: This can be `if mi == ma` (which would probably be better when there are NaNs).

almarklein:
Good thinking, you're right.

skrackow:
Maybe one might note that this bug stems from a division by zero whenever all px values are identical. Hence, I used a small number addition when scaling, and np.clip to keep values in 0, 255, so that all equal images should become black (hope). Since px values are manipulated anyway in this clause for undetermined dtype, that seems acceptable. Otherwise, np.unique(im).size could lead to some closer examination...
    else:
        mi = np.nanmin(im)
        ma = np.nanmax(im)
        if not np.isfinite(mi):
            raise ValueError("Minimum image value is not finite")
        if not np.isfinite(ma):
            raise ValueError("Maximum image value is not finite")
        im = im.astype("float64")
        # Scale the values between 0 and 1 then multiply by the max value
        im -= mi
        im /= (ma-mi) + 1e-5
        im *= np.power(2.0, bitdepth) - 1
        im += .5
    im = np.clip(im, 0, np.power(2.0, bitdepth) - 1)
    return im.astype(out_type)

imageio/core/util.py
clintg6: Ok I will change it so that it doesn't raise an error if mi == ma and just returns im.astype(out_type)

clintg6:
That code works fine if the image is all 0's but if you take an image of all anything between 1 too 255 it gets forced to all 0's as well. Also np.unique is much slower compared to np.all(im == ma)

`im = np.zeros((256,256), dtype=np.int32) + 255`
`ma = im.max()`
`%timeit np.unique(im).size == 1`
533 µs ± 1.37 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

`%timeit np.all(im == ma)
25.2 µs ± 2.82 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

As far as I can tell, if we eliminate the raise error and just return im.astype(out_type) it solves everything and prevents triggering the divison by zero error.

skrackow:
Not sure if it is possible to have nan in im, but if so, one would run into troubles again if all other values are equal. Found that something like im[~np.isnan(im)]) could be used?!

almarklein:
> Not sure if it is possible to have nan in im,

It happens. But these lines make sure that we don't have trouble with them:
```py
        mi = np.nanmin(im)
        ma = np.nanmax(im)
```

skrackow:
np.all(im == ma) will wrongly return FALSE, if there are nan in an im that otherwise contains only equal values. At least, that is what I suspected.

clintg6:
That's true and it has been updated where it only tests if mi == ma now in the latest commit to avoid this issue.

almarklein:
Nice to see how the final diff is really small!  if you remove one more space, than the style checker in Travis will stop complaining about indentation ;)

clintg6:
I removed a space after return im.astype(out_type). I must be missing it.

imageio/core/util.py
```diff
@@ -119,7 +119,7 @@ def image_as_uint(im, bitdepth=None):
         if not np.isfinite(ma):
             raise ValueError("Maximum image value is not finite")
         if ma == mi:
-            raise ValueError("Max value == min value, ambiguous given dtype")
+             return im.astype(out_type)
```
almarklein: hehe, I mean this space just left of "return" :)

almarklein:
Thanks for your contribution!

PR #458, opened 2019-06-20
Add support for compressed FITS images

joesingo:
Compressed FITS files, commonly named `*.fits.fz`, cannot currently be
loaded with the FITS plugin, but they are supported by astropy. See the [astropy documention on compressed image date in FITS](https://docs.astropy.org/en/stable/io/fits/usage/unfamiliar.html#compressed-image-data).

This commit adds support for loading compressed images.

If there is interest in merging this, I can add a compressed file to imageio-binaries and add a test.

### DIFF
--- a/imageio/plugins/fits.py
+++ b/imageio/plugins/fits.py
@@ -94,11 +94,7 @@ class FitsFormat(Format):
             hdulist = _fits.open(self.request.get_file(), cache=cache, **kwargs)
 
             self._index = []
-            allowed_hdu_types = (
-                _fits.ImageHDU,
-                _fits.PrimaryHDU,
-                _fits.CompImageHDU
-            )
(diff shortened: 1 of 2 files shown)


almarklein:
Great! Yes, please add an image here for use in the tests: https://github.com/imageio/imageio-binaries

You can run `invoke autoformat` to stop Travis complaining about style. (you may need to `pip invoke black`)

joesingo:
That should fix the style errors, and I've added a basic test for reading a compressed image. Of course the Travis build will not pass until the new image is added to the imageio-binaries repo. The PR for that is here: https://github.com/imageio/imageio-binaries/pull/13

The test does pass on my local machine though (after placing the new image in the cache directory to prevent the code actually trying to fetch it from imageio-binaries).

Thanks!

almarklein:
I restarted Travis and its green now. Great! Thanks for this contribution!

PR #462, opened 2019-07-21
Improve imageio import time by avoiding pkg_resources import

hmaarrfk:
My import time on python 3.7 is goes down from 0.2 s to 0.12 seconds.

make sure to do it twice so the first time it compiles all the pyc files
```
python -X importtime profile_imageio.py 2> import.log
python -X importtime profile_imageio.py 2> import.log
tuna import.log
```

https://github.com/pypa/setuptools/issues/510


almarklein:
Awesome! Travis complaints about style, but I will fix that in a later commit.

Thanks!

PR #466, opened 2019-08-01
better support for multiple cams with same name

### LINKED ISSUE #441: can not open second camera when two camera have same name
My system is windows 10.  
I have two camera. It have same name 'USB Camera'.  
I will get an error when I execute the following command.  
```
cam1 = imageio.get_reader('<video0>')
cam2 = imageio.get_reader('<video1>')
```
I found the problem because the first names of the two cameras are the same, and using alternative name to open camera will avoid this problem.  
I think imageio\plugins\ffmpeg.py‘s parse_device_names function needs to be modified to use alternative name.

almarklein:
Closes #441

Multiple cameras can be connected that have the same (friendly name). Fortunately ffmpeg also provides an "alternative name", which is a long and weird identifier string. The hard part seems to be getting this string over the CLI, since it has backslashes and ampersands. It works on Windows 10, but I am less certain about Unix.

Therefore, the default behavior is to use the friendly name, and switch to the alt name only when there are duplicates. In short, this should not break existing behavior, fixes the duplicate camera thing on Windows, and *might* fix it on Unix too :)

### DIFF
--- a/imageio/plugins/ffmpeg.py
+++ b/imageio/plugins/ffmpeg.py
@@ -73,17 +73,17 @@ def _get_ffmpeg_api():
 class FfmpegFormat(Format):
     """ The ffmpeg format provides reading and writing for a wide range
     of movie formats such as .avi, .mpeg, .mp4, etc. And also to read
-    streams from webcams and USB cameras. 
-    
+    streams from webcams and USB cameras.
+
     To read from camera streams, supply "<video0>" as the filename,
     where the "0" can be replaced with any index of cameras known to
     the system.
-    
+
     To use this plugin, the ``imageio-ffmpeg`` library should be installed
     (e.g. via pip). For most platforms this includes the ffmpeg executable.
     One can use the ``IMAGEIO_FFMPEG_EXE`` environment variable to force
     using a specific ffmpeg executable.
-    
+
     When reading from a video, the number of available frames is hard/expensive
     to calculate, which is why its set to inf by default, indicating
     "stream mode". To get the number of frames before having read them all,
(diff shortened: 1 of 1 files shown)


imageio/plugins/ffmpeg.py
```diff
@@ -662,20 +662,39 @@ def run(self):
...
             if in_video_devices and line.startswith('"'):
-                device_names.append(line[1:-1])
+                friendly_name = line[1:-1]
+                device_names.append([friendly_name, ""])
+            elif in_video_devices and line.lower().startswith("alternative name"):
+                alt_name = line.split(" name ", 1)[1].strip()[1:-1]
+                if sys.platform.startswith("win"):
+                    alt_name = alt_name.replace("&", "^&")  # Tested to work
+                else:
+                    alt_name = alt_name.replace("&", "\\&")  # Does this work?
```
almarklein: @One-sixth I was able to test this on Windows. Do you happen to be on Unix? If so, it would be great if you could give this a try.

imageio/plugins/ffmpeg.py
```diff
@@ -662,20 +662,39 @@ def run(self):
...
                 in_video_devices = True
             elif "devices" in line:
                 # set False for subsequent "devices" sections
                 in_video_devices = False
-    return device_names
+    # Post-process, see #441
+    # prefer friendly names, use alt name if two cams have same friendly name
+    device_names2 = []
+    for friendly_name, alt_name in device_names:
+        if friendly_name not in device_names2:
```
almarklein: An easy way to test the working of the alt names without the need for actually having two equally named cameras:

set the line above to `if False:'

almarklein:
I decided to just merge this PR, but would still love to know if this also works on Unix. We'll create a new PR to fix for Unix, or update the comment.

Issue #480, opened 2019-09-22
Compressing GIFs


LucaCappelletti94:
Hello, since I needed to compress GIF images created with imageio in a python script I've created a python package wrapping the [gifsicle](https://www.lcdf.org/gifsicle/) library called [pygifsicle](https://github.com/LucaCappelletti94/pygifsicle).

I wanted to ask if it could be of interest to add an "optimize" option to the imageio writer for GIF images using gifsicle. If that's the case, I could do a pull request adding the feature, but user requiring the feature would need to install separately the gifsicle library.

The installation process is straightforward (just a brew install on mac and an apt-get on Debian/Ubuntu), so nothing esoteric involved.

Do let me know what do you think about it!

Luca

almarklein:
What about Windows? :)

I'm not sure about adding a dependency (even if it's optional), especially if it requires more than `pip install`.

However, an example that combines imageio and pygifsicle to generate optimized gif's would be really nice, and helpful for people who need such a thing. We coulr add it here: https://imageio.readthedocs.io/en/stable/examples.html

LucaCappelletti94:
> What about Windows? :)
For windows [there's a port](https://eternallybored.org/misc/gifsicle/), so it should be covered, but I guess the installation process would not be that straightforward.
> I'm not sure about adding a dependency (even if it's optional), especially if it requires more than `pip install`.
I agree!
> However, an example that combines imageio and pygifsicle to generate optimized gif's would be really nice, and helpful for people who need such a thing. We coulr add it here: https://imageio.readthedocs.io/en/stable/examples.html
Sure, it could be something like the [one I posted here on stackoverflow](https://stackoverflow.com/questions/43140560/imageio-in-python-compressing-gif/58048255#58048255).

Should I do a pull request for the documentation?

LucaCappelletti94:
I have done [a pull request adding the example](https://github.com/imageio/imageio/pull/481), do let me know what do you think of it!

PR #481, opened 2019-09-23
Added example for compressing GIFs using pygifsicle

LucaCappelletti94:
Added example for compressing GIFs, as described in [issue 480](https://github.com/imageio/imageio/issues/480).


almarklein:
CI is complaining about the size of the branch. It looks like there was an example image which was later removed. I'll squash and merge so that the image won't be included.

Thanks!

PR #492, opened 2019-11-12
Fix SPE filesize check for SPE v3.

anntzer:
SPE v3 files can include an xml metadata footer, which needs to be
excluded when checking the file size; otherwise one gets a spurious
warning about nonmatching number of frames.

See
ftp://ftp.princetoninstruments.com/public/Manuals/Princeton%20Instruments/SPE%203.0%20File%20Format%20Specification%20Issue%205%20(4411-0140).pdf
table 1-2.

Parsing the actual xml footer to extract metadata would be a useful
(independent) addition.

### DIFF
--- a/imageio/plugins/spe.py
+++ b/imageio/plugins/spe.py
@@ -38,7 +38,9 @@ class Spec:
         "datatype": (108, "<h"),  # dtypes
         "xdim": (42, "<H"),
         "ydim": (656, "<H"),
+        "xml_footer_offset": (678, "<Q"),
         "NumFrames": (1446, "<i"),
+        "file_header_ver": (1992, "<f"),
     }
 
     metadata = {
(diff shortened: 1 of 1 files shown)


almarklein:
Great to see this resolved!

anntzer:
Fully converting the xml metadata to a dict is a bit too much for me to do, but I think it would still be nice to expose that xml (either as a raw bytestring, or as a parsed ElementTree).  Thoughts on how to expose that in the imageio metadata model?

almarklein:
A bit hard to say right now. The biggest challenge for imageio right now is to come up with a good way to expose/provide metadata, so that scikit-image can also use it. On the other hand, maybe it can be as simple as making it a subdict in the metadata dict?

anntzer:
You mean a subdict key?  It's a fairly nested xml with a not so obvious schema, so I can't just convert it trivially to a flat dict.

almarklein:
Sorry, I read your comment too quickly and overlooked that it can't really be a dict. In that case, attaching the raw bytestring seems like the most sensible option.

PR #493, opened 2019-11-18
Expose SPE3 xml footer.

anntzer:
per https://github.com/imageio/imageio/pull/492#issuecomment-554942515.

Note that (some) SPE3 files don't set the "classic" header metadata fields other than the version and header offset -- this is explicitly allowed by the spec -- so we don't read these entries either.

I don't have a strong opinion on how to name the dict entry -- `__xml` is just a placeholder.

_init_meta_data_pre_v3 is just cut-n-pasted from the old _get_meta_data.

spec: ftp://ftp.princetoninstruments.com/public/Manuals/Princeton%20Instruments/SPE%203.0%20File%20Format%20Specification%20Issue%205%20(4411-0140).pdf

### DIFF
--- a/imageio/plugins/spe.py
+++ b/imageio/plugins/spe.py
@@ -265,6 +265,7 @@ class SpeFormat(Format):
             self._char_encoding = char_encoding
 
             info = self._parse_header(Spec.basic)
+            self._file_header_ver = info["file_header_ver"]
             self._dtype = Spec.dtypes[info["datatype"]]
             self._shape = (info["ydim"], info["xdim"])
             self._len = info["NumFrames"]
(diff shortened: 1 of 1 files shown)


imageio/plugins/spe.py
```diff
@@ -387,6 +394,12 @@ def _parse_header(self, spec):
                 ret[name] = v
             return ret
 
+        def _init_meta_data_post_v3(self):
+            info = self._parse_header(Spec.basic)
+            self._file.seek(info["xml_footer_offset"])
+            xml = self._file.read()
+            self._meta = {"__xml": xml}
+
```
almarklein: This function looks much shorter than the `pre_v3` version. Does it mean that when we merge this PR, the meta dict will look different for some files? Some users may depend on the meta data structure.

almarklein:
Also some style issues. Please run `invoke autoformat` to fix all style issues.

imageio/plugins/spe.py
anntzer: Yes, but the point is that I don't have the old metadata entries available (I would need to understand the xml schema for that).
Right now they're read from the "old-style" header, but SPE3 files don't have that header filled correctly (this is explicitly allowed by the spec, which says "read the xml if you want the correct metadata"), so the metadata entries there are basically just made up with no warning to the end user, which seems worse.

anntzer:
autoformat done

imageio/plugins/spe.py
almarklein: Thanks. So if a user reads an SPE3 file *now*, and inspect its metadata, and does the same after the PR is merged, there won't be anything missing, only added?

anntzer: No; some fields, which were previously filled with nonsense, will be missing after this PR.

almarklein: > which were previously filled with nonsense

Ha! ok, sounds good to me.

almarklein:
Thanks!

PR #497, opened 2019-12-15
Fix false-positive zip detection

hexfaker:
Fix bug with false-positive zip detection when some directory in path contains `.zip` suffix in its name.
Add test case for that. It fails on current upstream master

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -8,6 +8,7 @@ import shutil
 import ctypes.util
 from zipfile import ZipFile
 from io import BytesIO
+import tempfile
 
 import numpy as np
 import pytest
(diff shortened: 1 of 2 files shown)


tests/test_core.py
```diff
@@ -200,6 +201,17 @@ def test_request():
     assert R._filename_zip[0] == get_remote_file("images/chelsea.zip")
     assert R.filename == get_remote_file("images/chelsea.zip") + "/chelsea.png"
 
+    # Test zip false-positive detection
+    with tempfile.TemporaryDirectory() as tmp_path:
+        bait_zip = Path(tmp_path) / "test.zip"
+        bait_zip.mkdir()
+        file_path = bait_zip / "foo.jpg"
+        chelsia = get_remote_file("images/chelsea.png")
+        shutil.copy(chelsia, file_path)
```
almarklein: The Py3.5 CI build fails on `shutil.copy` with `PosixPath` objects. Can you please convert to string here? :)

hexfaker: Sure. I migrated to old good os.path.

But now there is a weird problem with CI...

almarklein:
> But now there is a weird problem with CI...

I restarted the stalled build. The Litro error on the 3.6 build is a known issue - I just have not found the time to look into it yet.

almarklein:
Thanks!

PR #506, opened 2020-02-07
Adding GIF and reading from bytes examples

asm:
I've run into these uses a few times now and end up reading the source to remember how to do this.  I figured a quick example would save others from having to do the same 😄


almarklein:
Thanks! And sorry for the late reaction. Travis is wining, but I'll just merge this and patch it up.

PR #518, opened 2020-03-06
Adding Travis-CI support for Arm64

odidev:
Travis-CI has added support for ARM64. Added ARM64 job in Travis-CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ addons:
   apt:
     packages:
     - libfreeimage3
+    - ffmpeg
 
 # Only run travis build on push for master branch
 branches:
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -25,6 +26,10 @@ matrix:
     - os: linux
       python: "3.6"
       env: TEST_UNIT=0 TEST_STYLE=1
+    - os: linux
+      python: "3.6"
+      arch: arm64
+      env: TEST_UNIT=0 TEST_STYLE=1
```
almarklein: You can remove this one: for style tests the machine should not matter.

tests/test_ffmpeg.py
```diff
@@ -42,22 +42,25 @@ def setup_module():
...
+        path = imageio_ffmpeg.get_ffmpeg_exe()
+        #path = "/usr/bin/ffmpeg"
+        # cleanup
+        os.environ.pop("IMAGEIO_FFMPEG_EXE")
+        if oldpath:
+            os.environ["IMAGEIO_FFMPEG_EXE"] = oldpath
+        print(path)
+        assert os.path.isfile(path)
+    except Exception:
+        skip("Skipped for arm64")
```
almarklein: euhm ... if this test now fails on any system, it will simply report "skipped for arm64" :)  Can we instead detect whether this is running on arm and skip if so?

almarklein:
Thanks! I did not know we could actually test ARM on CI. Sweet! Two small comments.

.travis.yml
odidev: Sure...Will modify it.

tests/test_ffmpeg.py
odidev: Correct...thanks for pointing it out. Will update the same.

odidev:
Hi @almarklein,
Incorporated all the changes suggested by you, please have a look.
Thanks

almarklein:
Thanks!

Issue #391, opened 2018-10-10
Cannot read webcam on macOS


florianletsch:
Following the [webcam example in the docs](https://imageio.readthedocs.io/en/stable/examples.html#grab-frames-from-your-webcam), I am trying to access the webcam on my MacBook (external Logitech camera attached).

```
import imageio
reader = imageio.get_reader('<video0>')

# ... long error message (see full listing below)
# IndexError: No video4linux camera at <video0>.
```

This call fails with quite an extensive error message (see below)

ffmpeg 4.0.2 is installed from homebrew (`brew install ffmpeg`).

```
ffmpeg
ffmpeg version 4.0.2 Copyright (c) 2000-2018 the FFmpeg developers
  built with Apple LLVM version 10.0.0 (clang-1000.10.43.1)
  configuration: --prefix=/usr/local/Cellar/ffmpeg/4.0.2 --enable-shared --enable-pthreads --enable-version3 --enable-hardcoded-tables --enable-avresample --cc=clang --host-cflags= --host-ldflags= --enable-gpl --enable-libmp3lame --enable-libx264 --enable-libxvid --enable-opencl --enable-videotoolbox --disable-lzma
  libavutil      56. 14.100 / 56. 14.100
  libavcodec     58. 18.100 / 58. 18.100
  libavformat    58. 12.100 / 58. 12.100
  libavdevice    58.  3.100 / 58.  3.100
  libavfilter     7. 16.100 /  7. 16.100
  libavresample   4.  0.  0 /  4.  0.  0
  libswscale      5.  1.100 /  5.  1.100
  libswresample   3.  1.100 /  3.  1.100
  libpostproc    55.  1.100 / 55.  1.100
Hyper fast Audio and Video encoder
usage: ffmpeg [options] [[infile options] -i infile]... {[outfile options] outfile}...

Use -h to get full help or, even better, run 'man ffmpeg'
```

Is this related to ffmpeg or imageio?

How can I check if ffmpeg itself is able to read the webcam properly?

Error message:

```
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-3-71e67521605e> in <module>
----> 1 reader = imageio.get_reader('<video0>')

~/miniconda3/envs/test-imageio/lib/python3.6/site-packages/imageio/core/functions.py in get_reader(uri, format, mode, **kwargs)
    141
    142     # Return its reader object
--> 143     return format.get_reader(request)
    144
    145

[...]

PR #525, opened 2020-05-11
Fix that webcam does not work (Windows 10)

almarklein:
For some reason I am recently unable to read from my webcam. In #510 I made sure that the fps is always set when reading from a webcam, which fixed an issue with webcams on MacOS (see #391). The default fps there was set to 15. If I change this to 30 it works for me again. A bit flaky, but it works for me :)

### DIFF
--- a/imageio/plugins/ffmpeg.py
+++ b/imageio/plugins/ffmpeg.py
@@ -289,7 +289,7 @@ class FfmpegFormat(Format):
             # Specify input framerate?
             if self.request._video:
                 if "-framerate" not in str(self._arg_input_params):
-                    self._arg_input_params.extend(["-framerate", str(float(fps or 15))])
+                    self._arg_input_params.extend(["-framerate", str(float(fps or 30))])
             # Get local filename
             if self.request._video:
                 index = int(self.request._video[-2])
@@ -458,7 +458,7 @@ class FfmpegFormat(Format):
                                 "support for cameras."
                             )
                     raise IndexError(
-                        "No camera at {}.\n\n{}".format(self.request._video, err_text)
+                        "No (working) camera at {}.\n\n{}".format(self.request._video, err_text)
                     )
                 else:
                     self._meta.update(meta)


dennisvang:
Always passing `-framerate 30` may work for most webcams, by accident, but it breaks if the camera does not support 30 fps.

For example, my Avermedia video grabber only supports 25 fps when capturing in PAL-B mode, and fails with:

```
[dshow @ 0000021fefc99f80] Could not set video options
video=AVerMedia USB Polaris Analog Capture: I/O error
```
The same happens if I specify e.g. 40 fps for my laptop's built-in webcam, which only supports up to 30 fps.

On Windows, `ffmpeg` is smart enough to discover the proper framerate, as long as the `-framerate` option is not explicitly specified.

If the explicit `-framerate 30` setting is only really necessary on macOS (as you mentioned above for #391), perhaps it would make sense to add the option conditionally, only for macOS? Or at least remove the explicit `-framerate` setting on Windows?

Something like this perhaps, [here](https://github.com/imageio/imageio/blob/v2.12.0/imageio/plugins/ffmpeg.py#L299)?

```python
...
# Specify input framerate?
if self.request._video:
    if platform.system().lower() == 'darwin':  # <<<<<<<<< this is the only change
        if "-framerate" not in str(self._arg_input_params):
            self._arg_input_params.extend(["-framerate", str(float(fps or 30))])
...
```

The way it is now, I would need to either detect framerate myself, before calling `get_reader()`, or write a monkey-patch that removes the `-framerate 30` from [FfmpegFormat.Reader._arg_input_params](https://github.com/imageio/imageio/blob/v2.12.0/imageio/plugins/ffmpeg.py#L299)... Both are highly inconvenient.

almarklein:
Just tested on Linux, and it seems to work without setting the explicit framerate. It could be that its only needed on osx. Or that ffmpeg has improved over time. I'm in favour of making this mac-only, or even remove it completely if we someone with a mac can confirm that we can. It would also be good to add a note in the docs that if the webcam does not work, you may need to set the framerate to match. Would you have time for a PR?

FirefoxMetzger:
@almarklein @dennisvang Hm ... technically this would be not backward compatible, right? People on Windows may have started to rely on the fact that webcam streams are 30 FPS in imageIO even if the webcam would default to something else. Do we worry about this?

An alternative here could be to introduce `fps=30` as a kwarg and allow `fps=None` to let ffmpeg decide the framerate. This can be combined with a depreciation warning that the default will switch to `fps=None` in v3 leading to the desired behavior. We will break the API in multiple places for similar changes in default behavior so we could simply add this one to the list.

almarklein:
Since this feature is more or less broken depending on the platform and type of camera, I care much more about making this work for everyone out of the box, then to make sure that the fps does not change for some users. Also, the 30 fps, which was 15 earlier, is very arbitrary anyway.

dennisvang:
@FirefoxMetzger Technically it would be backward incompatible, but as far as I can see the 30 fps default for webcams is an undocumented feature in a "private" function, so I tend to agree with @almarklein.

> An alternative here could be to introduce `fps=30` as a kwarg ...

Not sure if I understand you correctly, but [FfmpegFormat.Reader._open()](https://github.com/imageio/imageio/blob/v2.13.1/imageio/plugins/ffmpeg.py#L260) already has an `fps=None` kwarg, which [*is* documented](https://imageio.readthedocs.io/en/stable/reference/_backends/imageio.plugins.ffmpeg.html#parameters-for-reading). I don't think this default kwarg should be modified, because it applies not only to webcams, but also to video *files*.

I can create a pull request. Should that refer to issue #391?

almarklein:
> I can create a pull request. Should that refer to issue #391?

Thanks! Yes, it would be great if you refer to the related issues/prs.

FirefoxMetzger:
> Not sure if I understand you correctly, but FfmpegFormat.Reader._open() already has an fps=None kwarg, which is documented. I don't think this default kwarg should be modified, because it applies not only to webcams, but also to video files.

Oh yes, you're right. I am confusing the function we are talking about. In that case, would it make sense to view this as a bug? 

This is the situation I've understood so far: `fps=None` should use the file's/webcam's default, but in fact, uses `fps=30` instead. We are setting it explicitly because it caused problems on MacOS a few years back, but we don't seem to need it on Linux (and likely not on Windows), so we can stop setting it explicitly for those two OSes.

> I can create a pull request. Should that refer to issue #391?

🚀 

@almarklein Will you review the PR or should I?

almarklein:
I can have a look at that PR.

PR #526, opened 2020-05-22
Fix for Python 3.9/3.10: Replace deprecated cElementTree

hugovk:
`xml.etree.cElementTree` has been deprecated since Python 3.3 and will be removed in Python 3.9 or Python 3.10 (https://github.com/python/cpython/pull/19921).

Since Python 3.3, `xml.etree.ElementTree` will use a fast implementation whenever available.

* https://docs.python.org/3/library/xml.etree.elementtree.html

### DIFF
--- a/imageio/plugins/_tifffile.py
+++ b/imageio/plugins/_tifffile.py
@@ -2207,7 +2207,7 @@ class TiffFile(object):
 
     def _ome_series(self):
         """Return image series in OME-TIFF file(s)."""
-        from xml.etree import cElementTree as etree  # delayed import
+        from xml.etree import ElementTree as etree  # delayed import
         omexml = self.pages[0].description
         try:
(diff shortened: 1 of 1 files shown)


almarklein:
Thanks!

almarklein:
No wait ... `_tifffile.py` is a vendored module. The fix should be applied [here](https://github.com/cgohlke/tifffile).

almarklein:
We can then update our version when the fix has been applied upstream.

hugovk:
Thanks, here's the upstream PR:

* https://github.com/cgohlke/tifffile/pull/8

hugovk:
Upstream:

> Thanks. This has already been fixed in the development version and will be released soon..

https://github.com/cgohlke/tifffile/pull/8#issuecomment-633635857

PR #529, opened 2020-05-26
more robust loading of FEI SEM metadata and support of top and bottom tif metadata

jon-lab:
There were issues with loading metadata from FEI SEM when the metadata was attached at the top of the tif file.
Also loading the file as utf8 with skipped errors should make it more robust.


almarklein:
LGTM, thanks@

PR #530, opened 2020-05-28
ENH: Better support for itk/sitk plugins

jond01:
- Add nifti suffixes to ITK_FORMATS
- Reader arguments:
  - pixel_type is relevant for both itk and sitk
  - fallback_only is True/False - itk only


almarklein:
LGTM Thanks!

Issue #533, opened 2020-06-08
How can we set the socket timeout on http(s) request?


jneuhauser:
Hello,

I´m using this great library for fetching random images from unsplash like:

```
import imageio

def unsplash_image_generator():
    while True:
        try:
            image = imageio.imread('https://source.unsplash.com/random')
        except OSError as e:
            print("Error fetching random unsplash image: ", e)
        else:
            yield image
```

Sadly on 6 of 8000 images there is a socket.timeout exception thrown and i wan´t to reduce the timeout to a short time like 3 seconds.
How can this be done and keep the simplicity of this great library?

Thanks in advance.

almarklein:
That timeout is hardcoded here: https://github.com/imageio/imageio/blob/master/imageio/core/request.py#L347

It would be nice to allow specifying the timeout. I think it makes the most sense to use an environment variable for that (https://imageio.readthedocs.io/en/stable/envvariables.html). Care for a PR? ;)

jneuhauser:
I've created a PR that uses the param timeout throught **kwargs.
I think this is more common for such params like timeout, but on the other side this is more a taste of the developer itself.

almarklein:
#closed by #534

PR #534, opened 2020-06-09
Allow setting timeout for http/ftp requests through environment

jneuhauser:
Hello devs,

as pointed out in https://github.com/imageio/imageio/issues/533, sometimes it makes sense to set the socket timeout for http/ftp requests.
Allow setting timeout for http/ftp requests through environment so that the following is possible:

```
import imageio

def unsplash_image_generator():
    os.environ['IMAGEIO_REQUEST_TIMEOUT'] = '3'
    while True:
        try:
            image = imageio.imread('https://source.unsplash.com/random')
        except OSError as e:
            print("Error fetching random unsplash image: ", e)
        else:
            yield image
```

I don't know if this is the best option to allow setting the timeout by the user.

Another solution could be:
```
Author: Johann Neuhauser <johann@it-neuhauser.de>
Date:   Tue Jun 9 10:43:31 2020 +0200

    Allow setting timeout for http/ftp requests through **kwargs

diff --git a/imageio/core/request.py b/imageio/core/request.py
index dbd455b..77b77ee 100644
--- a/imageio/core/request.py
+++ b/imageio/core/request.py
@@ -344,7 +344,11 @@ class Request(object):
 
         elif self._uri_type in [URI_HTTP or URI_FTP]:
             assert not want_to_write  # This should have been tested in init
-            self._file = urlopen(self.filename, timeout=5)
+            if 'timeout' in self._kwargs:
+                timeout = self._kwargs['timeout']
+            else:
+                timeout = 5
+            self._file = urlopen(self.filename, timeout=timeout)
             self._file = SeekableFileObject(self._file)
 
         return self._file
```

### DIFF
--- a/imageio/core/request.py
+++ b/imageio/core/request.py
@@ -344,7 +344,10 @@ class Request(object):
 
         elif self._uri_type in [URI_HTTP or URI_FTP]:
             assert not want_to_write  # This should have been tested in init
-            self._file = urlopen(self.filename, timeout=5)
+            timeout = os.getenv('IMAGEIO_REQUEST_TIMEOUT')
+            if timeout is None or not timeout.isdigit():
+                timeout = 5
+            self._file = urlopen(self.filename, timeout=float(timeout))
             self._file = SeekableFileObject(self._file)
 
         return self._file
--- a/docs/envvariables.rst
+++ b/docs/envvariables.rst
@@ -18,5 +18,7 @@ for the current Python process use
 * ``IMAGEIO_FORMAT_ORDER``: Determine format preference. E.g. setting this
   to ``"TIFF, -FI"`` will prefer the FreeImage plugin over the Pillow plugin,
   but still prefer TIFF over that. Also see the ``formats.sort()`` method.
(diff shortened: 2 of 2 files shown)


imageio/core/request.py
```diff
@@ -344,7 +344,10 @@ def get_file(self):
 
         elif self._uri_type in [URI_HTTP or URI_FTP]:
             assert not want_to_write  # This should have been tested in init
-            self._file = urlopen(self.filename, timeout=5)
+            timeout = os.getenv('IMAGEIO_REQUEST_TIMEOUT')
+            if timeout is None or not timeout.isdigit():
+                timeout = 5
+            self._file = urlopen(self.filename, timeout=timeout)
```
almarklein, line 350: Probably also need to convert that string to float here :)

almarklein:
Good point about using the kwargs. However, the kwargs are currently all fed to the plugin. Plus this parameter does not apply to most use-cases. Therefore I'd prefer the env var.

almarklein:
Thanks, a really nice addition!

PR #549, opened 2020-09-07
black

almarklein:
black changes its output


almarklein:
(CI fails on check that tests size of the commit)

PR #557, opened 2020-10-14
Preserve metadata when creating Array from another Array.

### LINKED ISSUE #544: Creating Array from preexisting Array drops metadata
```python
>>> import numpy as np, imageio
>>> t = imageio.core.Array(np.array([1, 2]), meta={"foo": 1})
>>> (t * 2).meta
Dict([('foo', 1)])
>>> t[:].meta
Dict([('foo', 1)])
>>> imageio.core.Array(t).meta
Dict()
```
Given that most operations on an Array appear to just keep the metadata as is, it seems reasonable to expect that creating a new Array from a preexisting one, without explicitly setting new metadata, should also keep the original metadata.

anntzer:
Closes #544.

### DIFF
--- a/imageio/core/util.py
+++ b/imageio/core/util.py
@@ -124,7 +124,7 @@ class Array(np.ndarray):
         if not (meta is None or isinstance(meta, dict)):
             raise ValueError("Array expects meta data to be a dict.")
         # Convert and return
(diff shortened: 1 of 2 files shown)


almarklein:
Looks good, thanks!

PR #559, opened 2020-10-19
More reliable tiff reading as volume

### LINKED ISSUE #558: volread not reading multi page tifs
Hi, I'm having an issue with volread where it only reads the first page of tiffs. Please see this example file:

https://www.dropbox.com/s/vob9iwwqzwrsz7s/20201012_17_11_48_344486.1_XYT_ch_1.tif?dl=0

I installed imagio just today and have the latest version, on Ubuntu 18.04. These tif files were created with SciScan on Windows 7 and on that OS imagio seemed to have no problem reading all the pages properly.

I can load the file with another program like ImageJ, re-save the tif and then imagio is fine.

I'm guessing there might be something unusual about the tif files I have? Not sure why it appears to platform dependent.

In the meantime, I can work around it with a mimread and a loop, but if there is a problem I thought you should know.

almarklein:
Closes #558 

By default a `TiffFile.as_array()` selects all data from the current series. So if there are multiple series, only the first is selected,

### DIFF
--- a/imageio/plugins/tifffile.py
+++ b/imageio/plugins/tifffile.py
@@ -244,7 +244,10 @@ class TiffFormat(Format):
                 # Read data as single 3D (+ color channels) array
                 if index != 0:
                     raise IndexError('Tiff support no more than 1 "volume" per file')
-                im = self._tf.asarray()  # request as singleton image
+                # There is self._tf.asarray(), but it picks the first series by
+                # default, so this seems more reliable. See #558.
+                ims = [p.asarray() for p in self._tf.pages]
+                im = np.stack(ims, 0)
                 meta = self._meta
             else:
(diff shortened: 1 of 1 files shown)


almarklein:
@kmilos @blink1073 I think that this change more closely reflects what should be expected from `volread()` on a tiff image. Could you please have a quick look to if I'm not overlooking something?

kmilos:
It seems it should return more data that the original code, yes, but sorry, I have no experience with volumetric TIFF data to say if this is the best thing to do...

Nit-picking on a style note if I may, the code readability could be improved maybe for posterity: test directly for `len(self._tf.pages)` first to show actual intent rather than go via derived `ims`?

almarklein:
Thanks @kmilos !

PR #565, opened 2020-11-11
Lytro plugin: read only metadata

MaxSchambach:
# TL;DR

Adds the option to only read the metadata from Lytro LFR, RAW, LFP and F01-RAW files.

# Description
Using the option `meta_only=True` when reading a Lytro image, an empty array with the corresponding metadata is returned.
Since the Lytro binary files are comparably large (~55 MB) this can significantly speed up cases where one is only interested in the metadata. For example, this can be useful when sorting/aggregating images with similar metadata.

Updates the tests accordingly.


almarklein:
Thanks, this looks great!

PR #567, opened 2020-11-17
tifffile plugin: Various fixes

lschr:
`Reader.get_meta_data(i)` already worked correctly, but `Reader.get_data(i).meta` contained the result of the last call to `Reader.get_meta_data` regardless of `i`.

`Writer.append_data(img, meta)` set the global metadata to `meta`, which was also not what I would expect.

Here is a possible fix which hopefully does not break anything. With this, any frame written to a file using `append_data(…, meta=None)` will have the global metadata associated with it. This is debatable but resembles the old behavior. The alternative would be to write the global metadata only for frame 0 and no metadata for other frames unless explicitly given—which I think would actually make more sense.

### DIFF
--- a/imageio/plugins/feisem.py
+++ b/imageio/plugins/feisem.py
@@ -52,6 +52,9 @@ class FEISEMFormat(TiffFormat):
             metadata : dict
                 Dictionary of metadata.
             """
+            if hasattr(self, "_fei_meta"):
+                return self._fei_meta
+
             md = {"root": {}}
             current_tag = "root"
             reading_metadata = False
@@ -80,7 +83,8 @@ class FEISEMFormat(TiffFormat):
                             md[current_tag][key] = val
             if not md["root"] and len(md) == 1:
                 raise ValueError("Input file %s contains no FEI metadata." % filename)
-            self._meta.update(md)
+
+            self._fei_meta = md
             return md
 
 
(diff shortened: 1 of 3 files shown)


lschr:
Thinking of it, I see two options for reading metadata:
1. Use only the frame's metadata for each frame (currently implemented).
2. Take the global metadata dict and update that with frame's metadata for each frame (not implemented).

almarklein:
> With this, any frame written to a file using append_data(…, meta=None) will have the global metadata associated with it. This is debatable but resembles the old behavior. The alternative would be to write the global metadata only for frame 0 and no metadata for other frames unless explicitly given—which I think would actually make more sense.

I think you're right here.

Does Tiff have a notion of global meta data, or would this come down to the metadata of the first image.

In `get_meta_data()`, when `index` is `None` it should return the global metadata. If that does not exist, the data at 0 makes the most sense.

> Thinking of it, I see two options for reading metadata:

I think option 1 is best, since there is a way to read the global meta data. (Albeit this might perhaps be documented better.)

lschr:
I don't know much about TIFF, but as far as I gather from tifffile, there is no real global metadata, but the first frame is used. As of now, global metadata will be written with the first frame, but only if no metadata was passed to `append_data` for the first frame. I clarified that in the documentation.

lschr:
I fixed two more things:
- As TiffWriter.save has been deprecated, use the new TiffWriter.write if available
- Explicitly set `contiguous=False` when saving. This is the default in tifffile nowadays and (AFAICS) treats the saved images more independently, e.g., allows for per-frame metadata, compression, and so on.

almarklein:
Looking good. Thanks @lschr !

PR #572, opened 2020-11-30
SPE plugin: Read metadata from SDT-control software

lschr:
This adds support for reading metadata written by the SDT-control microscopy software. Hardly any changes to old code, so it should be safe.


almarklein:
Thanks!

PR #573, opened 2020-12-01
Allow JPEG quality factors > 95

### LINKED ISSUE #571: Saving high-quality JPEG ignores quality factor
After storing an image in JPEG format with imageio's imwrite method and quality 100, I was surprised to see that the quantization table of the saved image did not match my expectations, i.e., I would have expected a quantization table with all 1s. This is also the quantization table that Pillow produces for quality 100. The quantization tables also differ for quality factors 96 through 99.
Could someone please check whether I am using the imwrite method in the intended way? Otherwise, I think this might be a bug in imageio.

Here is the code to reproduce this issue:
```python
import imageio
import numpy as np
from PIL import Image
print(imageio.__version__)


# Create random pixels
img = np.random.randint(0, 256, size=(512, 512), dtype=np.uint8)
quality = 100

# Use imageio to save the image in JPEG format
imageio_filename = "/tmp/imageio.jpg"
imageio.imwrite(imageio_filename, img, quality=quality)

# Read quantization table from saved JPEG file
im = Image.open(imageio_filename)
quantization_table = np.array(im.quantization[0]).reshape(8, 8)
print("[imageio] Quantization table\n", quantization_table)

# Use Pillow to save the image
pil_filename = "/tmp/pil.jpg"
im = Image.fromarray(img)
im.save(pil_filename, quality=quality)

# Read quantization table from saved JPEG file
im = Image.open(pil_filename)
quantization_table = np.array(im.quantization[0]).reshape(8, 8)
print("[Pillow] Quantization table\n", quantization_table)
```

Output:
```
2.9.0
[imageio] Quantization table
 [[ 2  1  1  1  1  1  2  1]
 [ 1  1  2  2  2  2  2  4]
 [ 3  2  2  2  2  5  4  4]
 [ 3  4  6  5  6  6  6  5]
 [ 6  6  6  7  9  8  6  7]
 [ 9  7  6  6  8 11  8  9]
 [10 10 10 10 10  6  8 11]
 [12 11 10 12  9 10 10 10]]
[Pillow] Quantization table
 [[1 1 1 1 1 1 1 1]
 [1 1 1 1 1 1 1 1]
 [1 1 1 1 1 1 1 1]
 [1 1 1 1 1 1 1 1]
 [1 1 1 1 1 1 1 1]
 [1 1 1 1 1 1 1 1]
 [1 1 1 1 1 1 1 1]
 [1 1 1 1 1 1 1 1]]
```

btlorch:
Removes the line that previously allowed a maximum JPEG quality factor of 95. Fixes #571.

### DIFF
--- a/imageio/plugins/pillow.py
+++ b/imageio/plugins/pillow.py
@@ -485,7 +485,6 @@ class JPEGFormat(PillowFormat):
             quality = int(quality)
             if quality < 1 or quality > 100:
                 raise ValueError("JPEG quality should be between 1 and 100.")
-            quality = min(95, max(1, quality))
 
             kwargs["quality"] = quality
             kwargs["progressive"] = bool(progressive)
--- a/imageio/plugins/pillow.py
+++ b/imageio/plugins/pillow.py
@@ -481,10 +481,10 @@ class JPEGFormat(PillowFormat):
     class Writer(PillowFormat.Writer):
         def _open(self, quality=75, progressive=False, optimize=False, **kwargs):
 
-            # Check quality - in Pillow it should be no higher than 95
+            # The JPEG quality factor can be between 0 (worst) and 100 (best)
             quality = int(quality)
-            if quality < 1 or quality > 100:
-                raise ValueError("JPEG quality should be between 1 and 100.")
+            if quality < 0 or quality > 100:
+                raise ValueError("JPEG quality should be between 0 and 100.")
 
             kwargs["quality"] = quality
             kwargs["progressive"] = bool(progressive)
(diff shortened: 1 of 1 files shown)


almarklein:
Thanks, that was quick :)    Could you please also update that (now invalid) comment just above?

btlorch:
Sorry, I overlooked that comment earlier. According to the [Pillow docs](https://pillow.readthedocs.io/en/stable/handbook/image-file-formats.html#jpeg), the lowest possible quality factor is 0, so I would also allow a value of 0.
libjpeg doesn't recommend quality factors above 95, that's probably why imageio had quality factors > 95 disabled earlier. However, I believe this decision should be left to the user.

almarklein:
Thanks for the fix!

btlorch:
Thanks for pointing me towards the relevant lines!

PR #583, opened 2021-01-05
Update README.md

charun80:
FLIF imagio plugin added to list of "Optional Python packages "


almarklein:
For the record, the other packages in this list are optional dependencies, and their plugins are part of imageio itself. The `imageio-flif` lib implements a plugin by itself. (And I think we should do this more instead of the former.)

almarklein:
Thanks!

PR #588, opened 2021-01-26
Adds fuzzing for integration with OSS-Fuzz

DavidKorczynski:
Hi Maintainers,

I would like to set up continuous fuzzing of imageio by way of OSS-Fuzz. Essentially, OSS-Fuzz is a free service run by Google that performs continuous fuzzing of important open source projects. OSS-Fuzz recently added support for Python and it would be great to have imageio integrated. The only expectation of integrating into OSS-Fuzz is that bugs will be fixed. This is not a "hard" requirement in that no one enforces this and the main point is if bugs are not fixed then it is a waste of resources to run the fuzzers, which we would like to avoid.

In this PR https://github.com/google/oss-fuzz/pull/4993 I have done exactly that, namely created the necessary logic from an OSS-Fuzz perspective to integrate imageio. If you would like to integrate, could you please provide a set of email(s) that will get access to the data produced by OSS-Fuzz, such as bug reports, coverage reports and more stats. The emails should be linked to a Google account in order to view the detailed reports and notice the emails affiliated with the project will be public in the OSS-Fuzz repo, as they will be part of a configuration file.

### DIFF
--- a/tests/fuzzers/fuzz_read.py
+++ b/tests/fuzzers/fuzz_read.py
@@ -1,0 +1,26 @@
+#!/usr/bin/python3
+import os
+import sys
+import atheris
+
+import imageio
+
+def TestOneInput(data):
+    with open("/tmp/img1.file", "wb+") as img1_f:
+       img1_f.write(data)
+    try:
+        imageio.imread("/tmp/img1.file")
+    except ValueError:
+        None
+    except RuntimeError:
+        None
+    os.remove("/tmp/img1.file")
+
+
+def main():
+    atheris.Setup(sys.argv, TestOneInput, enable_python_coverage=True)
+    atheris.Fuzz()
+
+
+if __name__ == "__main__":
+    main()


almarklein:
Thanks! I am not familiar with fuzzing, but it sounds like it could be very useful. 

> could you please provide a set of email(s) that will get access to the data produced by OSS-Fuzz, such as bug reports, coverage reports and more stats. 

You can use almar.klein at gmail.com

So how does this PR fit in exactly? I see that in the PR at OSS-fuzz very similar code is added.

DavidKorczynski:
> Thanks! I am not familiar with fuzzing, but it sounds like it could be very useful.
> So how does this PR fit in exactly? I see that in the PR at OSS-fuzz very similar code is added.

I will remove the fuzzer from the OSS-Fuzz repository (so it only exists here) - the reason I placed it there as well was to verify things worked with their CI and also in case you weren't happy with having the fuzzers in the imageio repository we could place them there.

The way it works is essentially that OSS-Fuzz will run servers that will clone the imageio repository and then run the fuzzer (by way of [Atheris](https://github.com/google/atheris)) continuously. If an uncaught exception is found then a bug report will be filed and you will get the information. Essentially, the Atheris fuzzer will send a lot of random input into the `TestOneInput` function in order to stress test the API. 

We can also add a CI integration such that the fuzzer will be run each time a PR happens for X amount of minutes. 

Does this make more sense now? Otherwise please ask me to elaborate and also let me know if you would CI integration as well.

almarklein:
Thanks for the explanation! I think the continuously running fuzzer should be fine for now.

almarklein:
And thanks for helping set this up 👍

DavidKorczynski:
Thanks! I have now completed set up the OSS-Fuzz infrastructure, however, there is an uncaught exception when an empty buffer is being passed to the fuzzer entry, i.e. when `imread` reads an empty file. This causes the fuzzer to not begin, since it always does a test run with an empty seed first.

The CI stack trace is here https://github.com/google/oss-fuzz/pull/4993/checks?check_run_id=1822162410
```
 === Uncaught Python exception: ===
error: unpack_from requires a buffer of at least 4 bytes for unpacking 4 bytes at offset 0 (actual buffer size is 1)
  File "fuzz_read.py", line 12, in TestOneInput
  File "imageio/core/functions.py", line 265, in imread
  File "imageio/core/functions.py", line 178, in get_reader
  File "imageio/core/format.py", line 690, in search_read_format
  File "imageio/core/format.py", line 188, in can_read
  File "imageio/plugins/pillow.py", line 106, in _can_read
  File "PIL/BmpImagePlugin.py", line 54, in _dib_accept
  File "PIL/_binary.py", line 57, in i32le

==37== ERROR: libFuzzer: fuzz target exited
SUMMARY: libFuzzer: fuzz target exited
MS: 0 ; base unit: 0000000000000000000000000000000000000000
0xa,
\x0a
artifact_prefix='./'; Test unit written to ./crash-adc83b19e793491b1c6ea0fd8b46cd9f32e592fc
Base64: Cg==
```

Is this due to a bug in the fuzzer? i.e. what error exceptions are thrown when `imread` fails?

almarklein:
If I try that, I get a ValueError from imageio telleing me that no format could be found. The error in CI comes from Pillow. We can't control what types of errors are thrown there. But perhaps we should catch it in imageio in this case, so we can fail in a better way.

DavidKorczynski:
> If I try that, I get a ValueError from imageio telleing me that no format could be found. The error in CI comes from Pillow. We can't control what types of errors are thrown there. But perhaps we should catch it in imageio in this case, so we can fail in a better way.

Yes, I think it would be nice if we could fail in a more graceful way. We are already fuzzing Pillow here: https://github.com/google/oss-fuzz/tree/master/projects/pillow and Pillow does not promise any specific errors so the calls to Pillow can sample catch `Exception`. 

Would you like me to set up these fixes or would you prefer to do it yourself?

almarklein:
Yeah, that would be a good workaround for now. If you could make these changes that would be great!

DavidKorczynski:
> Yeah, that would be a good workaround for now. If you could make these changes that would be great!

Sounds good - I fixed up two places that caused issues in terms of not catching errors in pillow in this PR https://github.com/imageio/imageio/pull/592 
The PR also makes the fuzzer run, so it should start to report to us more where there might be other places.

almarklein:
I thought you meant to catch more errors in the test script for the Fuzzer. The approach you proposed is actually better, but we'd need to move the code closer to the Pillow-specific code (imageio uses multiple plugins, of which Pillow is one). See #593

DavidKorczynski:
> I thought you meant to catch more errors in the test script for the Fuzzer. The approach you proposed is actually better, but we'd need to move the code closer to the Pillow-specific code (imageio uses multiple plugins, of which Pillow is one). See #593

Ah cool - and yes I kind of realised it wasn't as simple as I had first anticipated once the tests started failing in my PR. Is it reasonable to assume you will carry out https://github.com/imageio/imageio/pull/593 and then ping back here when it's ready for me to do fuzzer changes? From a fuzzing point of view, the key goal is really to have predictable exceptions, i.e. we know what type of exceptions a piece of code can throw so we can catch them and then use the fuzzer to find inputs that trigger unexpected exceptions.

almarklein:
I just merged #593. 

> From a fuzzing point of view, the key goal is really to have predictable exceptions, i.e. we know what type of exceptions a piece of code can throw so we can catch them and then use the fuzzer to find inputs that trigger unexpected exceptions.

Yeah, good point, broadening the types of exceptions caught in the fuzzing code would make the fuzzing much less effective.

We should also consider testing more API surface with the fuzzer eventually. But we also have an API change with quite a bit of refactoring coming up, so maybe wait until that's done ...

DavidKorczynski:
We still get an issue on this line:

https://github.com/imageio/imageio/blob/6f927e4efa58874ecc4ee4d6ea90eda4e454257f/imageio/plugins/pillow.py#L112

This is because the `accept` call may fail in Pillow, e.g.
```
 === Uncaught Python exception: ===
error: unpack_from requires a buffer of at least 4 bytes for unpacking 4 bytes at offset 0 (actual buffer size is 1)
  File "fuzz_read.py", line 13, in TestOneInput
  File "imageio/core/functions.py", line 265, in imread
  File "imageio/core/functions.py", line 178, in get_reader
  File "imageio/core/format.py", line 690, in search_read_format
  File "imageio/core/format.py", line 188, in can_read
  File "imageio/plugins/pillow.py", line 112, in _can_read
  File "PIL/BmpImagePlugin.py", line 54, in _dib_accept
  File "PIL/_binary.py", line 57, in i32le

```

almarklein:
Oh shoot, of course. My mistake! Will fix right now.

almarklein:
fixed, probably :)

DavidKorczynski:
> Oh shoot, of course. My mistake! Will fix right now.

Wohooo - CI for the fuzzer passes: https://github.com/google/oss-fuzz/pull/4993 :)!

Thanks for engaging with getting this fixed today. I think some other instances of these bugs will be reported by the fuzzer, so you should shortly get emails with some reports. Do feel free to ping me here when you have performed the refactoring mentioned above and are ready to do more API surface fuzzing.

DavidKorczynski:
@almarklein FYI the PR has been merged and it's all integrated into OSS-Fuzz now!

PR #592, opened 2021-02-12
Wrap pillow calls in try-catch.

DavidKorczynski:
This PR should make the fuzzers run without crashing on pillow calls. 

Cross-referencing https://github.com/imageio/imageio/pull/588#issuecomment-778087743


almarklein:
Supersceded by #593

PR #574, opened 2020-12-01
new public API

FirefoxMetzger:
Closes: #569 (Discussion for a new public API in v3.0.0)

As a first step towards the new API, I am putting together a PR to test/demonstrate backward compatibility. The idea is to introduce both, the `imopen` factory as well as a `legacy` plugin. This plugin serves two functions: 

1. it is called by the old API for backward compatibility, 
2. it acts as a fallback to interface with legacy plugins while existing plugins are being updated to take advantage of the new API design.

We could introduce the new API in v2.10.0 and default to the old API with an optional switch into the new API. Starting with v3.0.0 the new API would replace the old one. How much of the old API we want to keep around at that point is open for discussion.

For future reference, a link to a comment in this issue that I keep citing: https://github.com/imageio/imageio/pull/574#issuecomment-808789326

PR #595, opened 2021-02-16
Add CI via GH Actions

almarklein:
* [x] Translate travis yaml to GH actions yaml.
* [x] Get it working.
* [x] Remove Travis and Appveyor files.
* [x] Also test IMAGEIO_NO_INTERNET.
* [x] Maybe tweak invoke tasks.
* [x] Search code for other refs to Travis (e.g. env variables in the tests)


FirefoxMetzger:
awesome! I'll give this a closer look on Thursday evening; I should have time for some coding then.

almarklein:
This is ready from my end. Feel free to merge (please do a squash merge!)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,125 @@
...
+            python setup.py sdist
+            # Install again (with images)
+            pip install .
+            # Delete the local copy, so we're actually using the installed version
+            rm -rf ./imageio ./build ./egg-info
+
+      - name: Install optional dependencies for tests
+        if: matrix.TEST_FULL == 1
+        run: |
+            pip install -q -U simpleitk astropy
```
FirefoxMetzger, line 95:
```suggestion
            pip install -q -U simpleitk
```

If you agree with the suggestion above, then astopy will already be installed when installing optional dependencies.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,125 @@
...
+            invoke docs --clean --build
+
+      - name: PR size check
+        uses: ookami-kb/gh-pr-size-watcher@v1
+        if: matrix.TEST_STYLE == 1
+        with:
+          githubToken: ${{ secrets.GITHUB_TOKEN }} # required
+          errorSize: 500
+          warningSize: 300
+          excludeTitle: 'PR_SIZE_SKIP'
```
FirefoxMetzger, line 115: I'm not sure I get the idea behind this action; it could be that I misunderstand the step. Why would we want to fail if the PR has more than 500 lines of code?

setup.py
```diff
@@ -226,9 +223,9 @@ def run(self):
         "Operating System :: Microsoft :: Windows",
         "Operating System :: POSIX",
         "Programming Language :: Python :: 3",
-        "Programming Language :: Python :: 3.5",
         "Programming Language :: Python :: 3.6",
         "Programming Language :: Python :: 3.7",
         "Programming Language :: Python :: 3.8",
+        "Programming Language :: Python :: 3.9",
```
FirefoxMetzger, line 229: are we using this commit to bump the supported python versions? I think this warrants increasing the version-minor. v2.9 is available on python 3.5, but not on 3.9. v2.10 is no longer on python 3.5 but now also on 3.9

tasks/test.py
```diff
@@ -21,20 +21,18 @@ def lint(ctx):
         unit="run unit tests (pytest)",
         installed="like unit, but prefer using installed package, "
         + "and do not allow the use of internet.",
-        style="run style tests (flake8)",
+        style="run style tests (flake8 and black)",
```
FirefoxMetzger, line 24: I'm a little confused by our dependency on black. Are we enforcing it somewhere outside of the tests? Is it required to use black-formatted code for commits? If so, I think we should mention it in the contribution guidelines somewhere.

tests/test_swf.py
```diff
@@ -71,6 +71,14 @@ def test_reading_saving():
     W.set_meta_data({"foo": 3})
     W.close()
 
+    # Just make sure mimread works
+    assert len(imageio.mimread(fname1)) == 10
+
+    # I'm not sure why, but the below does not work on pypy, which is weird,
+    # because the file *is* closed, but somehow it's not flushed? Ah well ...
+    if IS_PYPY:
+        return
```
FirefoxMetzger, line 80: I guess we can leave this for the current PR (doesn't seem like a high priority), but could you create an issue referencing this, so that we can look into and fix this test later?

FirefoxMetzger (commented):
That's some solid work there @almarklein !

I added some comments, a few of which are questions to improve my understanding of the codebase. There's also some high-level items that didn't fit a specific line:

1. I noticed that we quite frequently branch on `matrix.TEST_STYLE == 1`. It might be cleaner to have a separate job for lining and style and one that just runs the unit tests. If you agree, I can look into this. (I guess members have rights to add to PRs?)

2. I also just realized that we have an implicit dependency on `invoke` which is not declared in `setup.py`. My suggestion would be to use `extra_requires` and maintain optional dependencies specific for test/dev. This should also make our CI config easier.

3. Another high-level question (for me to better understand the codebase): What is the intention behind the `tasks/clean.py`? I mean, it doesn't hurt to be able to wipe all caches, but when do we actually need/want it?

tests/test_meta.py
```diff
@@ -116,9 +116,7 @@ def test_import_dependencies():
     than the known dependencies.
     """
 
-    # Skip when we know there's more imports
-    if os.getenv("TRAVIS") and os.getenv("TEST_FULL"):
-        return
+    # todo: Skip when we know there's more imports?
```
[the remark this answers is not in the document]
almarklein: Well, we skipped it before, and I was kinda surprised that it kept working when I removed the skip, so I put back this comment to indicate that w could expect the need to skip here for some builds. I will rewrite the comment to make this more clear.

tests/test_core.py
```diff
@@ -124,14 +124,14 @@ def test_findlib2():
         os.mkdir(fi_dir)
     dirs, paths = core.findlib.generate_candidate_libs(["libfreeimage"], [fi_dir])
     # assert fi_dir in dirs -> Cannot test: lib may not exist
-    assert paths
+    # assert paths
```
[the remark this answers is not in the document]
almarklein: will remove

tests/test_core.py
```diff
@@ -124,14 +124,14 @@ def test_findlib2():
...
 
     open(os.path.join(fi_dir, "notalib.test.so"), "wb")
 
     # Loading libs
     gllib = ctypes.util.find_library("GL")
     core.load_lib([gllib], [])
-    core.load_lib([], ["libfreeimage"], [fi_dir])
+    # core.load_lib([], ["libfreeimage"], [fi_dir])
```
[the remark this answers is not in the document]
almarklein: Would be good to eventually refactor this test a bit. But we can leave that for another day, e.g. when we roll the freeimage out into its own package, like we did with imageio-ffmpeg (which is something worth doing, I think, because it is the only thing being not pure Python in imageio).

For now, I'll just delete the comment.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,125 @@
+name: CI
+
+
+on:
+  push:
+    branches: [ master ]
```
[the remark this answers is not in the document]
almarklein: Mmm yeah ... I think this to be able to "check the status on master". You can have multiple PR's open, each with a failing CI, but you can still check that the current master is OK. And indeed see that with the badge in the readme :)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,125 @@
...
+            # Install imagio-ffmpeg because its optional but we still want to test it
+            pip install -q -U imageio-ffmpeg
+            # The first local install is really to get the dependencies (as specified in setup.py)
+            pip install .
+            # This step is to force the download of standard images
+            python setup.py sdist
+            # Install again (with images)
+            pip install .
+            # Delete the local copy, so we're actually using the installed version
+            rm -rf ./imageio ./build ./egg-info
```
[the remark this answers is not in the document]
almarklein: Makes sense to use the `[extra]` option, but in that case I would do that in the other step where we install optional dependencies (which only runs for one build), because it can potentially slow.

The imageio-ffmpeg is a bit of an exception. We can either change the test so that it's skipped, or manually add it here. I was lazy and opted for the latter for now.

It took me quite some time to get it to working with the standard images. It could well be that we can run the download-script thing here, which would be nicer than the current solution. But for the moment I'm just happy that this works :D

.github/workflows/ci.yml
almarklein: We had something similar in the Travis CI. It's basically a guard for accidentally committing unintended huge files, which will then be in your Git history forever. I'm not even sure if this would work for binary files though, since it counts lines. Maybe we should just ditch it ...

setup.py
almarklein: I'm pretty sure that the past versions already supported Py39, even though our CI did not test that yet. Somewhere along the line (when CI was down?) we had a contribution that included type annotations, breaking support for 3.5. I think that's ok though, since all major Scipy packages support 3.6 and up, I believe.

Anyway, yeah, we should bump the version number before we release. If we do a release before the API changes ...

tasks/test.py
almarklein: Well, we do mention in the readme to run `invoke format`, which runs black. I'll tweak the text to mention Black.

tests/test_swf.py
almarklein: https://github.com/imageio/imageio/issues/598  - I did not spend too much time on this, since SWF is not very useful anymore since Flash is out of fashion. But it would be worth verifying whether this issue is specific to SWF or affects all multi-image writing.

.github/workflows/ci.yml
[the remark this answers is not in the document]
almarklein: Ok, I refactored this bit using `pip install .[dev]` etc.

almarklein:
>     1. I noticed that we quite frequently branch on `matrix.TEST_STYLE == 1`. It might be cleaner to have a separate job for lining and style and one that just runs the unit tests. If you agree, I can look into this. (I guess members have rights to add to PRs?)

That sounds good! I'm probably not using GH Actions as effectively as it could :)
 
>     2. I also just realized that we have an implicit dependency on `invoke` which is not declared in `setup.py`. My suggestion would be to use `extra_requires` and maintain optional dependencies specific for test/dev. This should also make our CI config easier.

Good point. Done!

>     3. Another high-level question (for me to better understand the codebase): What is the intention behind the `tasks/clean.py`? I mean, it doesn't hurt to be able to wipe all caches, but when do we actually need/want it?

It's really that. There can be quite a lot of directories and files that certain tools left. I like to clean that from time to time :)

almarklein:
>I guess members have rights to add to PRs?

members of `maintainers` like you, have full access.

setup.py
```diff
@@ -172,16 +169,18 @@ def run(self):
         sdist.run(self)
 
 
+install_requires = ["numpy", "pillow"]
+
 extras_require = {
+    "dev": ["invoke", "black", "flake8", "pytest", "pytest-cov", "psutil"],
+    "docs": ["sphinx", "numpydoc"],
+    "ffmpeg": ["imageio-ffmpeg"],
```
FirefoxMetzger, line 177: I really like this reorganization. A thought: afaik there is only `imageio-ffmpeg` as an additional dependency for more formats. In the future, there could be more. Could/would it make sense to add a key called `full` (naming suggestions welcome) so that we can install all formats via `pip install imageio[full]`? Whereas `pip install imageio` would only install commonly used ones (I guess pillow?)

FirefoxMetzger (commented):
Here is the next round :)

I was porting tests for #574 just now, and noticed that there are a lot of nicities in `pytest` that we aren't taking advantage of currently. I can do the same for the core tests (which I probably should before merging #574 ). I'll put that aside for now and look into a suggestion to refactor the CI flow into separate jobs for style/doc/unit testing.

I'll also open a new issue to discuss how to go about moving different plugins (as a backlog for after merging the new API).

> Makes sense to use the [extra] option, but in that case I would do that in the other step where we install optional dependencies (which only runs for one build), because it can potentially slow.

It is possible to cache stages: https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows

I can follow this up once our CI pipeline becomes too slow.

FirefoxMetzger:
I'll have to stop for today; tomorrow I will double-check the mac and windows builds for pypy. I think the condition for the MacOS fix doesn't work and it applies the fix to every build. I'll also see if I can get a similar fix for windows.

Comments are welcome @almarklein .

FirefoxMetzger:
I've decided to stop looking into the Windows + pypy build. I gave myself a 2-hour time limit for this, and I couldn't get the runner to use the pre-compiled wheel, and building from source seems more involved. If pypy + Windows is relevant for somebody, I'm happy to make this a new issue and investigate in a new PR.

FirefoxMetzger:
@almarklein I'm happy with the tests now. Please have a look and let me know if we should make any further changes.

README.md
```diff
@@ -2,8 +2,7 @@
 
 [![PyPI Version](https://img.shields.io/pypi/v/imageio.svg)](https://pypi.python.org/pypi/imageio/)
 [![Supported Python Versions](https://img.shields.io/pypi/pyversions/imageio.svg)](https://pypi.python.org/pypi/imageio/)
-[![Build Status](https://travis-ci.org/imageio/imageio.svg?branch=master)](https://travis-ci.org/imageio/imageio)
-[![Coverage Status](https://coveralls.io/repos/github/imageio/imageio/badge.svg?branch=master)](https://coveralls.io/github/imageio/imageio?branch=master)
```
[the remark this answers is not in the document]
almarklein: I did not integrate it with the current CI yet. Would be good to try adding it back though.

setup.py
almarklein: There already is an `imageio[full]`, defined slightly further down :)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,137 @@
...
+      - name: Build Docs
+        run: |
+            pip install -e .[dev,docs]
+      - name: Test docbuilding
+        run: |
+            invoke docs --clean --build
+      # TODO: Upload the docs to some place
+      # where they can be served. Potentially build
+      # this in a different CI tool that can serve
+      # html artifacts
```
almarklein, line 29: No need; we use RTD

.github/workflows/ci.yml
```diff
@@ -0,0 +1,137 @@
...
+
+  pr-size:
+    # Note: Might be deleted / unwanted
+    name: "PR File Size"
```
almarklein, line 33: Yeah, lets remove this

setup.py
```diff
@@ -172,16 +169,19 @@ def run(self):
         sdist.run(self)
 
 
+install_requires = ["numpy", "pillow"]
+
 extras_require = {
+    "linting": ["black", "flake8"],
+    "dev": ["invoke", "pytest", "pytest-cov"],
```
almarklein, line 176: Why split out linting? Anyone who does some dev work will also want to run `invoke format` (which requires black), and probably lint as well.

setup.py
```diff
@@ -172,16 +169,19 @@ def run(self):
...
+    "docs": ["sphinx", "numpydoc"],
+    "ffmpeg": ["imageio-ffmpeg", "psutil"],
```
almarklein, line 178: Nice catch - I'd assumed psutil was needed by the tests

.github/workflows/ci.yml
almarklein: > My vote would go to removing this step.

Agreed

almarklein:
Thanks for the additions!

README.md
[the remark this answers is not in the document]
FirefoxMetzger: Hm. Would this be an opportunity to switch from coveralls to codecov? I don't have a preference for either. I did notice that codecov seems much more popular, and their web interface looks a lot cleaner/modern. They also have a special action for integration, so it should pretty much be a one-liner to set it up in our workflow.

setup.py
FirefoxMetzger: Ah! I missed that one. My bad.

.github/workflows/ci.yml
FirefoxMetzger: Yes, and I would keep RTD for the docs for each version on `master`. I was thinking that it might be nice to also do this for each PR individually.

We are already building the docs as a test and it might be nice to also upload them somewhere, so we can inspect the formatting in the actual PR before merging it with `master`.

.github/workflows/ci.yml
FirefoxMetzger: done. PR File Size is no more.

setup.py
FirefoxMetzger: because black doesn't build on `pypy` and this was the fastest way to separate the lining dependencies and the testing dependencies.

I think a nice change (if you agree) would be to rename `dev` to `test` and add a new `dev` line, which combines `linting` and `testing`. This way the workflows can install `test`/`lint` respectively while developers can install `dev`

setup.py
FirefoxMetzger: you are right, too. `psutil` is needed exclusively for the tests of `imageio-ffmpeg`. I didn't want to introduce another `test-ffmpeg` target, so I combined it with `ffmpeg`.

FirefoxMetzger:
@almarklein I feel a bit uncertain regarding the assignment of this PR. Will you add the current changes for this round, and I will review, or should I add the changes and you review?

README.md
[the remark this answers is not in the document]
almarklein: I have no preference, if codecov is easy to integrate, than let's go with that :)  Shall we do that after this PR though?

setup.py
almarklein: Cool, yeah let's do that.

almarklein:
>  I feel a bit uncertain regarding the assignment of this PR. Will you add the current changes for this round, and I will review, or should I add the changes and you review?

I think there's not too much left now. Could you pleas fix up the `[dev]  <- [lint,test]` thing, then I think we're good to go. Let's add codecov in a new pr.

.github/workflows/ci.yml
FirefoxMetzger: actually, RTD seems to offers this feature: https://docs.readthedocs.io/en/stable/pull-requests.html

I don't think we need to add it in this PR, but we may want to keep the comment there for the future. If we decide to add it this feature in a new PR, I'd like to open an issue about it so that it is in the backlog.

almarklein: Sounds good!

FirefoxMetzger:
Done. Let's see what the test say and if that checks out then I am happy from my side; we could continue with a bugfix in the legacy pillow plugin (#601 ).

almarklein:
Cool! Feel free to (squash) merge!

PR #596, opened 2021-02-16
Add webm to list of supported ffmpeg formats

Skylion007:
FFMPEG has a webm muxer and can read and write to the format. We should support it so that we can read and write to the the format directly.


FirefoxMetzger:
Seems like a good idea indeed. I will have a look at this after #595 is merged :)

almarklein:
This looks good to me! Thanks!

FirefoxMetzger:
I've added our CI pipeline to the PR.

Skylion007:
@FirefoxMetzger Is this a false positive? I am not sure why adding this format would trigger an OS error

Skylion007:
Yeah, tests are bit flaky. ;)

Skylion007:
@almarklein @FirefoxMetzger awaiting merge.

Skylion007:
Should I add a changelog entry?

FirefoxMetzger:
@Skylion007 Yes that would be nice. Other than this, the PR looks good, and I will merge it towards the end of the week.

The CI pipeline breaking sporadically is not a good sign (it's new and shouldn't do that xD). I suspect that there may be a missing codec on windows that wasn't on one test runner but was available on the other. That's essentially why I didn't merge this already because I want to double-check that all the dependencies are tracked properly first. I'll have to dig into the logs to know this for sure though, which will happen after I completed setting up my new PC.

PR #601, opened 2021-02-21
Enable `pilmode` for GIF format

### LINKED ISSUE #600: Reading GIF returns no color channels axes
I think imageio sometimes does not read compressed GIFs correctly.

I would expect the output of `imageio.mimread('animation.gif')` to always return an array with four axes for time, width, height, and channels. However, some of the GIFs I'm trying to process return an array with only three dimensions.

This can be reproduced with [this example GIF](https://user-images.githubusercontent.com/2111293/108618859-78714800-73ef-11eb-8928-03e8dfb8581a.gif) and also with a solid-color GIF as shown in the minimal example below:

```python
import imageio
import numpy as np

imageio.mimsave('animation.gif', np.zeros((20, 64, 64, 3), np.uint8))
np.array(imageio.mimread('animation.gif')).shape  # (20, 64, 64)
```

FirefoxMetzger:
Closes #600 .

The previous stub that was used as a reader for the gif format didn't support the `pilmode` argument. This PR fixes it by removing the stub, as it wasn't offering anything on top of the base reader class.

Should be merged after #595 .


FirefoxMetzger:
Assuming the tests check out, I'm happy with this PR. @almarklein could you review and merge it if you are happy?

almarklein:
merged!

PR #606, opened 2021-03-07
delete old files

FirefoxMetzger:
This PR deletes the following files that appear to be no longer in use:

- `Requirements.txt`: We have moved to tracking dependencies in `setup.py`
- `imageio.proxy.py`: Appears to be a patch for functionality akin to `pip install -e`. Maybe pip didn't offer this at the time the file was added; now that it does, I think `pip` should take preference.
- `freezing`: The folder contained two files: (a) a script using `cx_Freeze` which is no longer a requirement for this repo and (b) a file containing absolute paths that will only work on one specific machine. Neither seemed to test code relevant to the library, nor was it executed by `pytest`

@almarklein If I have been too rash in deleting any of these files let me know, and I will restore those that are still relevant. Otherwise, feel free to merge once tests pass :)


almarklein:
Thanks for the cleanup! I'm not 100% sure about the `requirements.txt`, but we'll find out if someone actually depended on that. Better to not duplicate the definition of the dependencies, anyway :)

PR #609, opened 2021-03-13
enable documentation builds for PRs

### LINKED ISSUE #602: Add RTD builds for PRs
Read the docs offers to build documentation for each PR: https://docs.readthedocs.io/en/stable/pull-requests.html

It would be nice if we could have that, to check the styling of the docs before merging them. Currently, docs are tested via the GH Actions CI (build test, no styling). This will likely have to be modified and we may be able to remove this test from GH Actions in favor of using RTD.

FirefoxMetzger:
Closes #602 

I've added the ability to build documentation for PRs over at RTD. This adds a new status to each PR that will display a link to the build logs while still building or on failure. Upon success, the details will link to the temporarily built documentation so that we can visually inspect that it renders properly.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,25 +9,6 @@ on:
 
 
 jobs:
-  docs:
-    name: "Documentation"
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-python@v2
-        with:
-          python-version: '3.8'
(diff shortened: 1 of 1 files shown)


almarklein:
Cool, so this test now basically runs on rtd :)

FirefoxMetzger:
@almarklein Actually, it's more than that, because you can preview the build artifact when you hit "details", which should come in super handy.

almarklein:
Indeed, really nice!

PR #614, opened 2021-03-23
Refactor CI Linting

FirefoxMetzger:
I found a cool snippet in the black docs today that can make our CI a little simpler and thought I propose it.

https://github.com/psf/black#github-actions


FirefoxMetzger:
looks like it is trying to blackify tiffle and other files that are excluded via the invoke route ... do we want to blackify them in this PR and then rebase the new API? @almarklein

almarklein:
> do we want to blackify them in this PR and then rebase the new API?

Yeah we could do that.

However, now you've removed the linting check (flake8).

FirefoxMetzger:
> However, now you've removed the linting check (flake8).

True. I thought that flake8 is included in black, since black acts as both, a linter (fails the job if not formatted properly) and as a formatter (if run with `black .`). I will investigate.

FirefoxMetzger:
If somebody wonders what I'm trying to do here, I'm trying to configure the action in such a way that it will comment on the files changed in the PR (via status annotations) and report exactly where flake8 fails and how.

Here is an example in another repo: https://github.com/liskin/strava-offline/pull/2/files

FirefoxMetzger:
Success! I am happy with this PR now :)

image

FirefoxMetzger:
@almarklein I'm happy with this PR once the tests pass. Feel free to review and comment 😄

almarklein:
> [...] black acts as both, a linter (fails the job if not formatted properly) and as a formatter

Yes, it fails when the code is not formatted as it wants, but flake8 does much more, like checking for undefined and unused variables, unnecessary imports etc.

almarklein:
Looks good! Thanks!

PR #617, opened 2021-04-02
MAINT: Refactor Request.py

FirefoxMetzger:
In preparation to introduce the Requests object in #574 I began rewriting parts of it. While doing so, I've noticed that some of the changes are just porting it to python 3.7+ to make it extensible wrt. `imopen`. These changes exist independently from #574 so I figured we can make this a separate PR.

The idea for this PR is to not introduce any new functionality but to simply refactor existing code.


FirefoxMetzger:
Closed in favor of #622

PR #620, opened 2021-04-03
Temporarily pin version of black action

FirefoxMetzger:
The Black team is preparing for a new release and - for currently unknown reasons - the stable branch of the GH action was deleted in the process.

This breaks our GH Actions workflow, and to work around it we temporarily pin the version to the last stable version until the issue is resolved.

Reference to the issue at psf/black tracking the problem:
https://github.com/psf/black/issues/2079

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
-      - uses: psf/black@stable
+      - uses: psf/black@20.8b1
         with:
           args: ". --check"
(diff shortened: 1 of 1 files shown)


almarklein:
so much for the `stable` branch being stable :P

almarklein:
maybe they use `main` instead?

FirefoxMetzger:
> so much for the stable branch being stable :P

🤷 It was part of the official documentation. I thought that might mean something.

> maybe they use main instead?

Just checked. They don't seem to be using any branch like that. They have `master` for the bleeding edge build, but nothing resembling a `main` or `stable`.

PR #622, opened 2021-04-05
Make Request.Mode an enum

FirefoxMetzger:
To have an easier time integrating the Requests object in #574 I would like to make its mode an enum. In #574 we need to allow the request object to be constructible from a single character (`r`/`w`) while maintaining backward compatibility. I was thinking that this can be done by converting (`r`/`w`) -> (`r?`/`w?`) and allowing new plugins to ignore the second character of the current request mode.

In practice, this would mean that legacy plugins can use the mode as they are used to (pretend that it is a two-character string), while new plugins can do `request.mode.io_mode` and only get the read/write portion. I think `request.mode.io_mode is IOMode.read` is more readable than `request.mode[0] == "r"`. Maybe we can even do the casting implicitly and have something like `request.mode is IOMode.read`.

The idea for this PR is to not introduce any new functionality but to simply refactor existing code.

### DIFF
--- a/imageio/core/format.py
+++ b/imageio/core/format.py
@@ -35,15 +35,13 @@ import sys
 import numpy as np
 
 from . import Array, asarray
+from .request import ImageMode
 
 
-MODENAMES = {
-    "i": "single-image",
-    "I": "multi-image",
-    "v": "single-volume",
-    "V": "multi-volume",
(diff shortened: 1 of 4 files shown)


almarklein:
LGTM. Feel free to merge!

Issue #623, opened 2021-04-06
Performance issues when loading some .tif files


DavidStirling:
I have some 2k x 2k .tif files which I believe are compressed. I noticed that loading these with imageio was very slow.

On further investigation, specifying `format="TIFF-PIL"` with `imread` makes the file load almost instantly, so I believe this may be an issue with tifffile. In keeping with this, installing the latest version of tifffile via `pip` fixes the issue, though the plugin then also asks that `imagecodecs` be installed to load the file:

`ValueError: TiffPage 0: <COMPRESSION.LZW: 5> requires the 'imagecodecs' package`

Oddly enough pillow doesn't seem to need any other package to load the file quickly. Perhaps this is just a matter of updating the packed-in copy of tifffile, but if pillow can do just as a good a job I don't know if you have some way of selecting it when imagecodecs is absent?

Sample image attached. Just calling imread on it should replicate the issue.
[test_image.zip](https://github.com/imageio/imageio/files/6267409/test_image.zip)

Thanks

cgohlke:
The vendored version of tifffile in imageio uses a pure Python implementation of the LZW decompressor. The solution is to install tifffile and imagecodecs.

https://github.com/imageio/imageio/blob/23cdcf5e62aceb1a32ee37e529eb731612b98298/imageio/plugins/_tifffile.py#L8308-L8390

DavidStirling:
That makes sense, thanks for explaining. So is there no way to have imageio switch over to the PIL plugin if the incoming file is compressed and a user doesn't have the extra packages installed?

almarklein:
> is there no way to have imageio switch over to the PIL plugin if the incoming file is compressed and a user doesn't have the extra packages installed?

Imagio uses a plugin system, where a suitable plugin is automatically selected. Including rules like these would make that process much more complex. Tifffile has priority over Pillow for Tiff, because Pillow can only deal with a subset of Tiff. It could be that the vendored version of Tiffile will be removed in future versions, so users will automatically use the latest/faster version.

FirefoxMetzger:
> Or perhaps have some kind of warning about this, so that users know why loading performance isn't optimal.

The concern that I have with a warning here is that it may not be true for every user. I'm not a tifffile user, so I don't know for certain, but my guess is that there is a reason for imageio having a pure-python implementation in a place where tifffile has a dependency. Usually, this is to support as many platforms as possible (and sometimes to support pypy).

[...]

PR #625, opened 2021-04-13
Clarify documentation on .tif handling

DavidStirling:
Following on from discussion in #623, I've tried to document that the tifffile plugin will switch to using an external version of the package if one is present. I've also tried to clarify the scipy migration page, since that gave the impression that Pillow was the default loader for this format.

Hopefully that's okay, do feel free to edit further!

### DIFF
--- a/imageio/plugins/tifffile.py
+++ b/imageio/plugins/tifffile.py
@@ -63,7 +63,8 @@ READ_METADATA_KEYS = (
 
 
 class TiffFormat(Format):
-    """Provides support for a wide range of Tiff images.
+    """Provides support for a wide range of Tiff images using the ``tifffile`` 
+    plugin.
 
     Images that contain multiple pages can be read using ``imageio.mimread()``
     to read the individual pages, or ``imageio.volread()`` to obtain a
@@ -73,6 +74,10 @@ class TiffFormat(Format):
     Thus calling :py:meth:`Format.Writer.set_meta_data` after the first frame
     was written has no effect. Also, global metadata is ignored if metadata is
     provided via the `meta` argument of :py:meth:`Format.Writer.append_data`.
+    
+    If you have installed ``tifffile`` as a Python package, imageio will 
+    attempt to use that instead of the bundled version from this plugin. Doing
+    so can provide access to performance improvements and bug fixes.
 
     Parameters for reading
     ----------------------
--- a/imageio/plugins/tifffile.py
+++ b/imageio/plugins/tifffile.py
@@ -76,7 +76,7 @@ class TiffFormat(Format):
     provided via the `meta` argument of :py:meth:`Format.Writer.append_data`.
 
     If you have installed tifffile as a Python package, imageio will attempt
-    to use that instead of the bundled version from this plugin. Doing so can
+    to use that as backend instead of the bundled backend. Doing so can
     provide access to new performance improvements and bug fixes.
 
     Parameters for reading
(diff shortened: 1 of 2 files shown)


imageio/plugins/tifffile.py
```diff
@@ -63,7 +63,8 @@ def load_lib():
 
 
 class TiffFormat(Format):
-    """Provides support for a wide range of Tiff images.
+    """Provides support for a wide range of Tiff images using the tifffile
```
almarklein, line 66: The CI is complaining about formatting. I could ask to run `black`, but it looks like you made this change from the GH interface, let's try removing whitespace :)

```suggestion
    """Provides support for a wide range of Tiff images using the tifffile
```

imageio/plugins/tifffile.py
```diff
@@ -73,6 +74,10 @@ class TiffFormat(Format):
     Thus calling :py:meth:`Format.Writer.set_meta_data` after the first frame
     was written has no effect. Also, global metadata is ignored if metadata is
     provided via the `meta` argument of :py:meth:`Format.Writer.append_data`.
+    
+    If you have installed tifffile as a Python package, imageio will attempt 
+    to use that instead of the bundled version from this plugin. Doing so can 
+    provide access to new performance improvements and bug fixes.
```
almarklein, line 80: More trailing ws removal.
```suggestion

    If you have installed tifffile as a Python package, imageio will attempt
    to use that instead of the bundled version from this plugin. Doing so can
    provide access to new performance improvements and bug fixes.
```

DavidStirling:
Thanks @almarklein - you're right, I didn't set up a proper environment to add text. Sorry about that.


It looks like the Ubuntu CI is encountering issues now due to one of the Pillow tests, but since the relevant files are unchanged I'm not sure if that's related to this PR?

FirefoxMetzger:
@DavidStirling Yes, it is unrelated. The CI pipeline is a bit flakey, and I would like to investigate this at some point. Unfortunately, I have quite a tight schedule in the coming months, so it likely won't happen before summer.

In any case, I re-ran the CI, and it looks pretty good now :)

imageio/plugins/tifffile.py
```diff
@@ -63,7 +63,8 @@ def load_lib():
 
 
 class TiffFormat(Format):
-    """Provides support for a wide range of Tiff images.
+    """Provides support for a wide range of Tiff images using the tifffile
+    plugin.
```
FirefoxMetzger, line 67:
```suggestion
    backend.
```

Using "plugin" here may be a bit confusing. We happen to refer to the `Format` sub-classes interchangably as plugins or formats; hence, this can be seen as a circular definition. Using "tifffile backend" should make this more clear.

imageio/plugins/tifffile.py
```diff
@@ -74,6 +75,10 @@ class TiffFormat(Format):
     was written has no effect. Also, global metadata is ignored if metadata is
     provided via the `meta` argument of :py:meth:`Format.Writer.append_data`.
 
+    If you have installed tifffile as a Python package, imageio will attempt
+    to use that instead of the bundled version from this plugin. Doing so can
```
FirefoxMetzger, line 79:
```suggestion
    to use that as backend instead of the bundled backend. Doing so can
```

same as above refering to the `tifffile` module as backend should avoid ambiguity.

FirefoxMetzger (approved):
Thank you for these additions 🚀 

I only have a minor comment on terminology (to try and avoid confusion). Other than that I am very happy with the PR.

DavidStirling:
That's great, thanks for your input!

PR #630, opened 2021-05-11
MAINT: new black formatting rules

FirefoxMetzger:
It appears that black has new formatting rules which end up breaking the current CI.

I reformatted the repo using black.


almarklein:
Ah, the periodic black-changed-its-formatting update ;)

PR #637, opened 2021-05-21
Moved fits import above pillow and added unit test of get_reader

### LINKED ISSUE #636: "cannot find loader for this FITS file" error
Hi, I was looking at the `imageio` module and I see that the FITS format is supposedly supported ([here](https://imageio.readthedocs.io/en/stable/formats.html)). But, I get an error when I try to use `imageio.get_reader()` on a file without the `format` keyword set to fits. So, I guess there's some mismatch in how `imagio` is trying to automatically determine the file type.

For a minimum reproducible example, you can stand up this environment:
```
conda create -n testio -c conda-forge scikit-image astropy imageio-ffmpeg pims python=3.7
```

After activating the environment, in Python generate a fake image and try reading it in:
```
import numpy as np
from astropy.io import fits
from astropy.modeling.models import Sersic2D
import imageio
xx, yy = np.meshgrid(np.arange(512),np.arange(512))
mod = Sersic2D(amplitude = 1, r_eff = 25, n=4, x_0=50,y_0=50, ellip=.5, theta=-1)
img = mod(xx, yy)
phdu = fits.PrimaryHDU()
ihdu = fits.ImageHDU(img)
hdul = fits.HDUList([phdu,ihdu])
hdul.writeto('test.fits')
im = imageio.get_reader('test.fits')
```
The last line fails for me with the following error:
```
  File "/local/data/home/[user]/.conda/envs/testio/lib/python3.7/site-packages/imageio/plugins/pillow.py", line 669, in pil_try_read
    im.getdata()[0]
  File "/local/data/home/[user]/.conda/envs/testio/lib/python3.7/site-packages/PIL/Image.py", line 1292, in getdata
    self.load()
  File "/local/data/home/[user]/.conda/envs/testio/lib/python3.7/site-packages/PIL/ImageFile.py", line 327, in load
    raise OSError(f"cannot find loader for this {self.format} file")
OSError: cannot find loader for this FITS file
```

But, I noticed if I give the format like the following line, the `get_reader()` call succeeds:
```
im = imageio.get_reader('test.fits',format='fits')
```

astrophysaxist:
Moved the `import fits` line in `imageio/plugins/__init__.py` to occur before `import pillows` to close #636.  Also added a unit test in `tests/test_fits.py` for `imageio.get_reader()`.

### DIFF
--- a/tests/test_fits.py
+++ b/tests/test_fits.py
@@ -99,7 +99,7 @@ def test_fits_get_reader(tmp_path):
     phdu = fits.PrimaryHDU()
     ihdu = fits.ImageHDU(img)
     hdul = fits.HDUList([phdu,ihdu])
-    hdul.writeto('test.fits')
+    hdul.writeto(tmp_path / 'test.fits')
     try:
         im = imageio.get_reader('test.fits')
         os.remove('test.fits')
--- a/tests/test_fits.py
+++ b/tests/test_fits.py
@@ -26,6 +26,15 @@ def teardown_module():
     imageio.formats.sort()
 
 
+@pytest.fixture
+def normal_plugin_order():
+    # Use fixture to temporarily set context of normal plugin order for
+    # tests and return to module setup afterwards
+    teardown_module()
+    yield
+    setup_module()
+
+
 @pytest.mark.skipif("astropy is None")
 def test_fits_format():
     need_internet()  # We keep the fits files in the imageio-binary repo
@@ -84,13 +93,10 @@ def test_fits_reading():
 
 @pytest.mark.skipif("astropy is None")
 @pytest.mark.skipif("IS_PYPY", reason="pypy doesn't support astropy.fits.")
-def test_fits_get_reader(tmp_path):
+def test_fits_get_reader(normal_plugin_order, tmp_path):
     """Test reading fits with get_reader method
     This is a regression test that closes GitHub issue #636
     """
-
-    # Set precedence as normal
-    imageio.formats.sort()
 
     sigma = 10
     xx, yy = np.meshgrid(np.arange(512), np.arange(512))
--- a/tests/test_fits.py
+++ b/tests/test_fits.py
@@ -1,6 +1,5 @@
 """ Test fits plugin functionality.
 """
-import os
 import pytest
 from imageio.testing import run_tests_if_main, get_test_dir, need_internet
 
--- a/tests/test_fits.py
+++ b/tests/test_fits.py
@@ -1,5 +1,6 @@
 """ Test fits plugin functionality.
 """
+import os
 import pytest
 from imageio.testing import run_tests_if_main, get_test_dir, need_internet
 
@@ -79,5 +80,31 @@ def test_fits_reading():
     im = imageio.imread(compressed)
     assert im.shape == (2042, 3054)
 
+@pytest.mark.skipif("astropy is None")
+def test_fits_get_reader():
+    """Test reading fits with get_reader method"""
+    
+    #Set precedence as normal
+    teardown_module()
+    
+    if IS_PYPY:
+        return  # no support for fits format :(
+
+    import numpy as np
+    from astropy.io import fits
+    sigma = 10
+    xx, yy = np.meshgrid(np.arange(512),np.arange(512))
+    z = (1/(2*np.pi*(sigma**2)))*np.exp(-((xx**2)+(yy**2))/(2*(sigma**2)))
+    img = np.log(z)
+    phdu = fits.PrimaryHDU()
+    ihdu = fits.ImageHDU(img)
+    hdul = fits.HDUList([phdu,ihdu])
+    hdul.writeto('test.fits')
+    try:
+        im = imageio.get_reader('test.fits')
+        os.remove('test.fits')
+    except ValueError:
+        os.remove('test.fits')
+        pytest.fail('imageio.get_reader failed to find FITS loader')
 
 run_tests_if_main()
--- a/tests/test_fits.py
+++ b/tests/test_fits.py
@@ -86,20 +86,23 @@ def test_fits_reading():
 @pytest.mark.skipif("IS_PYPY", reason="pypy doesn't support astropy.fits.")
 def test_fits_get_reader(tmp_path):
     """Test reading fits with get_reader method
-       This is a regression test that closes GitHub issue #636
+    This is a regression test that closes GitHub issue #636
     """
-    
-    #Set precedence as normal
+
+    # Set precedence as normal
     imageio.formats.sort()
-    
+
     sigma = 10
-    xx, yy = np.meshgrid(np.arange(512),np.arange(512))
-    z = (1/(2*np.pi*(sigma**2)))*np.exp(-((xx**2)+(yy**2))/(2*(sigma**2)))
+    xx, yy = np.meshgrid(np.arange(512), np.arange(512))
+    z = (1 / (2 * np.pi * (sigma ** 2))) * np.exp(
+        -((xx ** 2) + (yy ** 2)) / (2 * (sigma ** 2))
+    )
     img = np.log(z)
     phdu = astropy.io.fits.PrimaryHDU()
     ihdu = astropy.io.fits.ImageHDU(img)
-    hdul = astropy.io.fits.HDUList([phdu,ihdu])
-    hdul.writeto(tmp_path / 'test.fits')
-    im = imageio.get_reader(tmp_path / 'test.fits')
+    hdul = astropy.io.fits.HDUList([phdu, ihdu])
+    hdul.writeto(tmp_path / "test.fits")
+    imageio.get_reader(tmp_path / "test.fits")
+
 
 run_tests_if_main()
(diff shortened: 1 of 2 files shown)


tests/test_fits.py
```diff
@@ -79,5 +80,31 @@ def test_fits_reading():
     im = imageio.imread(compressed)
     assert im.shape == (2042, 3054)
 
+@pytest.mark.skipif("astropy is None")
+def test_fits_get_reader():
```
FirefoxMetzger, line 84:
```suggestion
def test_fits_get_reader(tmp_path):
```

We can use the pytest fixture `tmp_path` here to get a temorary directory that will be automatically cleaned up after the test.

tests/test_fits.py
```diff
@@ -79,5 +80,31 @@ def test_fits_reading():
...
+    import numpy as np
+    from astropy.io import fits
+    sigma = 10
+    xx, yy = np.meshgrid(np.arange(512),np.arange(512))
+    z = (1/(2*np.pi*(sigma**2)))*np.exp(-((xx**2)+(yy**2))/(2*(sigma**2)))
+    img = np.log(z)
+    phdu = fits.PrimaryHDU()
+    ihdu = fits.ImageHDU(img)
+    hdul = fits.HDUList([phdu,ihdu])
+    hdul.writeto('test.fits')
```
FirefoxMetzger, line 102:
```suggestion
    hdul.writeto(tmp_path / 'test.fits')
```

We can use above mentioned tmp_path here to create the file inside the temporary directory.

tests/test_fits.py
```diff
@@ -79,5 +80,31 @@ def test_fits_reading():
...
+    try:
+        im = imageio.get_reader('test.fits')
+        os.remove('test.fits')
+    except ValueError:
+        os.remove('test.fits')
+        pytest.fail('imageio.get_reader failed to find FITS loader')
```
FirefoxMetzger, line 108:
```suggestion
    im = imageio.get_reader('test.fits')
```

We can refactor the `os.remove` call, because we can use tmp_path (it is cleaner).

If we couldn't refactor `os.remove`, it would be cleaner to write

```python
try:
    pass # code here
except ValueError:
    pass # handle exception
finally:
    # will _always_ be executed, regardless of success or failure
    os.remove(...)
```

If a test produces an unhandled exception, pytest will mark the test as failed automatically, hence we do not have to specifically tell pytest to fail manually.

tests/test_fits.py
```diff
@@ -79,5 +80,31 @@ def test_fits_reading():
     im = imageio.imread(compressed)
     assert im.shape == (2042, 3054)
 
+@pytest.mark.skipif("astropy is None")
+def test_fits_get_reader():
+    """Test reading fits with get_reader method"""
```
FirefoxMetzger, line 85: Could you mention the GitHub issue (#636 ) that resulted in this regression test in the above comment? This way, it will be clear that it is a regression test, and we also have a reference to its origin,

tests/test_fits.py
```diff
@@ -79,5 +80,31 @@ def test_fits_reading():
...
+    
+    #Set precedence as normal
+    teardown_module()
+    
+    if IS_PYPY:
+        return  # no support for fits format :(
```
FirefoxMetzger, line 91: I am aware that we do the same in other parts of the codebase, but it is actually an anti-pattern; likely from legacy code before pytest had a cleaner way to do things. We have to refactor this sooner or later ;)

Could you remove this and instead add the decorator `@pytest.mark.skipif("IS_PYPY", reason="pypy doesn't support astropy.fits.")`?

tests/test_fits.py
```diff
@@ -79,5 +80,31 @@ def test_fits_reading():
     im = imageio.imread(compressed)
     assert im.shape == (2042, 3054)
 
+@pytest.mark.skipif("astropy is None")
+def test_fits_get_reader():
+    """Test reading fits with get_reader method"""
+    
+    #Set precedence as normal
+    teardown_module()
```
FirefoxMetzger, line 88: Could you explain why we have to tear down any module-level setup to run this test?

tests/test_fits.py
```diff
@@ -79,5 +80,31 @@ def test_fits_reading():
...
+    
+    if IS_PYPY:
+        return  # no support for fits format :(
+
+    import numpy as np
+    from astropy.io import fits
```
FirefoxMetzger, line 94: We generally collect any imports at the top of the module. Could you refactor this and move them to the top?

tests/test_fits.py
```diff
@@ -1,5 +1,6 @@
 """ Test fits plugin functionality.
 """
+import os
```
FirefoxMetzger, line 3:
```suggestion
```

This should no longer be needed (see comments below).

FirefoxMetzger (changes requested):
Nice job! I went through your code and left some comments on how to further improve the code. 

Overall, I think it looks pretty good and we can probably merge after the coming iteration :)

FirefoxMetzger:
Ah, the linting test fails. We strictly enforce the `black` coding style here, and also ask that a PR resolves any `flake8` issues.

To make sure the code is `black` compliant, simply run `black .` at the root of the repository.

To check for `flake8` issues run `flake8` at the root of the repository. Each problem will be listed in a new line (with reference to where it occurs). Unfortunately, they aren't fixed automatically, but it is often clear what is amiss. If the command returns without printing anything it means everything is fine :)

tests/test_fits.py
astrophysaxist:
```suggestion
    hdul.writeto(tmp_path+'/test.fits')
```

astrophysaxist: Looks like `tmp_path /` is pytest syntax for temporary directories. So, I'll delete my previous suggestion and commit.

tests/test_fits.py
astrophysaxist: Thanks for that suggestion! Part of the reason I ended up adding in the explicit fail was I hadn't realized the default format was set for `FITS` up earlier in the test file, so I was not testing in the right precedence scenario and was not getting a failure when I expected. Eventually saw the `teardown_module` function and was able to use it. But didn't refactor.

tests/test_fits.py
astrophysaxist: Ah, I addressed this point on the `tmp_path` additions. The setup changes the precedence from normal to being default FITS format, so my test was always passing. I noticed the `teardown` function and decided to take advantage of it to set precedence levels back to normal, where the original issue occurs.

tests/test_fits.py
astrophysaxist: Just made this change in the most recent commit.

tests/test_fits.py
astrophysaxist: Added this decorator as well.

tests/test_fits.py
astrophysaxist: Done!

tests/test_fits.py
astrophysaxist: In the most recent commit I removed the `teardown_module` call and I just use the `imageio.formats.sort()` line to return the format precedence to normal for the test.

tests/test_fits.py
FirefoxMetzger: @astrophysaxist exactly. You can find the docs here: https://docs.pytest.org/en/6.2.x/tmpdir.html#the-tmp-path-fixture

tests/test_fits.py
FirefoxMetzger: I see, thanks for the explanation. `teardown_module` seems like a very big hammer for this. 

In another pending PR we are using the module-scope to download and generate a bunch of test images that can then easily be copied into a local folder for a test to consume. Tearing down the module may have unwanted side-effects here, such as forcing us to regenerate the files (which is expensive).

`imageio.formats.sort()` might also persist after the test has finished, since `imageio.formats` is a singleton and (I think) that libraries aren't reloaded between tests. I'll dig into this tomorrow and see if further changes are needed.

FirefoxMetzger: Okay, I went through the code. Both `teardown_module` and `imageio.formats.sort()` won't do the trick here, because they will both leave the environment in the wrong state after the test completes.

Instead, we need a new fixture that _temporarily_ reverts the order back to normal. It will look something like this

```python
@pytest.fixture
def normal_plugin_order():
    teardown_module()
    yield
    setup_module()
```

This will run the test using the normal plugin order and revert to the module's default state after the test is finished. It should be a fixture because that will take care of most corner cases for us, e.g. if the test crashes with an exception the environment will still be restored correctly.

We have to use that fixture by updating the test's signature `def test_fits_get_reader(normal_plugin_order, tmp_path):`.

@astrophysaxist Is this explanation sufficient to describe what has to be modified?

FirefoxMetzger:
I took the liberty to run the code through our linter; now the linting test is green as well, meaning that all the tests are happy :)

tests/test_fits.py
astrophysaxist: Yep that makes perfect sense @FirefoxMetzger . Happy to see pytest has this capability. Learned a lot about pytest functionality with this! Just committed those changes.

FirefoxMetzger (approved):

FirefoxMetzger:
LGTM! Thanks for this contribution, and for being so responsive regarding the changes :)

Issue #633, opened 2021-05-14
Use-Case based lists of common formats/plugins


FirefoxMetzger:
A while back @almarklein and I were discussing that we want to separate imageio and its backends (pillow, tifffile, ffmpeg, ...). To keep things convenient, we mused about creating a curated list of extra dependencies that would be used to install backends for a specific subject area. For example `pip install imageio[microscopy]` would install all backends related to microscopy.

To get this process rolling (and for us to get an overview of what plugin does what - and ultimately make the docs prettier) I created a spreadsheet listing all the extensions that imageio currently supports combined with the respective plugin for that extension. I've also added a suggestion for an initial set of install targets. Namely:

- *common*: Add support for commonly used every-day formats (.png, .gif, .jpg, ...)
- *microscopy*: Add support microscopy formats (.tiff, etc.)
- *video*: Add support video formats (.avi, .mp4, etc.)
- *all*: Install all backends

Here is a list to the spreadsheet: https://docs.google.com/spreadsheets/d/1na7fVHtdixXycRduPD0dGYdaLg6gEKR2dBW5LU9-FQk/edit?usp=sharing

Please let me know what you think. If I don't hear too large protests, I will go ahead and create a PR reflecting this suggestion.

almarklein:
I'm leaning towards making Pillow and tiffile standard dependencies, and making the freeimage plugin non-default. Otherwise this looks good to me. Nice work on the matrix spreadsheet :)

FirefoxMetzger:
By *standard* dependencies you mean that they would be installed via `pip install imageio[common]`? I agree with pillow, since it is very popular and supports everyday formats. I'm not sure I understand tifffile; pillow already offers basic tif support which makes tifffile seem like a specialist backend.

If you are thinking about core dependencies (`pip install imageio`), I would install neither pillow, nor tifffile. This would make us depend on numpy only, which should give a significant boost in flexibility.

kmilos:
I vote against narrow category naming like _microscopy_, higher bit depth and various compression TIFFs are used in **many** domains. Btw, `tifffile` will also happily read/interpret DNGs and some other TIFF-based photography raw files...

FirefoxMetzger:
> I vote against narrow category naming like microscopy,

Sure, any alternative suggestions? 

You will always be able to get any backend combination you want via something line `pip install imageio[freeimage, tifffile, spe]`, that is if your platform supports the respective backends. You can also add backends via good old `pip install tifffile`, etc.

[...]

PR #639, opened 2021-06-03
Setup: 'formats' for all user-facing extras

clbarnes:
This was initially going to be a prelude to adding more formats which require their own dependencies (`scipy` and `hdf5` for .mat files, `hdf5` for hdf5 files, `zarr` for zarr and N5 arrays), but from #574 it looks like the plugin system is being overhauled so I may hold off on those.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,14 @@ Optional Python packages:
     <li><a href='https://codeberg.org/monilophyta/imageio-flif'>imageio-flif</a> (for working with <a href='https://github.com/FLIF-hub/FLIF'>FLIF</a> image files)</li>
 </ul>
 
+These extras can easily be installed along with imageio:
+
+```bash
+pip install 'imageio[ffmpeg,itk,fits,gdal]'
+# or for all of the above
+pip install 'imageio[formats]'
+```
+
 Still on an earlier version of Python? Imageio version 2.6.x supports Python 2.7 and 3.4.
 
 
@@ -125,7 +133,9 @@ for maximum test coverage (100% for the core, >95% for each plugin).
 <p>Install imageio in edit mode, with dev tools:</p>
 
 ```bash
-pip install -e .[dev,docs]
+pip install -e '.[dev,docs]'
+# or, to additionally include all user-facing extras too
+pip install -e '.[full]'
 ```
 
 <p>Most developer command are done via <code>invoke</code>.</p>
--- a/setup.py
+++ b/setup.py
@@ -172,17 +172,23 @@ class build_with_images(sdist):
 install_requires = ["numpy", "pillow"]
 
 extras_require = {
-    "linting": ["black", "flake8"],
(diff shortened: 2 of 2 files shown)


FirefoxMetzger:
Thanks for this @clbarnes . I like the way you restructured the `setup.py` giving the `extra_requires`  a dedicated section for backends. The PR sounds related to https://github.com/imageio/imageio/issues/633 , I wonder if you are aware of it, and if you think this goes in the same direction.

Note: We used to call backends formats, but backends draws a more clear distinction between what is a way to represent images (format), what is an addon to imageio (plugin), and what is an external thing that drives a plugin (backend)).

clbarnes:
> The PR sounds related to #633 , I wonder if you are aware of it, and if you think this goes in the same direction

I think they are in the same direction, but can happily coexist. Extras are flexible enough that we could even group the extras (potentially with some namespacing); some could refer directly to plugins, then the use case extras could sit on top and refer to those plugin extras.

> We used to call backends formats

Sure, I'll change the name of that extra! I also considered an "everything-but-gdal" extra because I can never install the damn thing :roll_eyes:

FirefoxMetzger:
> Although then the "fits" -> "astropy" mapping is a little confusing.

That's true. Would it be cleaner to name the `extra_requires` based on the plugins? In this case the `fits` plugin uses `astropy` as a backend and provides support for the - equally named - `fits` format. Does this make sense?

> because I can never install the damn thing

This is another aspect I am thinking about regarding a "all"/"formats"/"plugins"/"backends" target in the extra requirements. Some people, e.g., those using pypy, may not be able to use all backends that are compatible with imageio due to platform constraints. On the other hand `pip install imageio[formats]` would suggest that this just works, or is at least smart enough to skip backends that don't work on the user's architecture. The latter being a bit unintuitive, because "I used `[formats]`, why do I not have backend X?". I'm not sure what a good solution is in this case.

clbarnes:
> Does this make sense?

Sounds good to me!

> I'm not sure what a good solution is in this case.

A few libraries (trimesh springs to mind), have an "easy" extra which installs all of the packages it expects to be easily installable; maybe that's the way. Users can always specify multiple individual extras to get around it so long as they're well documented. People using non-CPython distributions are likely to know enough about the limitations of those distributions to be able to work around but we could always drop a line into the readme about it. I think you're right that dynamically detecting which libraries are likely to be easy to install would be surprising behaviour, not to mention a headache to maintain, especially as more compiled libraries are (should be) transitioning to PEP518 build systems.

FirefoxMetzger:
@clbarnes #574 is merged :) Are you still interested in adding support for zarr, `,mat` etc.?

clbarnes:
Awesome! I'll raise issues and self assign them for when I get round to it. I'll close this for the moment too as quite a lot will have changed under the hood.

Issue #641, opened 2021-06-15
Smart searching of plugins by accounting for the file's extension.


FirefoxMetzger:
To speed up plugin selection, it would be good to check the extension of a file and - based on this extension - try plugins that are known to support this extension first. This should - of course - not exclude plugins that are known to not support the given extension, because the file may be labeled wrongly; instead, we just want to try those last.

If this sounds like an interesting feature to work on, please feel free to ask for more details :)

PR #648, opened 2021-06-16
Modernizing the docs

### LINKED ISSUE #621: (generated) API docs
One thing that I didn't find for the current API is documentation of the actual functions in the library. Are we generating it? If not, I think we should, since we already have some pretty reasonable docstrings for our functions.

### LINKED ISSUE #640: Document the V3 API
We need documentation beyond functional docs (we have those) that showcase the new API. This is for both the user-side and the plugin-writing side. Likely, it will also be useful to include some more high-level information on how the API is structured.

I recently upgraded the docs for one of my libraries to the pydata them and added some more sophisticated auto-generation for the API reference. I'll check if this is feasible here, too, so that our docs can look a little nicer :)

FirefoxMetzger:
Closes #621 

~I am going to write the docs for the new API and some use-case examples.~

I wanted to write the docs, but now I am thinking it is easier to clean up and modernize the docs a bit in this PR and then make a dedicated one that adds new content.

We didn't discuss it beforehand @almarklein , but what do you think about switching to the pydata theme for our docs? I like its clean and modern appearance.

This PR adds the following to the docs:

- [x] new theme (RTD -> pydata)
- [x] reorganize top menu following a more standard layout: install/examples/API reference. This is based on comparison with scipy, pydata, skimage, numpy and seaborn.
- [x] reorganized the file structure in the docs to reflect top level menus (there were too many files in the same folder to stay organized)
- [x] split dev install and normal install (mainly: move all dev stuff into "development")
- [x] add an overview page to Getting Started (this is the terminology comment we had in #574 )
- [x] explicitly reference drop27 in the toctree of Getting Started (address the warning we had before)
- [x] split the API docs into core (imread/mimread/volread) and backends (one page per backend for backend specific docs; mainly applicable kwargs)
- [x] add three examples to the example list for iterating files in a folder, iterating files in a .zip and reading single files from a zip.
- [x] add a Requests page to the docs (describe which kind of URIs are supported
- [x] link the backend reference pages via autosummary
- [x] specify the ffmpeg version needed
- [x] depreciate parts of our custom plugin in flavor of using autosummary for generation

Once all the boxes above are ticked, this PR is ready for review.


FirefoxMetzger:
Not entirely sure how I feel about the edit button ...

FirefoxMetzger:
@almarklein Was there any reason for disabling autosummary in favor of a homebrew version of it?

almarklein:
> what do you think about switching to the pydata theme for our docs? I

Sounds good!

> Was there any reason for disabling autosummary in favor of a homebrew version of it?

Can't remember. Let's run with autosummary. Worst case we find out why :)

FirefoxMetzger:
@almarklein If you have some spare time, could you give me feedback on the following two doc pages?

https://imageio--648.org.readthedocs.build/en/648/getting_started/overview.html
https://imageio--648.org.readthedocs.build/en/648/getting_started/requests.html

docs/getting_started/overview.rst
```diff
@@ -0,0 +1,76 @@
...
+    A blob of data that contains image data. Typically, this is a file on your
+    drive that is read by ImageIO. However, other sources such as HTTP or file
+    objects are possible, too. 
+
+
+Issues
+------
+
+In this repo, we maintain ImageIO Core as well as all the plugins. If you find a
+bug or have a plroblem with either the core or any plugin, please open a new
```
almarklein, line 51: typo, plus small tweak
```suggestion
bug or have a problem with either the core or any of the plugins, please open a new
```

docs/getting_started/overview.rst
```diff
@@ -0,0 +1,76 @@
...
+    all plugins if you didn't specify one explicitly.
+
+Plugin
+    A backend-facing adapter/wrapper that responds to a request from
+    iio.core. It can convert a request that comes from iio.core into a sequence of
+    instructions for the backend that fullfill the request (eg., read/write/iter). A
+    plugin is also aware if its backend is or isn't installed and handles this case
+    appropriately, e.g., it deactivates itself if the backend isn't present.
+
+Backend
```
almarklein, line 35: Idea: maybe we should call this "backend library" to make the distinction with plugin more clear.

docs/development/index.rst
```diff
@@ -0,0 +1,21 @@
...
+  Developer API <devapi>
+  Writing plugins <plugins>
+
+
+Developer Installation
+----------------------
+
+For developers, we provide a simple mechanism to allow importing
+imageio from the cloned repository <https://github.com/imageio/imageio.git>.
+See the file ``imageio.proxy.py`` for details.
```
almarklein, line 17: ... I think we should advocate using `pip install -e .` instead. That proxy thing is from before pip existed.

docs/getting_started/requests.rst
```diff
@@ -0,0 +1,157 @@
...
+    img = iio.imread("file://path/to/my/image.jpg")
+    img = iio.imread("file:///path/to/my/image.jpg")
+    img = iio.imread("file://C:\\path\\to\\my\\image.jpg")
+
+
+Byte Streams
+------------
+
+ImageIO can directly handle (binary) file objects and BytesIO objects. This includes
+subclasses of the above as well as generic objects that implement ``close`` and a ``read`` and/or ``write`` function.
```
almarklein, line 45: I got a bit confused by the "and" in "file objects and BytesIO objects", because a BytedIO object is a binary file object, right? So maybe something like:
```suggestion
ImageIO can directly handle (binary) file objects. This includes ``BytesIO`` objects (and subclasses thereof)
as well as generic objects that implement ``close`` and a ``read`` and/or ``write`` function.
```

docs/reference/index.rst
```diff
@@ -0,0 +1,44 @@
...
+
+
+Plugins & Backends (Advanced Usage)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Sometimes, you need to do more than just "load an image, done". Sometimes you
+need to use a very specific feature of a very specific backend. ImageIOs allows
+you to do so by allowing its plugins to extend the core API. Typically this is
+done in the form of additional keyword arguments (``kwarg``). Below you can find
+a list of available plugins and which arguments they support.
```
almarklein, line 25: I think the supported formats page already fulfills this purpose to some extend. I don't seem to be able to get to the plugin-specific docs from the docs here. Also, plugins are classes, and the modules documented here seen to contain functions and classes aimed for internal use.

docs/reference/index.rst
```diff
@@ -0,0 +1,44 @@
...
+
+Core API (Basic Usage)
+^^^^^^^^^^^^^^^^^^^^^^
+
+The API documented in this section is shared by all plugins
+
+.. toctree::
+    :maxdepth: 1
+
+    Core API v2.9 <userapi>
```
almarklein, line 15: It looks like the new API is not documented yet, correct?

almarklein (commented):
This is again some massive work, nice!

docs/development/index.rst
FirefoxMetzger: Oh wow, I knew imageIO is rather old, but I never realized it comes from the pre-pip era. I agree that this should be updated.

docs/getting_started/overview.rst
FirefoxMetzger: I like your thinking. Is it always a library? E.g., thinking of FFMPEG, we actually call an executable, but I guess this is wrapped into a library (imageio-ffmpeg / pyav / ...) and further this distinction may not be too important outside of plugin development.

That said, I'm happy to call it backend libraries or backend <something>.

docs/reference/index.rst
FirefoxMetzger: Correct. I wanted to use this PR to document the new API, but then realized that I made a lot of commits to refactor the docs to "get things read for documentation". Because of that I decided to make this PR about refactoring things, and the next PR about actually writing docs for it and updating the examples, etc.

This way we can have these docs for v2.9 and then have updated docs and examples for v2.10/v3.0.

I also decided to clean up some of the doc issues in the process, since I am shuffling around a lot of content and adding a paragraph or note here and there is easy to do.

docs/development/index.rst
almarklein: Not 100% sure if imageio is, but that proxy file is ;)

docs/getting_started/overview.rst
almarklein: > Is it always a library?

I guess it usually is. For the backends included in imageio it's technically a sub-package, but  ... close enough? I sometimes find myself slightly confused between `backend` and `plugin`, and I imagine it will be worse for people other than us. So anything to help reduce that confusion would help :)

docs/reference/index.rst
FirefoxMetzger: Yes. The plugins & backends are still work in progress :) I haven't had time to finish this part yet, since I was working on a research paper deadline and was a bit pressed for time, but I am planning on working more on this during this week.

---

My idea was to refactor the supported formats page since it currently fulfills a double purpose / has a bit of a split personality. It is a big list of all the format strings that exist and, once you know the correct string, you can navigate to the backend-specific docs to tell you the kwargs. 

The title hints at "here are all the formats you can use", which it does list, but if my desire is to "figure out which plugin I should use to read my _image.XYZ_ file" then I would indeed check the "Supported Formats", but the current list is a bit confusing. I would have expected to find a list or table that I can search (ctrl+F) for ".XYZ" and then it lists all the format strings (or backends) that I could use here.

Similarly, if I know that I want/need to use backend X (e.g. imageio-ffmpeg) then I need to _know_ that I need to search for this information in "Supported Formats" (and click on the correct format string in the list). Since this is something API-related, I would, however, have searched for this in the API reference.

Long story short, I would like to change "Supported Formats" into lists of the form

```
- <extension> [Format Full Name](link to description/wikipedia/standard)
    Backend Libraries: <list of format strings or backends>

- PNG [Portable Network Graphic](https://en.wikipedia.org/wiki/Portable_Network_Graphics).
    Format String: :ref:`PNG-PIL <imageio.plugins.pillow>`, :ref:`PNG-FI <imageio.plugins.freeimage>`, ...
```

One list for _backends by format_ and one for _formats by backend_, or as a table (if I manage to figure out how to make a nice table in sphinx) so that we can maintain a single structure but index it by row or column for either of the two scenarios.

I would like to have this section here (Plugins & Backends) become the backend-specific docs, i.e., kwargs, recommended/supported versions, intricacies/gotchas, ...

Then again, I might be overthinking this and it is cleaner and easier to leave it as is?

docs/development/index.rst
FirefoxMetzger: Done :)

docs/getting_started/overview.rst
FirefoxMetzger: Yep, I agree. I don't see anything right now that breaks the abstraction of calling it a "library", so let's roll with it and see how it goes.

FirefoxMetzger: done :)

FirefoxMetzger:
@almarklein The API docs for the plugins/backend libraries should render correctly now. Let me know if I accidentally dropped any function in the process :D

docs/reference/index.rst
almarklein: I think you're right. People can come at this from two directions and we should probably have two doc pages for that.

One note is that certain plugin-modules provide multiple plugin classes (e.g. FreeImage and Pil have a few specific classes for png and jpg iirc. Not sure if we'd want to list modules and then when you click one you get all the plugins defined in that module, or list all plugin classes to begin with (might be a bit much).

FirefoxMetzger: @almarklein I noticed this while I was moving docstrings around. At the moment I have merged them all and indicated for which format a specific argument is available. Except for pillow, this seems to work with how things were documented. For pillow I'm not yet sure if I want to leave it as is, or if I want to refactor further.

In either case, this is something I will worry about after I refactored the supported formats. (we can still discuss the best approach though)

FirefoxMetzger:
A little bit larger than originally anticipated, but I think it is worth cleaning stuff up a little ... and most parts are just copy/pasting docstrings around to align the doc layout a bit more with other familiar python libraries (scipy, numpy, pandas, ...)

docs/supported_formats.rst
```diff
@@ -0,0 +1,586 @@
...
+Below you can find a (non-exhaustive) list of formats that ImageIO supports
+together with the backend supporting it. This list is non-exhaustive, because it
+only lists the formats we were aware of at the time the docs were written. If
+any backend introduces support for new formats they, too, will be supported but
+our docs may take time to update and reflect this. If in doubt, check the docs
+of the respective backend.
+
+Format List
+-----------
+Each backend listed here is listed using the format string you would use to call
```
almarklein, line 18: Is this file still auto-generated? I see code-gen code being removed, but no new code to replace it. I hope it's a manual document, because I suspect that will be painful to maintain.

FirefoxMetzger: You hope it's a manual document or you hope its _not_ a manual document? But to answer your question either way: At the moment it is a manual document.

Moving forward, I think it will become a little easier, since a lot of the formats are `XXX-FI` or `XXX-PIL`, and in the new API all of these will become `plugin="freeimage"` or `plugin="pillow"`, at which point we could think of removing the Format List (by extension) and just keep a format list by backend and ask people to Ctrl+F the extension they are looking for. Then again it might be nice to have a long list of extensions/formats, because it looks impressive.

FirefoxMetzger: Oh, and some of the codegen is no covered by autosummary/autodoc. I also removed all the template strings from the extension and placed them into dedicated template files.

almarklein: Yeah I meant *not* a manual document, sorry.

> [...] in the new API all of these will become `plugin="freeimage"` or `plugin="pillow"`.

mmm what would that look like? I imagine that many formats still have their own specific keyword argument (e.g. compression for png, exif rotation for jpg, etc, etc.).

almarklein: In either case, this list seems like it's easy to generate from the information that is already available. It being manual will mean having to keep things in sync, and discrepancies will creep in at some point.

FirefoxMetzger: > mmm what would that look like? I imagine that many formats still have their own specific keyword argument

@almarklein Sure, you can just append those to the call and they will get sorted out by the plugin or the backend, e.g., `iio.imread("some.jpg", plugin="pillow", rotate=False)`. If there are a lot of backend-specific ones like there are for ffmpeg a plugin can specify `**kwargs` in its implementation of the interface and then forward them appropriately or throw an Exception for unsupported ones.

If you were/are wondering how it will look like on the documentation side then that will depend on the individual plugin/backend. Each plugin has its module parsed by autosummary/autodoc and then it can decide how it wants to be structured. At the moment all plugins are single page and follow numpydoc style where possible, but it doesn't have to be like this. You could have a `Format Specific Parameters` section where you list the parameters that are unique to a format if there are many, or you can simply list them in a single `Parameters` list and flag them appropriately (what I've been doing so far since it's only a few parameters if any). 

The main point is that if you use a plugin, say `plugin=dicom`, then the DICOM plugin API reference is _the_ place to search for any parameters; format-specific or otherwise. No more need to start digging through Supported Formats (this info doesn't really belong there anyway). Supported Formats will instead list what format is read by which plugin; if you wonder "(Can imageio read)/(Does imageio support) X " (for X = RTSP streams :D https://github.com/imageio/imageio/issues/658) then you go to Supported Formats, Crtl + F for the extension or format name, and check which plugin you need to install.

If you _need_ to read a specific file with a specific plugin using as a specific format and the format is - for whatever reason - different from the format that the backend detects (via extension magic number or otherwise), then you need to check the plugin docs regarding the keyword that forces this particular backend to interpret the specific file as a specific format. This does seem rather niche, so I think it is okay that the call is a bit more sophisticated (`iio.imread(specific_file, plugin="specific_plugin", kwarg_enforcing_format=...)`.

---

> this list seems like it's easy to generate from the information that is already available

Yes, partially. In the legacy plugin manager a plugin must explicitly register formats, which I guess we could tap into to populate the list. The new plugin manager currently doesn't care about formats (by design) and sees them as more of an optional optimization (https://github.com/imageio/imageio/issues/641) than mandatory information. We will need to write some kind of dict/json file/thing for https://github.com/imageio/imageio/issues/641 that signals which plugins to prioritize based on file extension. We could again use this to populate the list.

What we can't easily automate is the association of an extension with its full format name (and link to additional documentation).  I think this information is important because it clarifies what an extension is and it (may) de-alias formats that share the same extensions. The only way I can think of maintaining this is to template the list and then simply populate pre-existing entries based on the extension-plugin pairs we get from somewhere. We will still have to manually add new formats and we won't gain guarantees on completeness (since backends might introduce new formats without us being aware for a while). The library will support those new formats immediately without changes, but the docs will lag behind. Considering this, automation didn't offer enough bang for the buck imo, so I just wrote a manual one.

That said, I'm happy to automate if you have a better idea.

almarklein: > If you were/are wondering how it will look like on the documentation side then that will depend on the individual plugin/backend. 

Yes, I was wondering mostly about the docs. Ok, I think I see what you mean. So the Pillow plugin will be a relatively large page describing the parameters for generic formats, and a few sections for png, jpg, etc.? I hadn't thought about that, but it sounds good. Would be nice, though, if the links in the list of supported formats would bring the user to the relevant section (Eg for png, it lists at least pillow, freeimage, and if you click on pillow you get to the section on the pillow page describing the png format params). I guesst this could be as simple as adding the format at the end of the link ("/plugins/pillow.html#png"), and making sure that the doc page has `<a name='png' />`.

----

> What we can't easily automate is the association of an extension with its full format name (and link to additional documentation). 

So perhaps we should add some information to the plugin classes so that we can? ;)  Maintaining some meta-data at the plugins is easier than maintaining a large manual list.

FirefoxMetzger: > Would be nice, though, if the links in the list of supported formats would bring the user to the relevant section

Nice idea! I'm sold :D This might be a tad iffy to maintain if we automatically generate the list - only a few plugins will have the need for this -, but we can see.

> So perhaps we should add some information to the plugin classes so that we can? ;) 

What kind of info did you have in mind and where should it live? In my mind that would be a description of each format supported by a plugin/backend. This would mean a lot of replication if multiple formats can read the same format (tiff, jpg, png, ...) that needs to be kept in sync. I guess I am misunderstanding what you have in mind.

almarklein: Haha, ok, I think I see what you mean. You have a format, say PNG, and with it a short description and a link to the wiki page about that format, and then a link to each plugin that supports that format. So maybe we should create a definition of formats, if only for the sake of documentation. Something like:

```
format_meta_data = {
     "png": {
        "description": """ PNG is a lossless image format, blabla
            ....
        """,
       "plugins": ["pillow", "freeimage"], 
}
```
or
```
class PngFormat:
   """ PNG is a lossless bla bla
   ..
   """
   plugins = "pillow", "freeimage" 
```

And plugins that specify what formats they support based on the key ("png" in this case).

Maybe we could also specify common extension here, instead of on the plugins. mmmm

FirefoxMetzger: > So maybe we should create a definition of formats, if only for the sake of documentation. [...] And plugins that specify what formats they support based on the key ("png" in this case).

Just to poke the bear: If we only do it for the docs, why not use a `_templates/format.rst` file instead of a meta dict and have it populated by our extension or autodoc? This would give us much easier control over the formatting and allow IDEs with rst support to perform syntax and cross-ref checks. We can also place the description into it directly since it won't be used by any other part of our codebase.

> Maybe we could also specify common extension here, instead of on the plugins.

How do you mean? I thought the (dict keys)/(class names) would be based on extensions? Or are you thinking of basing them on a format's full name and then add extensions as a metadata key?

almarklein: > why not use a _templates/format.rst file

I have never worked with rst templates, so yeah, if that works, great!

> How do you mean? I thought the (dict keys)/(class names) would be based on extensions? Or are you thinking of basing them on a format's full name and then add extensions as a metadata key?

So right now (in the old system) a plugin class has an `extensions` attribute that is used to pre-filter possible plugins. We mean to bring something like this back. But instead of specifying extensions, we could specify format names, and declare what extensions are common to this format in a different (central) place. If we'd go with the rst solution, thus would make much less sense though, I guess :)

FirefoxMetzger:
@almarklein Okay, I didn't get a chance to work on ImageIO for a while now, but I made some time today to add the auto-generation we discussed last month.

I changed the `imageio_ext` to use sphinx's internal templating engine (jinja2) and rewrote the generation routines we had taking advantage of it.

I also added a new module to imageio called `config`, which stores a list of all extensions (as python classes as discussed), and a dict of all formats/plugins. Both are used to generate the list of known formats in the docs. Moving forward, I want to use the list of extensions to implement https://github.com/imageio/imageio/issues/641

I'm thinking if we should/could use the list of plugins to configure the plugins that imageio is aware of. We could extend this to on-demand/lazy loading of plugins the first time it would be used (instead of loading all installed ones on `import imageio`). I've done something similar recently for some XML bindings [here](https://github.com/FirefoxMetzger/scikit-bot/blob/2abe509b3a555048717d29c2408b407d7e8b8315/skbot/ignition/sdformat/sdformat.py#L149-L151).

almarklein:
Looking good! Is this ready for merge?

FirefoxMetzger:
@almarklein Yep. all green from my side. If you don't have anything else to add I'm happy to merge the PR.

Issue #562, opened 2020-10-28
A simpler, more scalable way to write plugins


almarklein:
@jni mentioned [this post](https://blog.danallan.com/posts/2020-03-07-reader-protocol/), and that got me thinking ...

This would be quite a bit change, and some things (*cough* meta data) are more urgent. However, it might be that the required changes in metadata break compatibility in some ways, making it easier to roll these changes out together?

I'd like to consider a new way to specify plugins. These plugins will not be required to inherit from `imageio.core.Format`, nor will they need to be registered at runtime. This means that libs can specify imageio formats without needing to depend on imageio. If we use entriepoints to let libs advertise their "imageio formats", then imageio can detect the format even if the lib is not imported (though it must be installed).

When a user reads a file, imageio can query possible formats and auto-import these formats as needed. That way, imageio stays light, while the number of supported formats can be increased in a very scalable way (in terms of memory, but also maintenance).

Further, we can keep a dict of common formats specified in 3d party libraries, so that when a user tries to read a file type for which no format is available on the system, imageio can suggests the installation of a library.

Imageio can then spin off some of its plugins into separate libraries, and mark them as dependencies in `setup.py`. These formats are then still always available, but are only loaded when needed.

Note, that with a little bit of work, we can also delay the import of plugins as things currently are ...

Preferably, this format should be generic, so it can also be consumed by other tools. Not sure if this is feasible though.

Info that we need (based on the current format system):
* Name and docs
* Common extensions
* Supported image modes (2d, 3d, multi-image)
* A way to tell whether the plugin can read the data, based on the first n bytes
* a way to get meta data
* a way to read image data (and meta data)
* a way to write image and meta data

Example:

```py
class FooFormat:
    """
    Docs
    """

    extensions = ".foo", ".bar"
    modes = "ri", "rI", "wi"  # read images and sequences, write single images
    modes_alt = "r2", "r2m", "w2"  # maybe use numbers for dimension, and 'm' for multi?

    def __init__(self, request, mode=''):
        # Imageio will pass a request object here, via which we can
        # obtain a file object, or a local filename, whatever we need.
        # To make this more generic, we should also support plain file
        # objects.
        if not can_handle(request.firstbytes, mode):
            raise NotImplementedError("This format can not read/write this data")

    def get_meta(self, index=None):
        # Return the metadata of the file or the specified frame.
        # Formats are expected to provide metadata with standardized
        # fields.
        pass

    def read(self, index=None):
[...]

PR #654, opened 2021-06-30
factor out PillowPlugin.open()

almarklein:
See #562 for the discussion. At first I did not realize that `open()` was in fact a "private" method only used by the class itself. Nevertheless, with request doing most of the handling, I think the pattern of opening a file when the plugin object is created makes more sense.

### DIFF
--- a/imageio/plugins/pillow.py
+++ b/imageio/plugins/pillow.py
@@ -119,11 +119,10 @@ class PillowPlugin(object):
                 ) from None
 
         self._request = request
-        self._image = None
-
-    def open(self):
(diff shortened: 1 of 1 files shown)


FirefoxMetzger (approved):
Looking good from my side. 

The PR coverage failure comes from us reducing the total number of covered lines in the project while keeping the number of uncovered lines constant. This is a false-positive and we could think of decreasing the required coverage to 99.95% or 99.9% to avoid this. Either way, not an issue here.

LGTM, unless you are thinking of adding more to this PR.

Issue #553, opened 2020-09-22
Shift in output array when reading large spe files (Princeton Instrument : Lightfield)


gh4ag:
I had written a wrapper around the get_reader module to read Princeton Instrument .spe files:

f```
from imageio import get_reader
from imageio.core.util import asarray
import logging
import re
import xmltodict

# Capture the warning message sent by get_reader (if any)
log_capture_string = io.StringIO()
ch = logging.StreamHandler(log_capture_string)
ch.setLevel(logging.WARNING)
module_logger = logging.getLogger("imageio.plugins.spe")
module_logger.addHandler(ch)

# call to get_reader
reader = get_reader(f)

# extraction of the warning message
log_contents = log_capture_string.getvalue()
log_capture_string.close()
module_logger.removeHandler(ch)

# in case an error log pops up, we update the value of ._len to make it match the actual number of frames
if "frame" in log_contents:
    p = "(\d+)(?: frame)"
    res = re.findall(p, log_contents)
    if len(res) > 0:
        reader._len = int(res[1])

im = [frame for frame in reader]

if "__xml" in reader._meta:
    xml = xmltodict.parse(reader._meta["__xml"])

return [asarray(im), xml]
```
Then I observed recently that the output file, when the .spe file was large (2 Go), I was getting an output where the image was circularly shifted (i.e. the columns were pushed to the left and placed and inserted at the beginning of the frame).

Let's consider 4 columns, tagged 1 to 4, instead of having 1234, I am getting 4321, etc ...

Any idea why?

almarklein:
> Let's consider 4 columns, tagged 1 to 4, instead of having 1234, I am getting 4321, etc ...

That looks like it has the frames reversed then, right? 

The imageio plugin for SPE is here: https://github.com/imageio/imageio/blob/master/imageio/plugins/spe.py

I looked at the file, and cannot find anything suspicious. I also don't see how it could behave differently for larger data files, so this behavior is surprising indeed.

I see you also hook into the logger. Are you sure that the observed behavior is not due to the logging?

cc @lschr as the original author of the SPE plugin.

gh4ag:
Sorry, I have made a small mistake in my explanation.

The image was not reversed, I was just shifted

1234  -> 4123

I will set a complete MWE with a set of data for reproduction

Le lun. 28 sept. 2020 à 11:50, Almar Klein <notifications@github.com> a
écrit :

anntzer:
[...]

PR #662, opened 2021-09-11
Get SPE frame stride from XML metadata.

anntzer:
Closes https://github.com/imageio/imageio/issues/553 (see discussion there).  Unfortunately I don't have a file I can share for tests, but perhaps @gh4ag can help there?

### DIFF
--- a/imageio/plugins/spe.py
+++ b/imageio/plugins/spe.py
@@ -199,6 +199,7 @@ from datetime import datetime
 import logging
 import os
 from typing import Any, Callable, Dict, Mapping, Optional, Sequence, Union
+from xml.etree import ElementTree as ET
 
 import numpy as np
 
@@ -707,26 +708,32 @@ class SpeFormat(Format):
             if index >= self._len:
                 raise IndexError("Image index %i > %i" % (index, self._len))
 
+            frame_size = self._shape[0] * self._shape[1]
+            stride = frame_size * self._dtype.itemsize  # Default stride.
+            meta = self._get_meta_data(index)
+            if "__xml" in meta:
+                et = ET.fromstring(meta["xml"])
+                strides = {
+                    int(e.attrib["stride"])
+                    for e in et.findall(".//{*}DataBlock[@type='Frame'][@stride]")
+                }
+                if len(strides) == 1:
+                    (stride,) = strides
+                elif len(strides) > 1:
+                    raise ValueError("Unsupported irregular DataBlock spacing")
+            dtype = [
+                ("data", self._dtype, self._shape[:2]),
+                ("pad", "u1", stride - frame_size * self._dtype.itemsize),
+            ]
+            self._file.seek(Spec.data_start + index * stride)
+
             if self.request.mode[1] in "vV":
                 if index != 0:
                     raise IndexError("Index has to be 0 in v and V modes")
-                self._file.seek(Spec.data_start)
-                data = np.fromfile(
-                    self._file,
-                    dtype=self._dtype,
-                    count=self._shape[0] * self._shape[1] * self._len,
-                )
-                data = data.reshape((self._len,) + self._shape)
+                read = np.fromfile(self._file, dtype, count=self._len)
             else:
-                self._file.seek(
-                    Spec.data_start
-                    + index * self._shape[0] * self._shape[1] * self._dtype.itemsize
-                )
-                data = np.fromfile(
-                    self._file, dtype=self._dtype, count=self._shape[0] * self._shape[1]
-                )
-                data = data.reshape(self._shape)
-            return data, self._get_meta_data(index)
+                read = np.fromfile(self._file, dtype, count=1).squeeze(0)
+            return read["data"], meta
 
 
 def roi_array_to_dict(a):


imageio/plugins/spe.py
```diff
@@ -707,26 +708,32 @@ def _get_data(self, index):
             if index >= self._len:
                 raise IndexError("Image index %i > %i" % (index, self._len))
 
+            frame_size = self._shape[0] * self._shape[1]
+            meta = self._get_meta_data(index)
+            try:
+                et = ET.fromstring(meta.get("__xml", ""))
+                (stride,) = {
+                    int(e.attrib["stride"])
+                    for e in et.findall(".//{*}DataBlock[@type='Frame'][@stride]")
```
FirefoxMetzger, line 717: Regarding the x-path:

1. What is the function of the `{*}` here? If I understand correctly, we are actually interested in `DataBlock` elements, so why do we need a wildcard here?
2. Reading the SPE spec (I'm not familiar with intrinsics) I found that `DataBlock`s of type `Frame` always have a `stride` element. Why would we explicitly filter for it here? If a `@type="Frame"` block doesn't have a stride it is malformatted, and (similar to my point below) we would like to `raise` this instead of silently ignoring it.

imageio/plugins/spe.py
```diff
@@ -707,26 +708,32 @@ def _get_data(self, index):
...
+                }
```
FirefoxMetzger, line 718: We are creating a set here, which we immediately unpack; any reason why we want to combine stride values that occur multiple times?

If I understand correctly, we expect exactly one DataBlock here (frame-level metadata), if my understanding is correct, wouldn't a simple `et.find(...)` be cleaner, since it avoids any unwrapping/filtering/etc.?

imageio/plugins/spe.py
```diff
@@ -707,26 +708,32 @@ def _get_data(self, index):
             if index >= self._len:
                 raise IndexError("Image index %i > %i" % (index, self._len))
 
+            frame_size = self._shape[0] * self._shape[1]
```
FirefoxMetzger, line 711: This will produce unexpected/undefined behavior in SPE 3.0 if the file contains differently sized ROIs. (This appears to be one of the shiny new features of SPE 3.0 if I understand correctly.)

Do you have interest/plans for adding support for this, too?

FirefoxMetzger (changes requested):
Thanks, @anntzer for getting the ball rolling on this! It looks quite good already, and I left a few comments for things we might be able to improve/clarify.

One thing that is currently missing from the PR is a  test (as you have pointed out already); once we have that, I think it is only a short way until we can get it merged 🚀

FirefoxMetzger:
I'm also re-running the tests ... one of the pillow tests failed on pypy, but this PR shouldn't affect them (and tests are known to be flakey on pypy + macOS). Let's see what happens.

imageio/plugins/spe.py
```diff
@@ -707,26 +708,32 @@ def _get_data(self, index):
...
-                self._file.seek(
-                    Spec.data_start
-                    + index * self._shape[0] * self._shape[1] * self._dtype.itemsize
-                )
-                data = np.fromfile(
-                    self._file, dtype=self._dtype, count=self._shape[0] * self._shape[1]
-                )
-                data = data.reshape(self._shape)
-            return data, self._get_meta_data(index)
+                read = np.fromfile(self._file, dtype, count=1).squeeze(0)
```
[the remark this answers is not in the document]
anntzer: > Use count=self._get_length() in both np.fromfile calls to handle the kwarg

Actually it's the opposite, in "vV" mode _get_length() returns 1 but you want to read `self._len` entries whereas in "iI" mode _get_length() returns `self._len` but you want to read just 1 entry.  (I guess you could use constructs like `self._len // self._get_length()` but that seems pretty unreadable.)

For everything else, it comes down to whether you want to duplicate the call to `np.fromfile` or the check on `self.request.mode[1]`, so I'm not sure there's much to be gained.  The best I can think of would perhaps be `reader = functools.partial(np.fromfile, self._file, dtype)` but even that just seems like confusing things more than anything else.

But feel free to push a better version if you can think of something.

imageio/plugins/spe.py
```diff
@@ -707,26 +708,32 @@ def _get_data(self, index):
...
+            meta = self._get_meta_data(index)
+            try:
+                et = ET.fromstring(meta.get("__xml", ""))
+                (stride,) = {
+                    int(e.attrib["stride"])
+                    for e in et.findall(".//{*}DataBlock[@type='Frame'][@stride]")
+                }
+            except (
+                ET.ParseError,  # No XML or bad XML.
+                ValueError,  # Not numeric or multiple values.
```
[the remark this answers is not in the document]
anntzer: I don't have a super strong opinion here (I haven't actually seen malformed files in the wild), but the basic idea is that quite often the frames are just contiguous (that's why #553 wasn't discovered when the SPE reader went in, I guess), so the error is indeed recoverable by assuming a contiguous layout (I'd say this hardly qualifies as "magic").
But I can throw an exception instead if you prefer.

imageio/plugins/spe.py
anntzer: re xpath: the xml is actually namespaced, so in theory you should write something like ".//{http://www.princetoninstruments.com/spe/2009}DataBlock..." which is quite a mouthful for not much semantic information in practice; using a namespace wildcard allows skipping it.

re: stride: same argument as above for lenientness, but again I can change that, just pick either way.

imageio/plugins/spe.py
anntzer: At least in theory I guess(?) you can have multiple DataBlocks with different strides, which we simply don't support (I guess in this specific case we should indeed raise, because the contiguous-size assumption definitely doesn't work).

imageio/plugins/spe.py
anntzer: I don't plan to work on this, but (per above) I'll make it error out cleanly.

imageio/plugins/spe.py
FirefoxMetzger: re xpath: I see, thanks for explaining

re stride: Yes, I'd prefer a strict implementation. We can always choose to support functionality that goes beyond the  spec itself (like failure recovery in easy cases), but I'd be happier if this happens explicitly. Opt-in or Opt-out both work for me, so we could do something like a `strict=False` kwarg and do error recovery by default. As long as it is possible to fall back to a "_just_ the standard and no-frills" implementation I'm happy :)

imageio/plugins/spe.py
FirefoxMetzger: That would be much appreciated :)

imageio/plugins/spe.py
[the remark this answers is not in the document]
FirefoxMetzger: Good catch! I agree; in this case, it is indeed cleaner to stick with two `np.fromfile` calls. Let's leave it as is.

imageio/plugins/spe.py
anntzer: I just let the errors propagate out for now.

anntzer:
Closing for lack of progress and because I don't work with SPE files anymore.  Anyone should feel free to pick up this patch if it is useful for them.

Issue #644, opened 2021-06-15
Roadmap to ImageIO v3


FirefoxMetzger:
With the basics of the new API in place, there are a few more things that need doing and then we can release a new version of imageio.

Here is a (loose) checklist. Please feel free to make additions:

- [x] Set up a CD pipeline (I've been thinking a lot about [semantic versioning 2.0.0](https://semver.org/)) Maybe it is worthwhile here?
- [x] Fix issue #640 
- [x] Fix issue #643 
- [x] Fix issue #641
- [ ] Add depreciation warnings to v2 functions

FirefoxMetzger:
@almarklein What are your thoughts about this library: https://github.com/semantic-release/semantic-release

Could it be useful for us to set up a CD pipeline?

almarklein:
At first glance my feeling is that the tool you mentioned is overkill, and the fact that it tries to determine how the version should be bumped based on commit messages seems like it might also work against us.

I am not against some form of automation, but since imageio is pure Python, the release procedure is already quite easy, so I would warn against overdoing it and thereby locking us into a certain toolchain. Some ideas for inspiration:
* Run a CD build when a tag named `v...` is pushed, that makes a GH release, and a package that is pushed to pypi (like e.g. [here](https://github.com/pygfx/wgpu-py/blob/main/.github/workflows/ci.yml#L164)).
* Another idea, since imageio is pure Python, is to create a script that does the release locally: bump the version, git tag that commit, pushes that to github, builds a wheel and pushes that to pypi (like e.g. [here](https://github.com/almarklein/timetagger/blob/main/tasks.py#L129)).
* I know some packages (skimage?) auto-generate release notes based on closed issues and PR's. Very interesting, but the amount of activity (once this big refactoring is over) might not be high enough for this to be worth the effort?

almarklein:
All this said, if that tool (or another) is easy to set up and does not have too much implications elsewhere, feel free to give it a spin.

FirefoxMetzger:
> I would warn against overdoing it and thereby locking us into a certain toolchain

I have the same concerns; I'd like to remain lightweight and flexible. It could be that I am keen on some kind of framework, because I have much more experience with Dev than with Ops, so whenever I get to delegate Ops responsibility I'm happy. Perhaps that skews my view.

[...]

PR #667, opened 2021-10-18
FEAT: Add CD pipeline

FirefoxMetzger:
Closes: #644 

This PR adds a CD pipeline that does the following:

- check nightly for any new commits to master (if none, exit)
- check if they require a new semver version (if no, exit)
- run through full unit-tests (if they fail, exit)
- check if the docs build (if they fail, exit)
- check if the wheel can be built (if it fails, exit)
- bump ImageIO's version
- update the changelog with the changes made (breaking, feature, fix, doc, ...)
- commit, tag, and push the new version + changelog to master (this also pushes the new tag)
- build the new imageIO version
- release the new version to GitHub
- [x] (TODO, once I have a pypi token) release the new version to PyPi

Optional:
- [ ] TODO: remove "stable" docs from RTD (they will be at most 24h apart from "latest")

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,12 @@
 # IMAGEIO
 
-[![PyPI Version](https://img.shields.io/pypi/v/imageio.svg)](https://pypi.python.org/pypi/imageio/)
-[![Supported Python Versions](https://img.shields.io/pypi/pyversions/imageio.svg)](https://pypi.python.org/pypi/imageio/)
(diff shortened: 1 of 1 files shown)


FirefoxMetzger:
@almarklein I am pretty much ready here. Could you set up a PyPi secret for this repo? Alternatively, you can give me privileges to modify things over at pypi. That way, I will setup the secret.

Issue #643, opened 2021-06-15
How to best pass file ownership from imopen to a plugin?


FirefoxMetzger:
The basic idea behind the file-management of the new API is simple:

```python
with iio.imopen(...) as img_file:
    img = img_file.read(...)
    ...
```

The file is opened by `imopen`, held by `img_file` during the context, and closed once the context manager exits. This works nicely.

A problem occurs when we use `imopen` outside a context manager:

```python
instance = iio.imopen(...)
# ... program crashes or returns here
instance.read(...)
```

In this case, the open file may not be cleaned up properly or could cause a dangling reference. How can we smooth this out?

One option would be to have `imopen` close the file after it is done selecting a plugin and have the plugin reopen it. This might work, but is undesirable because it slows down the default case of using `imopen` inside a context manager (as intended)

Another option would be to use a promise-style setup where `imopen` returns a `PluginPromise` which gets turned into the correct plugin by its `__enter__` method. This would ensure that all file access is properly guarded. Promise-based interfaces are more common in JS than they are in python though, so I don't know if this idea is easily maintainable. 

A third option could be to make the destructor of `imopen` close the file. However, this may limit applications, because they cant use the file outside `imopen`'s scope.

All the options have pros and cons. The question here is which option is the best.

almarklein:
> #also, is the file closed here or left open?

This would typically close the file right there, because the object is never even stored in this scope.

You're right that this behavior is an implementation detail of the Python interpreter. But maybe this is ok? The same aplies to Python file objects; you have to use them with care.

I think we have three options:
1. Do open+close when instantiating and then again when reading. This feels like a waste.
2. Follow same approach as Python file objects. Yes, the users need to use it with care, but this *is* our somewhat lower-level API, and it helps that it's the same as Python's `open()`.
3. Provide keywords args to specify the behavior. This complicates the API, and the default behavior would still be (1).

I'm leaning towards (2) because I don't like the other options. I think we can get away with shifting responsibility to the user here, since we have the higher level API that does not have this issue, and is what 90% of users will use.

FirefoxMetzger:
@almarklein Okay then let's opt for a behavior that is similar to `open` in CPython and see if any pypy user runs into issues. I will add a warning to the `iio.imopen` docs to inform pypy users about this behavior.

almarklein:
> Edit: This will essentially make each plugin responsible for closing the file once it is done with it.

[...]

PR #673, opened 2021-11-01
MAINT: consistently handle file opening/closing

FirefoxMetzger:
Closes: #643 

This PR adds a warning (for pypy users) to `iio.imopen` to take care when using it outside a context manager. It also calls request.finish() inside `__del__` of `LegacyPlugin` and the `Pillow` plugin.

Finally, I did some small cleanup while looking over the code and adding above changes, adding type annotations and removing unneeded branching.

### DIFF
--- a/imageio/core/request.py
+++ b/imageio/core/request.py
@@ -494,11 +494,11 @@ class Request(object):
                     shutil.copyfileobj(self.get_file(), file)
             return self._filename_local
 
-    def finish(self):
-        """finish()
-        For internal use (called when the context of the reader/writer
-        exits). Finishes this request. Close open files and process
-        results.
+    def finish(self) -> None:
+        """ Wrap up this request.
+
+        Finishes any pending reads or writes, closes any open files and frees
+        any resources allocated by this request.
         """
 
         if self.mode.io_mode == IOMode.write:
@@ -506,8 +506,7 @@ class Request(object):
             # See if we "own" the data and must put it somewhere
             bytes = None
             if self._filename_local:
-                with open(self._filename_local, "rb") as file:
-                    bytes = file.read()
+                bytes = Path(self._filename_local).read_bytes()
             elif self._file_is_local:
                 bytes = self._file.getvalue()
 
@@ -534,10 +533,7 @@ class Request(object):
 
         # Remove temp file
         if self._filename_local:
-            try:
-                os.remove(self._filename_local)
-            except Exception:  # pragma: no cover
-                pass
+            os.remove(self._filename_local)
             self._filename_local = None
 
         # Detach so gc can clean even if a reference of self lingers
--- a/imageio/core/request.py
+++ b/imageio/core/request.py
@@ -539,7 +539,7 @@ class Request(object):
         if self._filename_local:
             try:
                 os.remove(self._filename_local)
-            except Exception: # pragma: no cover
+            except Exception:  # pragma: no cover
                 warnings.warn(
                     "Failed to delete the temporary file at "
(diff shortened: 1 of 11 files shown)


imageio/core/request.py
```diff
@@ -534,10 +532,7 @@ def finish(self):
 
         # Remove temp file
         if self._filename_local:
-            try:
-                os.remove(self._filename_local)
-            except Exception:  # pragma: no cover
-                pass
+            os.remove(self._filename_local)
```
almarklein, line 535: I'd keep the try-except. It should very rare, but if the remove fails for some reason, its tough luck, but lets not error our because of it.

FirefoxMetzger: Hm, I think we really should error out because of this. All the situations I can come up with in which the call fails are undesirable, and the user should be informed about it:

1) The file disappears somewhere between the previous read and this `remove`. In this case, ImageIO should complain, because the user is messing with our internals in the wrong way. Alternatively, the user has a _very_ unreliable file system and likely bigger problems than ImageIO failing.

2) The file is still in use and can not be deleted. In this case, we have a bug in our codebase and would like to know about it. Eating the exception won't help.

3) ImageIO does not have permission to delete the file.  This is odd, because we created the file there earlier and (unless I am mistaken) create and delete permission comes as a pair. Either way, this is likely a bug in ImageIO as well, because we should avoid writing temp files to places we can't clean up.

That's the three scenarios I could come up with and in all of these, I'd rather fail noisily than end up with a potentially hard-to-trace bug.

almarklein: In my experience its hard to estimate what can go wrong if you talk about file systems. I've been bitten by this in other projects. But you're right that swallowing the exceptions may also hide internal bugs. What about we display a warning instead of raising an error?

FirefoxMetzger: Yes, we could go with a warning instead of an error. 

Do you have a good idea on how to set up a test that would trigger this case? If we keep the except block, we might as well cover it with a test.

FirefoxMetzger:
I readded the code that catches the exception on `os.remove` and added a warning that this shouldn't happen.

I also realized that my master branch was out of wack with imageio/master and tried to set the version to v2.11.0 (residue from developing the CD which I didn't notice). I fixed this and rebased this branch on imageio/master.

imageio/core/request.py
almarklein: Maybe this?:
* Create request object and make it use the `_filename_local`.
* Delete that file in the test.
* Call request.finish()

FirefoxMetzger: @almarklein  This would fail earlier on in the same function (line 509) because we can't read from it. This is what caused me to remove the try/except block initially.

What we could possibly try a weird hack like:

- Create request object and make it use the _filename_local
- change the permissions on the directory to read-only (is this from inside python?)
- Call request.finish()
- set the permissions on the directory to read/write again

almarklein: Sometimes writing a test is not worth the effort. I would also be perfectly ok with:
* Open an interactive session
* Create request object and make it use the _filename_local
* Manually make that file read-only
* Call request.finish()
* Check that it does what you expect
* Add a comment that this piece of code was only tested manually.

FirefoxMetzger: @almarklein Sure. I will see if I can work out how to tell codecove to ignore the except branch. 

From the top of my head, I could ignore any warnings that we raise, but that seems like too coarse a solution. Maybe there is some line magic to ignore a scope...

almarklein (approved):
👍

PR #675, opened 2021-11-03
Fix ffmpeg can_read for webcams with id greater than 10

hashb:
In our use case, we have over 20 webcams connected to a device at any given time. Currently only webcams up to id 9 are supported. This PR adds support for arbitrary number of cameras.

Please let me know if you want me to add any tests

### DIFF
--- a/imageio/plugins/ffmpeg.py
+++ b/imageio/plugins/ffmpeg.py
@@ -291,7 +291,7 @@ class FfmpegFormat(Format):
             self._arg_input_params += ffmpeg_params or []  # backward compat
             # Write "_video"_arg - indicating webcam support
             self.request._video = None
-            if self.request.filename in ["<video%i>" % i for i in range(10)]:
+            if re.match(r"<video\d+>", self.request.filename):
                 self.request._video = self.request.filename
             # Specify input framerate?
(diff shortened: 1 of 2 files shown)


FirefoxMetzger (changes requested):
Thanks for this PR @hashb ! I agree that it makes a lot of sense to change to an implementation that isn't arbitrarily limited to a maximum number of video devics.

The PR itself looks pretty good. I would like to see one thing added here, which is a regression test at the bottom of `tests/test_ffmpeg.py` to make sure we don't accidentally undo this behavior later on. Can you add this, or should I do that?

kchenna-nimble:
done. I have tested this only on linux but looks like it should work for other platforms too.

hashb:
I have tested this only on linux but looks like it should work on other platforms too.

FirefoxMetzger (approved):
Nice work @hashb . I added two suggestions on the test, but otherwise I have nothing else to add.

If you agree with them, and the CI passes it is LGTM 🚀

FirefoxMetzger:
Thanks for this contribution @hashb 🚀 The CD should release them tonight, so try upgrading imageio tomorrow and it should be fixed.

PR #678, opened 2021-11-05
BUG: Fix imwrite bug when filename contains '#'.

### LINKED ISSUE #674: imageiov2.10.1. The imwrite will raise ValueError when filename include "#".
Imageio V2.10.1
```
import imageio
import numpy as np

imageio.imwrite('#1.jpg', np.zeros([512, 512, 3], dtype=np.uint8))
```
It will raise ValueError: Could not find a format to read the specified file in ImageMode.single_image mode.

But it will good. No raise.
```
imageio.imwrite('#1.jpg', np.zeros([512, 512, 3], dtype=np.uint8), '.jpg')
```

One-sixth:
Fix #674

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -783,6 +783,10 @@ def test_imwrite_not_array_like():
         imageio.imwrite("foo.bmp", "asd")
 
 
+def test_imwrite_symbol_name():
+    imageio.imwrite('''#!~@$%^&`-+{};',.() []_=.jpg''', np.zeros([128, 128, 3], dtype=np.uint8))
+
+
 def test_legacy_empty_image():
     with pytest.raises(RuntimeError):
         with iio.imopen("foo.bmp", "wI", format="GIF-PIL") as file:
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -786,11 +786,11 @@ def test_imwrite_not_array_like():
 def test_imwrite_symbol_name():
     # this is a regression test for https://github.com/imageio/imageio/issues/674
     if sys.platform == "linux":
-        name = '''#!~:@$%^&`-+{};',.()? []_=.jpg'''
+        name = """#!~:@$%^&`-+{};',.()? []_=.jpg"""
     elif sys.platform == "win32":
-        name = '''#!~@$%^&`-+{};',.() []_=.jpg'''
+        name = """#!~@$%^&`-+{};',.() []_=.jpg"""
     elif sys.platform == "darwin":
-        name = '''#!~@$%^&`-+{};',.()? []_=.jpg'''
+        name = """#!~@$%^&`-+{};',.()? []_=.jpg"""
     else:
         pytest.skip("Unknown OS.")
     tmp_request = imageio.core.Request(name, "r")
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -786,9 +786,9 @@ def test_imwrite_not_array_like():
 def test_imwrite_symbol_name():
     # this is a regression test for https://github.com/imageio/imageio/issues/674
     name = '''#!~@$%^&`-+{};',.() []_=.jpg'''
-    imageio.imwrite(name, np.zeros((128, 128, 3), dtype=np.uint8))
-    im = imageio.imread(name)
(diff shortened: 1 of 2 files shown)


tests/test_core.py
```diff
@@ -783,6 +783,13 @@ def __init__(self):
         imageio.imwrite("foo.bmp", "asd")
 
 
+def test_imwrite_symbol_name():
```
FirefoxMetzger, line 786:
```suggestion
def test_imwrite_symbol_name():
    # this is a regression test for https://github.com/imageio/imageio/issues/674
```

added xref to keep track of things.

tests/test_core.py
```diff
@@ -783,6 +783,13 @@ def __init__(self):
...
+    name = '''#!~@$%^&`-+{};',.() []_=.jpg'''
```
FirefoxMetzger, line 787:
```suggestion
    if sys.platform == "linux":
        name = '''#!~:@$%^&`-+{};',.()? []_=.jpg'''
    elif sys.platform == "win32":
        name = '''#!~:@$%^&`-+{};',.() []_=.jpg'''
    elif sys.platform == "darwin":
        name = '''#!~@$%^&`-+{};',.()? []_=.jpg'''
    else:
        pytest.skip("Unknown OS.")
```

On linux `?` may be in the path, but is reserved for URIs. I think it makes sense to test this explicitly, too.

Same for `:` on any OS except MacOS.

tests/test_core.py
```diff
@@ -783,6 +783,13 @@ def __init__(self):
...
+    imageio.imwrite(name, np.zeros((128, 128, 3), dtype=np.uint8))
+    im = imageio.imread(name)
+    assert im.shape == (128, 128, 3)
```
FirefoxMetzger, line 790:
```suggestion
    tmp_request = imageio.core.Request(name, "r")
    assert tmp_request.extension == ".jpg"
    tmp_request.finish()
```

I think explicitly testing the `Request` object here is cleaner. This way we don't need a plugin installed that can read/write JPG. It also tests that the extension is what we expect it to be, which was the primary symptom that made us notice this.

Alternatively, we can do both assertions. Pillow is a core dependency, so we should always be capable of reading and writing JPEG.

FirefoxMetzger (approved):
Thanks for this fix @One-sixth . 

I've added two small suggestions on the test. One OS-specific suggestion, and one where you can choose which variant you prefer.

Other than these two comments I have nothing to add. Once the suggestions are resolved and the CI is happy we can merge it in :)

almarklein (approved):
Thanks @One-sixth! I agree with @FirefoxMetzger's suggestions.

tests/test_core.py
One-sixth: I think the symbol ':' should not be added to the win32 test, because ':' cannot appear arbitrarily in win32. It will cause the win32 test to fail.

FirefoxMetzger: @One-sixth Good catch. I updated the suggestion.

tests/test_core.py
```diff
@@ -783,6 +783,21 @@ def __init__(self):
...
+    # this is a regression test for https://github.com/imageio/imageio/issues/674
+    if sys.platform == "linux":
+        name = """#!~:@$%^&`-+{};',.()? []_=.jpg"""
+    elif sys.platform == "win32":
+        name = """#!~@$%^&`-+{};',.() []_=.jpg"""
+    elif sys.platform == "darwin":
+        name = """#!~@$%^&`-+{};',.()? []_=.jpg"""
+    else:
+        pytest.skip("Unknown OS.")
+    tmp_request = imageio.core.Request(name, "r")
```
FirefoxMetzger, line 796:
```suggestion
    tmp_request = imageio.core.Request(name, "w")
```

The CI fails when creating the read request, because it checks if the file exists upon construction (It should probably do that upon first read instead). Writing will not perform this check, and we technically don't need it here, so preparing to write is probably the better thing to do.

FirefoxMetzger:
LGTM 🚀 Thanks for spotting - and fixing - this bug @One-sixth !

The fix should be on pypi tomorrow.

Issue #677, opened 2021-11-04
FreeImage identifies images written via PPM-FI as PBM-FI.


FirefoxMetzger:
I found a strange behavior in freeimage. When I write an image using `PPM-FI` then FreeImage identifies the written file as an image in PBM format instead of PPM format. Here is the snippet to show the failure

```python
Python 3.8.8 (tags/v3.8.8:024d805, Feb 19 2021, 13:18:16) [MSC v.1928 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import imageio as iio
>>> from imageio.core.request import Request
>>> fname = "test.ppm"
>>> im = iio.imread("imageio:chelsea.png")
>>> iio.imwrite(fname, im, format="PPM-FI")
>>> test_req = Request(fname, "ri")
>>> assert iio.formats["PBM-FI"].can_read(test_req) 
>>> assert iio.formats["PPM-FI"].can_read(test_req) 
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AssertionError
```

Are PPM and PBM equivalent on a file-level? The following suggests that they are likely very similar:

```python
>>> im1 = iio.imread(fname, format="PPM-FI")
>>> im2 = iio.imread(fname, format="PBM-FI") 
>>> import numpy as np
>>> np.allclose(im1, im2)
True
```

I have traced the source code for writing up to the point we hand the image over to FreeImage and confirmed that it is indeed PPM all the way. Later on, we read the file's header and identify the format based on that using `fi.getFIF(request.filename, "r", request.firstbytes)`. This then re-identifies the file as PBM which is treated as incompatible with PPM.

@almarklein I am a bit lost, because I am only loosely familiar with FreeImage. I discovered this while refactoring `imopen` (more on that in the PR, in short: I want to configure the plugin manager using `imageio.config` and add lazy-loading of plugins upon first use.) and it would be important here that `PPM-FI` correctly reports that it can read PPM-FI.

almarklein:
This is the code path deeper in freeimage. You could see which step eventually sets `ftype`. It might be a bug in freeimage.
https://github.com/imageio/imageio/blob/e1226c751fb30196879620977d6f4072c45f8719/imageio/plugins/_freeimage.py#L584-L592

FirefoxMetzger:
@almarklein This could be indeed inside freeimage. We associate `ftype==14` in our wrapper here:

[...]

Issue #665, opened 2021-10-13
'import imageio' on M1 MacBook with Python 3.9 causes "MemoryError" - installed with conda/pip


slerman12:

slerman12:
```
Python 3.9.1rc1 | packaged by conda-forge | (default, Nov 28 2020, 22:21:58) 
[Clang 11.0.0 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import imageio
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/samlerman/code/miniforge3/lib/python3.9/site-packages/imageio/__init__.py", line 38, in <module>
    from . import plugins
  File "/Users/samlerman/code/miniforge3/lib/python3.9/site-packages/imageio/plugins/__init__.py", line 87, in <module>
    from . import freeimage
  File "/Users/samlerman/code/miniforge3/lib/python3.9/site-packages/imageio/plugins/freeimage.py", line 13, in <module>
    from ._freeimage import fi, download, IO_FLAGS, FNAME_PER_PLATFORM  # noqa
  File "/Users/samlerman/code/miniforge3/lib/python3.9/site-packages/imageio/plugins/_freeimage.py", line 1332, in <module>
    fi = Freeimage()
  File "/Users/samlerman/code/miniforge3/lib/python3.9/site-packages/imageio/plugins/_freeimage.py", line 428, in __init__
    def error_handler(fif, message):
MemoryError
```

slerman12:
Hi, is there still a known bug with M1 Mac? I get the following error:

`RuntimeError: No ffmpeg exe could be found. Install ffmpeg on your system, or set the IMAGEIO_FFMPEG_EXE environment variable.`

FirefoxMetzger:
@slerman12 Your error doesn't sound like it is related to this issue. Instead, you simply didn't perform the post-install step needed to get the ffmpeg plugin. In the docs on the ffmpeg plugin you can find how to do this: https://imageio.readthedocs.io/en/stable/_autosummary/imageio.plugins.ffmpeg.html#module-imageio.plugins.ffmpeg

slerman12:
The strange thing is I definitely have installed imageio-ffmpeg.

```console
pip install imageio-ffmpeg
```

However, when I run the recommended install `pip install imageio[ffmpeg]`, I get this error, might be a mistake:
`zsh: no matches found: imageio[ffmpeg]`.

FirefoxMetzger:
[...]

Issue #580, opened 2020-12-17
Build imageio without ctypes


oeway:
Hi, I would like use imageio in Pyodide but it has been currently disabled because of the issue related to `ctypes`. As I understand, an issue was encountered when building imageio as a dependency of scikit-image in Pyodide ([see here for the issue in the scikit-image PR](https://github.com/iodide-project/pyodide/pull/866#issuecomment-746021613)). 

Can someone please take a look and see if we can make ctypes as an optional dependency for imageio? Maybe allowing a build flag to mask out `freeimage` plugin?

almarklein:
The freeimage plugin indeed uses ctypes. There already is a plan to move that to its own repository. But I'm not 100% sure whether there are other plugins that use ctypes.

FirefoxMetzger:
I just checked Pyodide and they seem to happily depend on the latest ImageIO. The plugin that is discussed in this issue (freeimage) still uses ctypes, but is now optional and dynamically imported only if needed.

I will close this issue because it appears to be solved, but if it isn't please feel free to let me know and I will reopen.

Issue #334, opened 2018-04-12
Improve error message when reading an EXR file without plugin installed


Overdrivr:
It's just a minor thing, but if the EXR plugin is not installed, EXR images cannot be read and `imageio.imread` fails with this error message:
 
```
>>> img = imageio.imread(path)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Users\remib\.virtualenvs\qzMurbXh\lib\site-packages\imageio\core\functions.py", line 206, in imread
    reader = read(uri, format, 'i', **kwargs)
  File "C:\Users\remib\.virtualenvs\qzMurbXh\lib\site-packages\imageio\core\functions.py", line 126, in get_reader
    'in mode %r' % mode)
ValueError: Could not find a format to read the specified file in mode 'i'
```

All that is required to fix this is to run

```
imageio.plugins.freeimage.download()
```

Maybe this call should be mentioned in the error message ?

almarklein:
That would be much better indeed. Would not be trivial to add though. When the format manager iterates over the plugins to find one capable of handling a request, plugins can only say yes or no. For this to work we would need some sort of option to allow a plugin to say "no, but ...".

I think this would be worth the effort, but this will have to wait for when I have/allocate some serieus imageio-time :)

FirefoxMetzger:
I think we can fix this soon™ (via #680 ).

If a user tries to read using `iio.imread(..., format="EXR-FI")` and FreeImage is not installed, it will ask the user to install FreeImage first.

For files that have the `.exr` extension this will still give a generic error, but we can discuss/look into if this should raise a warning or error that it could be read after installing FreeImage.

FirefoxMetzger:
Okay, this will get fixed in #680. 

If no currently installed backend can fulfill the request we check the image resource for its extension. If it has an extension, we look up all plugins that are associated with this extension and check if they are installed. For each missing plugin, we inform the user in the error message that installing that plugin might resolve this error.

This should turn the above exception into something like:

```
img = imageio.imread(path)
...
OSError: Could not find a backend to open `path` with iomode `r`. Based on the extension, the following plugins might add capable backends:
  EXR-FI: pip install imageio[freeimage]
```

@almarklein What do you think?

almarklein:
Sounds good!

PR #680, opened 2021-11-05
FEAT: choose plugin based on extension and lazy-import plugins

FirefoxMetzger:
Closes https://github.com/imageio/imageio/issues/641
Closes https://github.com/imageio/imageio/issues/677 (fix is implemented here)
Closes https://github.com/imageio/imageio/issues/665
Closes https://github.com/imageio/imageio/issues/334

Issues that may be closed or will become easy to close
https://github.com/imageio/imageio/issues/580

This PR is essentially a large refactor of the plugin manager that aligns `imopen`, the old `FormatManager` and the docs and forces them to remain in sync. It does so by adding (and changing) a few things:

1. `imageio.core` now relies on `imageio.config` for global configuration (formerly used for docs only).
2. It adds `known_plugins` and `known_extensions` in `imageio.config` to track extensions and plugins.
3. `imopen` now looks up the request's extension in `known_extensions` and (upon success) tries relevant plugins first. (https://github.com/imageio/imageio/issues/641)
4. `imopen` now checks `known_plugins` when checking all plugins (if extension-based handling doesn't work).
5. `imopen`'s state is no longer needed, so it is now a function (formerly it was a class).
6. `FormatManager` now internally delegates all format lookups to `imopen`. (formerly, `imopen` called the format manager as a fallback when it couldn't find a plugin).
7. `FormatManager`'s `add_format` now adds a format to `known_formats` and it's `sort` method sorts `known_formats`.
8. `imopen` will now lazy-import a plugin during the first time it selects that plugin to handle a request. (fixes https://github.com/imageio/imageio/issues/665 because FreeImage will only be imported when actually used by the code)
9. `imopen` will now display a friendly error message in cases a request could not be handled, but a user might be able to install a plugin to handle it. (https://github.com/imageio/imageio/issues/334)
10. `FormatManager` will now lazy-import a plugin during the first time it is asked to return a format of that plugin.
11. It adds a check to FreeImage's `_can_read` to handle a bug in FreeImage related to PPM and PBM files (https://github.com/imageio/imageio/issues/677)

Lazy-importing should also be a good first step towards factoring out some of the backends (including FreeImage), which would eventually address https://github.com/imageio/imageio/issues/580.

---

What is left for this PR:

- [x] Docstrings for the relevant functions.
- [x] Tests to hit the PR coverage target.


FirefoxMetzger:
@almarklein RTR. 

Codecov says that 0.5% of the code in the PR is not covered, but it doesn't indicate the line that it thinks is missing. I did use the opportunity to add coverage for some pieces we were not covering before though, so overall, we are improving coverage with this PR.

imageio/core/imopen.py
```diff
@@ -1,23 +1,109 @@
...
+
+
+def imopen(
+    uri,
+    io_mode: str,
+    *,
+    plugin: str = None,
+    legacy_mode: bool = True,
+    **kwargs,
+):
```
almarklein, line 64: Really happy to see this being a function ;)

imageio/core/request.py
```diff
@@ -550,6 +552,7 @@ def get_result(self):
         """For internal use. In some situations a write action can have
         a result (bytes data). That is obtained with this function.
         """
+        # Is there a reason to disallow reading multiple times?
```
almarklein, line 555: Mostly because the Request object is intended for one-time usage. But the below is also for removing references, to make it easier for the gc and thus lower the chance on memory leaks.

imageio/plugins/__init__.py
```diff
@@ -79,40 +79,6 @@
...
-from . import dicom
-from . import npz
-from . import swf
-
-from . import feisem  # special kind of tiff, uses _tiffile
-from . import simpleitk  # depends on itk or SimpleITK
-from . import gdal  # depends on gdal
-
-from . import lytro
-from . import spe
```
almarklein, line 114: Does all this mean that a user would have to `import imagio.plugins.pillow` in order to use the Pillow plugin?

almarklein (approved):
This is an impressive piece of work @FirefoxMetzger! I reviewed this, but its a lot, so I could not go into all the details, but this looks good! I also see a lot of new tests, so that gives confidence.

A few comments, but in general LGTM.

imageio/core/imopen.py
FirefoxMetzger: @almarklein I will move imopen into `v3_api.py` in the future, because it seems to belong there now. With this change, we will have a purely functional user-facing API, and I think there is no need for an extra file for `imopen`.

imageio/core/request.py
FirefoxMetzger: 😅 Well, currently we are reusing the same request object many times. Mainly because I wasn't aware that it is intended to be a one-time throw-away type object.

`imopen` passes the same Request to each plugin (which is the source of https://github.com/imageio/imageio/issues/642) and the same request is re-used by a plugin when successive `read()` or `write()` calls are performed. If you can think of any problems (beyond https://github.com/imageio/imageio/issues/642) that this could cause, please let me know so that we can track them down to make sure things stay nice and functional.

imageio/plugins/__init__.py
FirefoxMetzger: Nope 🚀 We run this for the user ([link to relevant lines](https://github.com/FirefoxMetzger/imageio/blob/68819d27a669364daba53e271e4583868c81b335/imageio/config/plugins.py#L101-L102)).

Roughly, what happens now is 

```python

import imageio as iio 

# modules imported up to this point
# - imageio
# - numpy

iio.imread("some.tiff")  # tifffile is automatically imported here

# modules imported up to this point
# - imageio
# - numpy
# - tifffile

iio.imread("imageio:chelsea.png")  # pillow is automatically imported here

# modules imported up to this point
# - imageio
# - numpy
# - tifffile
# - pillow

iio.imread("some.bogus")  # automatically imports and tries all plugins listed in `iio.config.known_plugins`
# raises IOError

# modules imported up to this point
# all of them ... we had to load them to try our best to read `some.bogus`
```

The above introduces a one-time cost for each plugin's first `imread`, but that can be avoided by your suggestion, i.e., if the user manually calls `import imagio.plugins.pillow` then the one-time cost for loading the plugin is paid there instead of being paid during `imread`.

On the flip side, if a user never uses `imageio-ffmpeg`, then he/she will never pay the cost of importing it (nor will he/she ever face exceptions because that plugin doesn't work on their machine). It also avoids us doing the "is this backend installed" dance inside a plugin. The plugin simply assumes that it is installed, and if it isn't it fails with an ImportError, which `imopen` picks up and handles appropriately.

FirefoxMetzger:
I noticed a "regression" introduced in this PR: If the user selected a specific format using the old `FormatManager`, it would be used immediately without doing any checks. In `imopen`, a `LegacyPlugin` (wrapping an old `Format`) will test `can_read` first and will fail if the format claims it can't read.

As a result, `imageio-ffmpeg` can no longer read from FileObjects or byte strings, because its plugin only checks the file extension, which is `None` in this case and gets rejected. 

@almarklein This would need to be fixed by having ffmpeg check the header of the file (or what ever the equivalent thing is for that backend). Any thoughts on how to best go about this?

imageio/core/request.py
almarklein: mmm but then its not disallowed, right?

imageio/plugins/__init__.py
almarklein: I agree that this approach has certain advantages. I think I also prefer it over the earlier approach.

One note though, is that we may want to take measures for people freezing apps with imageio. See e.g. https://github.com/pygfx/wgpu-py/tree/main/wgpu/__pyinstaller

almarklein:
> Any thoughts on how to best go about this?

Nope. I bet that there would be an awful lot of headers that ffmpeg could parse, but I don't think we can give ffmpeg a few bytes and ask it to let us know if it looks familiar :)  I guess this is the way it is?

imageio/plugins/__init__.py
FirefoxMetzger: @almarklein I've been meaning to ask about the files related to freezing support in our repo. I never come across this before and haven't seen it anywhere else. When I googled it a while back I only found old resources that seemed to be from the python2.7 era, and I couldn't find anything in the official python docs (probably not looking at the right place?)

Is this still a thing that people actually use? And, consequentially, is it worth supporting? What would it be used for?

FirefoxMetzger:
> but I don't think we can give ffmpeg a few bytes and ask it to let us know if it looks familiar :) I guess this is the way it is?

Theoretically, there is [ffprobe](https://ffmpeg.org/ffprobe.html#toc-Description) that might give us that information in a relatively efficient way. I don't know how easy/hard it would be to feed in a file object though (I also don't really know how we do that in ffmpeg itself :D).

imageio/core/request.py
FirefoxMetzger: > mmm but then its not disallowed, right?

@almarklein Sorry, I don't think I follow. It is not disallowed to read multiple times?

FirefoxMetzger:
Status update: Currently the tests are breaking (again), because we give DICOM special treatment and allow it to read folders. I ran into this situation multiple times while writing this PR, and I remember dealing with this in previous PRs.

@almarklein Can we formalize folder support for `Requests` (introduce a `URI_FOLDER` or something) and factor out this code-path, or explicitly deny it with a clean error message?

Currently, `Request.get_file()` breaks with a `PermissionDeniedError` somewhere deep inside the library, if the request points to a folder. None of our docs suggest that it is possible to use directories as URIs and, with the exception of DICOM, all plugins break when they encounter a folder. I would consider this a bug and -  in fact - our [official example](https://imageio.readthedocs.io/en/stable/examples.html#read-all-images-in-a-folder) suggests that this is impossible.

At the same time, we have supported folder access for DICOM for several years now, and simply dropping it seems like a bad idea.

imageio/plugins/__init__.py
almarklein: Yes, people definitely freeze apps. E.g to create apps to distribute to non-coders. Since a year or so, PyInstaller provided an API that packages can use to tell pyinstaller what extra data and modules must be included. See these docs: https://pyinstaller.readthedocs.io/en/stable/hooks.html#provide-hooks-with-package, and also the links in my post above (I added a 2nd link).

imageio/core/request.py
almarklein: I mean, the comment states that is is not allowed to use the request object multiple times, but we *do* use it multiple times now, don't we?

almarklein:
Interesting. This is indeed an edge case. In the old version, we seemed to have worked around it. Do you know why we do run into it now? Maybe that gives a hint on how to best solve this.

> Introduce a URI_FOLDER or something

That may seem the most clean solution, but I don't oversee how this will affect the rest of the code. If we can make this work without needing changes in all other plugins then maybe this is a way forward.

Alternatively, we may need to force reading a dicomdir by providing the path to an arbitrary file in that directory. This is also not pretty though...

Do you think we can keep DICOM working as-is if the plugin is explicitly selected? That might help.

FirefoxMetzger:
> Do you know why we do run into it now? Maybe that gives a hint on how to best solve this.

We ran into this before but simply patched it in the function that encountered it instead of dealing with it formerly. For example:

https://github.com/imageio/imageio/blob/b4cbde81164cceba7c38c39334364f127beb5cfb/imageio/core/request.py#L572-L579

In this particular case, we are encountering it because I made tifffile's `_can_read` more sophisticated. I now check the file header in addition to the extension to decide if the file is a TIFF. tifffile gets checked before DICOM (the unit test didn't enforce the DICOM plugin), and that then breaks the test.

> That may seem the most clean solution, but I don't oversee how this will affect the rest of the code. If we can make this work without needing changes in all other plugins then maybe this is a way forward.

I'm not sure if this is possible to change incrementally; my hunch is that it isn't. However, at the moment, we even struggle to detect if the request points to a file vs. a folder, so even if we would be willing to add changes to all plugins in one go, we would first have to establish how this info is shared with plugins.

> Do you think we can keep DICOM working as-is if the plugin is explicitly selected? That might help.

Yes, that's what I implemented in https://github.com/imageio/imageio/pull/680/commits/4923c7d5a2f5ad50b6847c4248072b785e2d7e35. If the request is called with an explicit `plugin=value` then it will work fine. Otherwise, an error will be raised when the URI points to a dictionary.

This is, however, me adding a quick-fix to solve breaking CI, so we might still want to formalize this behavior to avoid doing more quick-fix patches (and remove the ones we have already).

almarklein:
> This is, however, me adding a quick-fix to solve breaking CI, so we might still want to formalize this behavior to avoid doing more quick-fix patches (and remove the ones we have already).

Given the circumstances I think that fix is actually not that hacky. Formalizing support for folders just because of this one plugin, without any further advantages, is also not great. What I mean to say: I don't there is a "pure" way to deal with this, and the current solution is clear for the user and confines the odd feature of being able to read a directory to a specific plugin.

imageio/core/request.py
FirefoxMetzger: @almarklein The comment states that we can't use `Request.get_result` multiple times, and I don't see why this is necessary. There doesn't seem to be anything generally wrong with using the same request object multiple times (apart from #642 ). We have been doing that for the past 8 months or so, doing so passes all of our CI, and so far there are no complaints from downstream 😄.

imageio/plugins/__init__.py
FirefoxMetzger: Okay, then we keep supporting freezing. I am not an expert here, but I doubt we need to do much extra work. The burden of including all required modules lies with the packaging utility, and the plugin code itself is still part of the repo. We simply delay loading it for as long as possible.

FirefoxMetzger:
> Given the circumstances I think that fix is actually not that hacky. Formalizing support for folders just because of this one plugin, without any further advantages, is also not great. What I mean to say: I don't there is a "pure" way to deal with this, and the current solution is clear for the user and confines the odd feature of being able to read a directory to a specific plugin.

Yep. Given my perspective (maintainer level background knowledge of what ImageIO does and does not do), I agree here. Explicit support for folders adds much more problems than it solves. It can be like this for now, but I will dare to make a prophecy and say that we will bump into this issue again further down the road x).

When I said `we might still want to formalize this behavior` I mainly meant that we should have an agreed-upon standard on what we expect to happen here. If it is "we don't do folders, except for legacy reasons (DICOM)" that solves it for me. Though, in this case, I would like to see if we can check if that ability is still needed, or if we can deprecate it in v3 and make things consistent. (This way, `Request` can `raise` when it encounters a folder and it would make for cleaner code.)

On the other hand the question "how to read a folder of images with ImageIO?" is one of the more popular questions that I encounter when I am digging for problems with the library to find features we might wish to add. [Our example on reading folders](https://imageio.readthedocs.io/en/stable/examples.html#read-all-images-in-a-folder) does address this, however, so I don't see it as a high-priority issue for now :)

---

Unless you have any unresolved points @almarklein I think the above should resolve/conclude all open points. 

There is still the `ffmpeg` "regression"/bug of rejecting byte streams because ffmpeg's `can_read` isn't sophisticated enough. This, however, requires a bigger change (https://github.com/imageio/imageio/pull/680#issuecomment-968174272) potentially involving changes to `imageio-ffmpeg`, so I would do this in a dedicated PR. If you agree @almarklein , I am happy to accept this temporary regression here and solve it down the line once we pick up the general slack around ffmpeg (missing formats, stream support, potential replacement with pyAV, ...)

imageio/core/request.py
almarklein: I think the confusion comes from the fact that this looks like a public method, but it was meant for internal use only in a specific worflow where it is only called once :)  Setting the `._result` to `None` is just good practice., but we could do that in `finish()` instead. Nevertheless, might be a good idea to make this method private?

imageio/plugins/__init__.py
almarklein: Tools like pyinstaller and cx_freeze look at the actual imports at the top of the file, so it probably won't include the plugins now :)  But implementing the hooks should be relatively simple. Its also easy to test using PyInstaller tooling. I could have a look at this soonish.

imageio/core/request.py
FirefoxMetzger: > Nevertheless, might be a good idea to make this method private?

Yep, or refactor and stop using it altogether. I like the idea of `Request` being something similar to a fat pointer in C++. I like `Request` to facilitate and manage resource (file/stream/byteobject) access, but I am not so happy with it holding any state beyond that (firstbytes/result/...). That, I would like to see refactored, e.g., into a `memmap` akin to https://github.com/imageio/imageio/issues/616 .

imageio/plugins/__init__.py
FirefoxMetzger: > Tools like pyinstaller and cx_freeze look at the actual imports at the top of the file,

This sounds quite scary. Grouping imports at the top of the file is usually fine, but dynamic imports (e.g. with a `import` nested inside a function or via `importlib`) have valid use-cases. `importlib` is part of the python spec since v3.1 so I would hope that packaging tools have adjusted to that by now...

Anyways, let's figure out how this works and then take the appropriate actions from there. The solution can be as simple as adding a note to the docs that says "If you want to freeze your script, remember to import all the plugins your script need at the top".

FirefoxMetzger:
> Could you please remind me how this will affect users?

Sure. On current master, `iio.mimread(b"...", format="ffmpeg")` will happily load the video, assuming that `b"..."` is indeed a byte string containing a valid video. This is because using `format=...` with the old `FormatManager` will skip the `can_read` check.

In this PR `iio.mimread(b"...", format="ffmpeg")` will raise a `ValueError` (I think that's the legacy exception type here), because `imopen` will handle the call itself instead of delegating to `FormatManager`. `imopen` will use a `LegacyPlugin` that wraps the v2 ffmpeg format/plugin and that provides compatibility. In v3 the ability to handle a request is tested during plugin initialization (`Plugin.__init__` may fail with a `InitializationError`), and `LegacyPlugin` does this by calling the wrapped plugin's `can_read`/`can_write` functions. This causes problems for `ffmpeg`, because its `can_read` decides purely based on the file extension and hence returns `False` for `b"..."` (extension is None). 

I currently see no trivial way around this, and the correct way to resolve it is to make `ffmpeg._can_read` more sophisticated and/or port the ffmpeg plugin into a V3 plugin. The latter is already on the TODO list, so I was thinking to clean all that up in one go.

imageio/plugins/__init__.py
almarklein: > This sounds quite scary.

Most freeze apps do it this way, and there are very valid reasons for it. If an import is not in the module level, you cannot know whether it will be imported at runtime. If it would be included regardless, it would add up to the size of the final binary (especially if those imports pull in other imports etc). Also, these imports may well be platform specific.

Tracking this in #684

almarklein:
Ok, seems like very much an edge case to me, and we can probably live with that regression temporarily.

FirefoxMetzger:
@almarklein Okay looks like we agree on all the open points. Anything else to add or should we merge once the tests pass?

(They currently beak because of a merge conflict with master. We introduced `tifffile` as a test dependency and I need to track down why that did not update for the CI runners.)

almarklein (approved):

huangxin168:
I failed to use Nuitka to pack imageio with 'pyav' plugin, 
I saw pyinstaller sulution in imageio doc, is it possible for Nuitka?

FirefoxMetzger:
@huangxin168 I have never used Nuitka myself. That said, based on my brief google search it will discover files to include by scanning the source for `import` statements. This doesn't work for dynamic imports, however, and the docs suggest to use `--include-plugin-directory` to manually include a folder. I suspect that you can use that option to include ImageIO and all its plugins (which are dynamically imported) to make this work.

Since Nuitka also acts as a compiler, another point to investigate is how it interacts with C extensions. This is because `pyav` is a cython project, which means that some of its modules are actually compiled C-code, which may or may not cause problems depending on how smart Nuitka is.

huangxin168:
@FirefoxMetzger   I tried --include-plugin-directory, unfortunately the same.  Thank you for your help.

Issue #681, opened 2021-11-08
volread on tifffile produces the wrong shape


FirefoxMetzger:
This is a regression from 2.9.0 first shared by @mkcor in https://github.com/scikit-image/scikit-image/pull/5262.

Reproducing example:

```python
img = iio.volread('http://cmci.embl.de/sampleimages/NPCsingleNucleus.tif')
img.shape  
# in v2.10.3 (30, 180, 183)
# in v2.9.0 (15, 2, 180, 183)
```

It traces back to something that, at least to me, is unexpected behavior in v2.9.0. Calling `imageio.imread` on the above tiff (in either version of ImageIO) returns a single page of the file, not a single image. I.e.,

```python
import imageio as iio
img = iio.imread('http://cmci.embl.de/sampleimages/NPCsingleNucleus.tif')
img.shape  # (180, 183) not the expected (2, 180, 183)
```

Consequentially, iterating over single images and stacking them, doesn't yield a stack of channel-first images, but a stack of pages:

```python
reader = iio.get_reader('http://cmci.embl.de/sampleimages/NPCsingleNucleus.tif', mode="i") 
img = np.stack([reader.get_data(index=x) for x in range(reader.get_length())])
img.shape # (30, 180, 183)
```

Reading a volume, on the other hand, does the expected thing and produces a stack of channel-first images (in v2.9.0)

```python
img = iio.volread('http://cmci.embl.de/sampleimages/NPCsingleNucleus.tif')
img.shape  # (15, 2, 180, 183)
```

however, does the page-stacking thing in v2.10.3:

```python
img = iio.volread('http://cmci.embl.de/sampleimages/NPCsingleNucleus.tif')
img.shape  # (30, 180, 183)
```

@cgohlke @almarklein @mkcor (and others who use TIFF more than me) What is expected behavior here? Would you expect `imread` to return a single *image* (shape: (2, 180, 183)) or a single *page* (shape: (180, 183))? I am leaning towards a single image, but I am open for comments here.

Depending on this answer, I would either look into a bugfix fir `get_data(index=...)` (any version) or `volread(...)` (v2.10.3).

FirefoxMetzger:
> I wonder if you'd like to introduce an optional channel_axis parameter like we do in scikit-image @grlee77 😉 (where the default usually corresponds to channel-last images: channel_axis=-1)...

[...]

PR #682, opened 2021-11-09
BUG: Resolve regression on imagej TIFF hyperstacks

FirefoxMetzger:
Closes: #681 

This PR implements a fix that resolves the regression reported in #681 . It also cleans up the tifffile plugin a bit by adding the ability to read multiple volumes (if present in the data). 

It does not address the `imread` bug discussed in the same issue. That is, it still returns a single page of the TIFF instead of returning the first image.

@almarklein It appears that the `volread` bug is a regression introduced by https://github.com/imageio/imageio/issues/558 which attempts to fix https://github.com/imageio/imageio/pull/559. That PR didn't add a regression test, and the test image from #559 has disappeared, so it's not easy to guard against regression for that issue.

### DIFF
--- a/tests/test_tifffile.py
+++ b/tests/test_tifffile.py
@@ -50,7 +50,8 @@ def test_tifffile_reading_writing():
         assert ims[i].shape == im2.shape
         assert (ims[i] == im2).all()
 
-    # Read all planes as one array - we call it a volume for clarity
+    # volumetric data
+    imageio.volwrite(filename1, np.tile(im2, (3, 1, 1, 1)))
     vol = imageio.volread(filename1)
     vols = imageio.mvolread(filename1)
     assert vol.shape == (3,) + im2.shape
@@ -58,7 +59,7 @@ def test_tifffile_reading_writing():
     for i in range(3):
         assert (vol[i] == im2).all()
 
-    # remote multipage rgb file
+    # remote channel-first volume rgb (2, 3, 10, 10)
     filename2 = get_remote_file("images/multipage_rgb.tif")
     img = imageio.mimread(filename2)
     assert len(img) == 2
@@ -93,7 +94,7 @@ def test_tifffile_reading_writing():
 
     # Ensure imread + imwrite works round trip - volume like
     filename3 = os.path.join(test_dir, "test_tiff2.tiff")
-    im1 = imageio.volread(filename1)
+    im1 = np.stack(imageio.mimread(filename1))
     imageio.volwrite(filename3, im1)
     im3 = imageio.volread(filename3)
     assert im1.ndim == 4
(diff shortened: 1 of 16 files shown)


almarklein (approved):
I saw that there has been an extensive discussion on how Tiff works, and how we should interpret it in imageio. I've only superficially read through it. I'm just going to trust your judgement in this matter :)

FirefoxMetzger:
I replaced the test that uses the original image (for which we are still waiting for copyright/licensing) with an artificial hyperstack that is filled with 0s. This reproduces the original problem and circumvents licensing issues.

FirefoxMetzger:
@almarklein I created an entry for `tifffile` in our setup.py and updated our CI to install it when running the tests. Please check if you are okay with this.

The regression test in this PR relies on `tifffile` to create an ImageJ hyperstack (was broken if read via ImageIO). 

Given that our future plan is to remove our vendored tifffile, I am using this opportunity to start moving towards doing that :)

almarklein (approved):

tests/test_tifffile.py
```diff
@@ -3,14 +3,15 @@
...
 import datetime
-
 import numpy as np
 
 from pytest import raises
 from imageio.testing import run_tests_if_main, get_test_dir, need_internet
 from imageio.core import get_remote_file
-
 import imageio
+import imageio as iio
```
mkcor, line 12: I would pick one...

tests/test_tifffile.py
```diff
@@ -130,11 +131,17 @@ def test_tifffile_reading_writing():
         assert md["description"] == "another desc"
 
 
-# # commented out until we get permission to use this image
-# # in our test suite
-# def test_imagej_hyperstack():
-#     img = imageio.volread("NPCsingleNucleus.tif")
-#     assert img.shape == (15, 2, 180, 183)
+def test_imagej_hyperstack(tmp_path):
+    # artifical hyperstack
```
mkcor, line 135:
```suggestion
    # create artifical hyperstack
```

tests/test_tifffile.py
FirefoxMetzger: > I would pick one...

Yep, I agree @mkcor and you are very correct in pointing this out.

I will move the entire test suite to `import imageio as iio` once I go through and do a test-styling PR. In the current suite, some files use `import imageio`, others use `import imageio as iio` and some use both. The change here actually doesn't worsen the situation overall 😅 .

There are a few other test-related things on my bucket list (serve standard images via a pytest-fixture, ensure that plugins are enforced for plugin-specific tests, etc.) so I think it makes sense to tread all of them in a dedicated PR that cleans up the test suite ... maybe we can even improve coverage in the process (wishful thinking™).

So yes, will be fixed. Unfortunately not in this PR, but in a future one instead. On the bright side, this PR will merge once the tests all pass :)

Issue #687, opened 2021-11-18
imageio.imwrite no longer works with io.BytesIO


ugtony:
Hi, 
The code below works well imageio==2.10.4 while doesn't work with imageio==2.11.0, 
please take a look, thanks.
```
import io
import imageio
import numpy as np

image = np.zeros((200,200))
bytes_io = io.BytesIO()
imageio.imwrite(bytes_io, image, 'jpeg')
```

error message
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/dist-packages/imageio/core/imopen.py", line 149, in imopen
    return config.plugin_class(request, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/imageio/config/plugins.py", line 108, in partial_legacy_plugin
    return LegacyPlugin(request, legacy_plugin)
  File "/usr/local/lib/python3.8/dist-packages/imageio/core/legacy_plugin_wrapper.py", line 68, in __init__
    raise InitializationError(
imageio.core.request.InitializationError: `JPEG-FI` can not write to `<_io.BytesIO object at 0x7f100f74f270>`.

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.8/dist-packages/imageio/core/functions.py", line 195, in imwrite
    with imopen(uri, "wi", plugin=format) as file:
  File "/usr/local/lib/python3.8/dist-packages/imageio/core/imopen.py", line 167, in imopen
    raise err_type(err_msg) from err_from
RuntimeError: `jpeg` can not handle the given uri.
```

FirefoxMetzger:
That reproduces for me on master. Looks like an omission in the new fast-path for file extensions. The following workaround (using the v3 API) appears to work fine:

```python
import io
import imageio
import numpy as np

image = np.zeros((200,200))
bytes_io = io.BytesIO()

# explicitly set the backend to use and mode to write in
iio.v3.imwrite(bytes_io, image, plugin="pillow", format="jpeg", mode="L")
```

I will have a look into where legacy imwrite breaks soon™ (hopefully this afternoon). It sounds like it should be a small fix, in which case we should be able to release it this week.

PR #688, opened 2021-11-18
BUG: Fix BytesIO regression for legacy pillow

FirefoxMetzger:
Closes: https://github.com/imageio/imageio/issues/687

The fix changes two things:
1. Make ImageIO prefer pillow over freeimage for the .jpeg extension
2. makes `PillowFormat._can_write` return `True` for `URI_FILE` requests

It also adds regression tests to both the v2 and v3 pillow plugin.

### DIFF
--- a/imageio/config/extensions.py
+++ b/imageio/config/extensions.py
@@ -415,7 +415,7 @@ extension_list = [
     FileExtension(
         name="Joint Photographic Experts Group",
         extension=".jpeg",
-        priority=["pillow", "JPEG-FI", "JPEG-PIL", "ITK", "GDAL"],
+        priority=["pillow", "JPEG-PIL", "JPEG-FI", "ITK", "GDAL"],
     ),
     FileExtension(
         name="JPEG 2000",
--- a/imageio/plugins/pillow_legacy.py
+++ b/imageio/plugins/pillow_legacy.py
@@ -173,7 +173,7 @@ import threading
 import numpy as np
 
 from ..core import Format, image_as_uint
-from ..core import RETURN_BYTES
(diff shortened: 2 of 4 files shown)


FirefoxMetzger:
@almarklein RTR. Let's see if there are other places where plugins think they are less capable than they actually are.

Carreau (approved):
That fixes the napari test suite, it might not be much but at least a data point.

FirefoxMetzger:
@Carreau Could you point me to the broken log? Maybe there is a test or two that I can <s>steal</s> faithfully borrow so that our test suite can become more complete :)

Carreau:
I think the current test you added is fine, 

https://github.com/napari/napari/runs/4246769223?check_suite_focus=true is one of the failing test suite.

it gets down to:
```
        with BytesIO() as file_obj:
            imsave(file_obj, self.image, format='png') # self.image is an X,Y,4 array RGBA 0-255.  IIRC
            file_obj.seek(0)
            png = file_obj.read()

```

Carreau:
(at least I think this is where it fails, as it's in a pytest fixture it's non-trivial to find).

FirefoxMetzger:
> it gets down to:
> [...]

@Carreau This goes off on a tangent, but any particular reason you are involving bytesIO here? If all you want is a `bytes` array, you can use the `<bytes>` special target (though I really think we should rename it into `imageio:bytes`):

```python
png = iio.imwrite("<bytes>", self.image, format="png")
```

Carreau:
No particular reason I'm not the one who wrote the code, and it seem to be from May 2020, so we'll put that on covid.

almarklein (approved):

Issue #679, opened 2021-11-05
"Local" plugins


anntzer:
I would like to be able to "locally" register a plugin, within a bounded context.  An example would be a library where I know, locally that I would be dealing with ".dat" / ".raw" files which actually have a certain binary format, so I want to register a plugin for these, but I don't want this plugin to affect the entire Python process (because ".dat"/".raw" could be anything, and I don't want to make promises for other modules in the same process that could also be using imageio).  IOW, I'd like something like (API names up to bikeshedding)
```python
def my_libs_public_api(path: Path, ...):
    # path can be a .dat/.raw, or perhaps any other format supported by imageio
    imopen_with_extra_plugins = imopen.with_plugins(extra_plugins)  # perhaps instantiated at module level
    with imopen_with_extra_plugins(path, "r") as file: return file.read()
```
I *think* this is "mostly" just a matter of having an extra class attribute `_instance_plugins`, which is checked before `_known_plugins`, have `with_plugins` return a new imopen instance with extra entries in `_instance_plugins`, and check `_instance_plugins` before `_known_plugins` in `imopen.__call__`.

Thoughts?

FirefoxMetzger:
There is a chance that I don't fully understand your use-case yet, so my thoughts may not apply here.

First off, are we talking about old-school style plugins (imageio v2) or new style plugins (upcoming imageio v3)? The old style is [documented here](https://imageio.readthedocs.io/en/stable/development/plugins.html#what-is-a-plugin) and the new style is currently undocumented, but you can find an example implementation for pillow [here](https://github.com/imageio/imageio/blob/master/imageio/plugins/pillow.py). Documentation will come soon™, but certainly before v3.

For old style plugins you will have to implement `_can_read` and `_can_write`, which could check the beginning of the file instead of or in addition to the extension when deciding if they can handle the file. We already have several formats (via FreeImage and LYTRO) using `.raw` that coexist and - to my knowledge - get sorted out correctly using this method. 

For reading, we might check if your plugin can read a RAW file, but then move on, because your plugin will reject it.  FOr writing it is generally best practice to specify the full name of the format in `format=` instead of the extension to avoid exactly this situation with aliased extensions.

New style plugins work the same, except that they do this test (and potentially others) upon construction and raise a `InitializationError` if anything goes sideways, e.g., the plugin can't read/write the file.

Would that do the trick in your case?

---

[...]

PR #689, opened 2021-11-18
FEAT: allow plugin objects in plugin kwarg

FirefoxMetzger:
Closes: #679 

Adds the ability to use plugin classes in `imopen`'s `plugin=` kwarg.

```python
iio.imread(url, plugin=MyPlugin)
```

### DIFF
--- a/imageio/core/imopen.py
+++ b/imageio/core/imopen.py
@@ -1,5 +1,5 @@
 from pathlib import Path
-from typing import Any
+from typing import Any, Union
 
 from .request import (
     IOMode,
@@ -65,7 +65,7 @@ def imopen(
     uri,
     io_mode: str,
     *,
-    plugin: str = None,
+    plugin: Union[str, Any] = None,
     legacy_mode: bool = True,
     **kwargs,
 ) -> Any:
@@ -81,7 +81,7 @@ def imopen(
     uri : str or pathlib.Path or bytes or file or Request
         The :doc:`ImageResources <getting_started.request>` to load the image
         from.
-    io_mode : {str}
+    io_mode : str
         The mode in which the file is opened. Possible values are::
 
             ``r`` - open the file for reading
@@ -98,14 +98,14 @@ def imopen(
             ``V`` for multiple volumes,
             ``?`` for don't care (default)
 
-    plugin : {str, None}
+    plugin : str, Plugin, or None
         The plugin to use. If set to None (default) imopen will perform a
         search for a matching plugin.
     legacy_mode : bool
         If true (default) use the v2 behavior when searching for a suitable
         plugin. This will ignore v3 plugins and will check ``plugin``
         against known extensions if no plugin with the given name can be found.
-    **kwargs : {any}
+    **kwargs : Any
         Additional keyword arguments will be passed to the plugin upon
         construction.
 
@@ -139,14 +139,23 @@ def imopen(
     # plugin specified, no search needed
     # (except in legacy mode)
     if plugin is not None:
-        try:
-            config = _get_config(plugin, legacy_mode)
-        except (IndexError, ValueError):
-            request.finish()
-            raise
+        if isinstance(plugin, str):
+            try:
+                config = _get_config(plugin, legacy_mode)
+            except (IndexError, ValueError):
+                request.finish()
+                raise
+
+            def loader(request, **kwargs):
+                return config.plugin_class(request, **kwargs)
+
+        else:
+
+            def loader(request, **kwargs):
+                return plugin(request, **kwargs)
 
         try:
-            return config.plugin_class(request, **kwargs)
+            return loader(request, **kwargs)
         except InitializationError as class_specific:
             err_from = class_specific
             err_type = RuntimeError if legacy_mode else IOError
--- a/imageio/core/imopen.py
+++ b/imageio/core/imopen.py
@@ -173,7 +173,7 @@ def imopen(
         except Exception as generic_error:
             err_from = generic_error
             err_type = IOError
-            err_msg = f"An unknown error occured while initializing `{plugin}`."
+            err_msg = f"An unknown error occured while initializing plugin `{plugin}`."
 
         request.finish()
         raise err_type(err_msg) from err_from
--- a/imageio/core/imopen.py
+++ b/imageio/core/imopen.py
@@ -149,10 +149,13 @@ def imopen(
             def loader(request, **kwargs):
                 return config.plugin_class(request, **kwargs)
 
-        else:
+        elif not legacy_mode:
 
             def loader(request, **kwargs):
                 return plugin(request, **kwargs)
+
+        else:
+            raise ValueError("The `plugin` argument must be a string.")
 
         try:
             return loader(request, **kwargs)
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -975,7 +975,10 @@ def test_imopen_explicit_plugin_input(clear_plugins, tmp_path):
 
     from imageio.plugins.pillow import PillowPlugin
 
-    with iio.v3.imopen(tmp_path / "foo.tiff", "w", legacy_mode=False, plugin=PillowPlugin) as f:
+    with iio.v3.imopen(
+        tmp_path / "foo.tiff", "w", legacy_mode=False, plugin=PillowPlugin
+    ) as f:
         assert isinstance(f, PillowPlugin)
+
 
 run_tests_if_main()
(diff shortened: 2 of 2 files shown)


FirefoxMetzger:
@anntzer What do you think?

anntzer:
Looks nice.  Do you need to explicitly pass legacy_mode=False in that case?  Couldn't passing a plugin instance autoset legacy_mode as needed?

FirefoxMetzger:
> Do you need to explicitly pass legacy_mode=False in that case?

Would using `iio.v3.imread(uri, plugin=MyPlugin)` be good enough? Everything in the `iio.v3` namespace should set `legacy_mode=False` (perhaps with the exception of `imopen` itself, because it also lives in `iio` for legacy compatibility).

You are right that, at the moment, this would bleed into the v2.9 API in the sense that you can use `iio.imread(uri, format=MyPlugin)`. Where `MyPlugin` is some v3 compatible plugin. Chances are we don't want that, so it is probably smart to only do this if we are not in legacy mode.

FirefoxMetzger:
> I don't have a good enough view of the whole API to decide on that.

Ah, sorry, I meant for your use case. On the general ImageIO API level the `legacy_mode` will disappear in v3 together with the legacy behavior, so both options are equally viable.

anntzer:
I guess it depends on what the long term plans for API stability are.  If you're going to deprecate v2 and then remove it and then move v3 into the toplevel namespace and then remove v3 because "it's not necessary anymore", it'd be a bit annoying to have to update the downstream code twice (but I can live with that too, it's not the end of the world).

FirefoxMetzger:
> [...] then remove it and then move v3 into the toplevel namespace and then remove v3 because "it's not necessary anymore", it'd be a bit annoying to have to update the downstream code twice (but I can live with that too, it's not the end of the world).

I agree, and I have been thinking about this as well. I think I would like to keep the `iio.v3` namespace around even after ImageIO v3.0.0 is released. This way people transitioning to v3 early can keep their code as-is and the namespace seems like a fairly clean thing to have.

We could even think about having an `iio.v2` namespace during the ImageIO v3.X cycle as a fallback option for people migrating slowly/incrementally. Actually, I think I kind of like this idea. The root namespace can then be v3 during v3.X, and we can repeat this cycle for every version major, e.g., add a v4 namespace during the end of v3 and drop the v2 namespace in ImageIO v4.

anntzer:
Seems good to me!

imageio/core/imopen.py
```diff
@@ -139,14 +139,26 @@ def imopen(
...
+            try:
+                config = _get_config(plugin, legacy_mode)
+            except (IndexError, ValueError):
+                request.finish()
+                raise
+
+            def loader(request, **kwargs):
+                return config.plugin_class(request, **kwargs)
+
+        elif not legacy_mode:
```
almarklein, line 152: I know we cant do isinstance here, but maybe do a `hasattr` to see whether this looks like a plugin?

FirefoxMetzger: That would be a pretty extensive test, but it is certainly possible. Currently, we would have to run that check for `read`, `write`, `iter`, `close`, `get_meta`, `__enter__` and `_exit__`.  We could do this cleanly by factoring this out into a separate `looks_like_a_plugin` function. However, at this point why don't we ask people to inherit from a common abstract `Plugin`? I am currently writing one anyway as a means of documenting the plugin interface. people will already have to depend on ImageIO to get access to the `Request` object used by `Plugin.__init__(request:Request, **kwargs)` so it won't be a big change.

Also, do you think it will happen often that we encounter classes/objects that (accidentally) succeed on `MyClass(request, **kwargs)`, but lack the plugin interface? It seems like a quite specific thing to "accidentally get right", and the cases that do fail we catch in the subsequent `try` block already. I guess if you try hard enough then this isn't bulletproof validation, but in that case `hasattr` won't stand much ground either.

My current thinking was to do it "numpy style" and omit explicit input validation (beyond what is needed to decide control flow). If some functionality that we didn't document works, good for you. If it breaks (potentially with a cryptic error message) then because you are doing non-standard stuff. Not sure to what extent we should try to catch these cases and try to provide clear error messages.

almarklein: I was thinking more along the lines of checking one or two attributes, mainly to avoid the use case of accidentally providing something that is definitely not a plugin.

If a non-plugin is provided now, how will it fail, with what error message. If this message is clear enough to see what went wrong, then lets not bother. But if its hard to deduce that this is the point where it went wrong, then it would be worth adding a check here.

FirefoxMetzger: Assiming the constructor fails (e.g., because it doesn't know what to do with `Request` and ends up doing the wrong thing), then this piece of code will catch that:

https://github.com/imageio/imageio/blob/924e1c5c43576acf16247c3c47b7baf6c6f68fe6/imageio/core/imopen.py#L161-L164

so the error looks something like this:

```plugin
>>> img = iio.v3.imread("imageio:chelsea.png", plugin=dict)         
  File "C:\Users\Sebastian\Documents\Coding-Projects\imageio\imageio\core\imopen.py", line 161, in imopen
    return loader(request, **kwargs)
  File "C:\Users\Sebastian\Documents\Coding-Projects\imageio\imageio\core\imopen.py", line 155, in loader
    return plugin(request, **kwargs)
TypeError: 'Request' object is not iterable

The above exception was the direct cause of the following exception:

  File "<stdin>", line 1, in <module>
  File "C:\Users\Sebastian\Documents\Coding-Projects\imageio\imageio\core\v3_api.py", line 39, in imread
    with imopen(uri, "r", **plugin_kwargs) as img_file:
  File "C:\Users\Sebastian\Documents\Coding-Projects\imageio\imageio\core\imopen.py", line 179, in imopen
    raise err_type(err_msg) from err_from
OSError: An unknown error occured while initializing `<class 'dict'>`
```

almarklein: Right, perhaps instead of doing a check, we can catch those errors and reraise with a better error message? If you don't think this makes sense, then let's go with what we have now :)

FirefoxMetzger: > Right, perhaps instead of doing a check, we can catch those errors and reraise with a better error message?

@almarklein I'm not sure I understand. What kind of error message would you expect to see here?

almarklein: I mean that the error messages you list above may be rather cryptic and depend a lot on what is provided. If we catch the error where the plugin is instantiated, we can re-raise with something like "the provided plugin does not look like a plugin. Original error message: ..."

FirefoxMetzger: I think this is already very close to what we are doing:

> If we catch the error where the plugin is instantiated

That's what we do :) the exception handling is a bit lengthy, because we also take care of other causes of failure, e.g. failing because the backend is missing (for real plugins), or plugins rejecting the requires in the indented way (raising `InitializationError`), but it is the same place:

https://github.com/imageio/imageio/blob/924e1c5c43576acf16247c3c47b7baf6c6f68fe6/imageio/core/imopen.py#L148-L164

> "the provided plugin does not look like a plugin. Original error message: ..."

In the current code `generic_error` is the original error message produced by the imposter plugin. Since we `raise from generic_error` its exception will be placed above our exception like in the example above. If that is too verbose, then we can remove that and `raise from None` instead. This would produce the following stack trace:

```python
  File "<stdin>", line 1, in <module>
  File "C:\Users\Sebastian\Documents\Coding-Projects\imageio\imageio\core\v3_api.py", line 39, in imread
    with imopen(uri, "r", **plugin_kwargs) as img_file:
  File "C:\Users\Sebastian\Documents\Coding-Projects\imageio\imageio\core\imopen.py", line 179, in imopen
    raise err_type(err_msg) from err_from
OSError: An unknown error occured while initializing `<class 'dict'>`
```

We could also place the imposter plugin's stack-trace below our stack trace (just `raise` no `from`), but that seems even more confusing to me.

Regarding the error message itself, we could check for `isinstance(plugin, str)` again and customize the error message to something like `f"The provided plugin does not appear to be a plugin. Got: {plugin}. "`.

almarklein: Aaah! If only I had read 3 more lines downs :)  Sorry about the noise :blush:

FirefoxMetzger: No worries, I can relate.

Do you think we should change the current exception if the plugin is not a string (i.e, if it is an imposter plugin), or is it fine as is?

almarklein: Yeah, maybe change that error message to mention "pugin", e.g. "Could not initialize plugin`{plugin}`"

FirefoxMetzger: Done. it is now

```
OSError: An unknown error occured while initializing plugin `<class 'dict'>`.
```

FirefoxMetzger:
@almarklein I updated the error message. Do you have any other suggestions, or should we merge?

almarklein:
Merged!

PR #693, opened 2021-11-21
BUG: lazy-import plugin namespace

### LINKED ISSUE #690: AttributeError: module 'imageio.plugins' has no attribute 'freeimage'
Hello,

I have started to get [CI failures](https://github.com/colour-science/colour/runs/4270969224?check_suite_focus=true#step:8:177) today, most likely due to some changes introduced by *imageio 2.11.0*.

I usually install the Freeimage plugin as follows:

```bash
poetry run python -c "import imageio;imageio.plugins.freeimage.download()"
```

Which now raises an exception:

```python
AttributeError: module 'imageio.plugins' has no attribute 'freeimage'
```

FirefoxMetzger:
Closes #690 

This PR reformats `imageio.plugins` so that it, too, can lazy-import plugins. This fixes the bug/regression reported in #690 and also deals with any other cases where people may have been relying on syntax of the form

```python
import imageio

imageio.plugins.<...>
```

---

@almarklein This PR might warrant some discussion. Essentially, we replace the entire `imageio.plugins` module with a class `plugins` class that is defined in said module during import. This relies on a semi-official trick (sanctioned by Guido, but not part of the spec), which we may or may not want to rely on. (see: https://stackoverflow.com/questions/2447353/getattr-on-a-module)

The main point of discussion is if we want to keep this behavior moving into v3, or if we should flag it as deprecated, ask people to use an explicit plugin import (e.g., `import imageio.plugins.freeimage`), and remove it in v3.

### DIFF
--- a/imageio/plugins/__init__.py
+++ b/imageio/plugins/__init__.py
@@ -3,87 +3,128 @@
 
 # flake8: noqa
 
-"""
-
-Imagio is plugin-based. Every supported format is provided with a
-plugin. You can write your own plugins to make imageio support
-additional formats. And we would be interested in adding such code to the
-imageio codebase!
-
-
-What is a plugin
-----------------
-
-In imageio, a plugin provides one or more :class:`.Format` objects, and
-corresponding :class:`.Reader` and :class:`.Writer` classes.
-Each Format object represents an implementation to read/write a
-particular file format. Its Reader and Writer classes do the actual
-reading/saving.
-
-The reader and writer objects have a ``request`` attribute that can be
-used to obtain information about the read or write :class:`.Request`, such as
-user-provided keyword arguments, as well get access to the raw image
-data.
-
-
-Registering
------------
-
-Strictly speaking a format can be used stand alone. However, to allow
-imageio to automatically select it for a specific file, the format must
-be registered using ``imageio.formats.add_format()``.
-
-Note that a plugin is not required to be part of the imageio package; as
-long as a format is registered, imageio can use it. This makes imageio very
-easy to extend.
-
-
-What methods to implement
---------------------------
-
-Imageio is designed such that plugins only need to implement a few
-private methods. The public API is implemented by the base classes.
-In effect, the public methods can be given a descent docstring which
-does not have to be repeated at the plugins.
-
-For the Format class, the following needs to be implemented/specified:
-
-  * The format needs a short name, a description, and a list of file
-    extensions that are common for the file-format in question.
-    These ase set when instantiation the Format object.
-  * Use a docstring to provide more detailed information about the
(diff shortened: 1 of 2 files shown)


almarklein:
Do we need a class for this? I thought that `__getattr__` was supported on modules now too? Or maybe only in recent versions?

Anyway, I think I prefer `from imageio.plugins import freeimage` for those that need it. People who use that can be concidered power-users anyway. Trying to hide the fact that we're lazily importing seems to add quite a bit of magic+complexity for small gain. my 2ct :)

almarklein:
Looks like my proposal would upset at least [some people](https://github.com/imageio/imageio/issues/690#issuecomment-975851258) :)  -  for the sake of backwards compatibility, let's use the auto-import magic :)

FirefoxMetzger:
> Do we need a class for this? I thought that __getattr__ was supported on modules now too? Or maybe only in recent versions?

As far as I understand, this behavior got removed at some point and has now been reintroduced in python 3.7+. Since we still support python 3.6 I chose to go with the `class` variant, but we can refactor once we drop py3.6 support :)

> Looks like my proposal would upset at least some people :) - and I guess he's right. For the sake of backwards compatibility, let's use the auto-import magic :)

Yes, for the current API we shouldn't break anything and this PR brings back what we lost. 

My question was moreso how we should handle this moving forward, i.e., do we want to officially deprecate things like

```
import imageio; imageio.plugins.<some_plugin>.<some_func>
```

In this case, we should start raising a deprecation warning and ask users to import explicitly. Alternatively, we can keep this behavior under the argument that `imageio.plugins` lazy-import is no more or less magical than `imopen` automatically loading/importing backends.

almarklein:
Recently [this skimage PR](https://github.com/scikit-image/scikit-image/pull/5101) was brought to my attention, where they do something similar. Looks like they dropped 3.7, a snippet from their new `__init__.py`:
```py
from ._shared.version_requirements import ensure_python_version
ensure_python_version((3, 7))

from ._shared import lazy
__getattr__, __lazy_dir__, _ = lazy.attach(__name__, submodules)

almarklein:
> Alternatively, we can keep this behavior under the argument that imageio.plugins lazy-import is no more or less magical than imopen automatically loading/importing backends.

Practicality beats purity. I suspect maintaining this lazy-import logic is less work then doing a deprecation scheme that we need to track. Lets just try this, and if there are no negative effects we keep it.

FirefoxMetzger:
> I suspect maintaining this lazy-import logic is less work then doing a deprecation scheme that we need to track. Lets just try this, and if there are no negative effects we keep it.

Sounds good to me. Then we do it this way.

@almarklein RTR once the last tests have passed. If you have no other suggestions, feel free to merge :)

PR #695, opened 2021-11-25
Fix fuzzer build

DavidKorczynski:
Fixes fuzzer coverage instrumentation following upgrades in the Atheris fuzzer

### DIFF
--- a/tests/fuzzers/fuzz_read.py
+++ b/tests/fuzzers/fuzz_read.py
@@ -19,7 +19,8 @@ def TestOneInput(data):
 
 
 def main():
-    atheris.Setup(sys.argv, TestOneInput, enable_python_coverage=True)
+    atheris.instrument_all()
+    atheris.Setup(sys.argv, TestOneInput)
     atheris.Fuzz()
 
 


FirefoxMetzger:
Thanks, @DavidKorczynski for updating the fuzzer tests.

I think fuzzers got added before I got actively involved in ImageIO and I actually only have high-level knowledge of how they work. Do you have a good resource for me to read up on them and learn how to best apply them here?

DavidKorczynski:
Starting here is probably a good idea: https://github.com/google/atheris

Then, move onto OSS-Fuzz as that's what's managing the imageio fuzzer: https://github.com/google/oss-fuzz

FirefoxMetzger:
Thanks for the links 🚀  I think I have a good enough understanding now to be able to play around with atheris and fuzzing in general.

@DavidKorczynski I saw that there is a dedicated issue tracker for oss-fuzz discoveries. Where do I need to make some noise to be added to the CC list for ImageIO? I saw that @almarklein is there together with one email I don't recognize, and I think it makes sense for me to be on that list, too.

DavidKorczynski:
You can add the noise here and I'll add your email in this PR to the project's configuration `project.yaml` https://github.com/google/oss-fuzz/pull/6906

So, if you let me know which email (it must be linked to a Google account to log into the platform) then I'll fix it

PR #699, opened 2021-12-01
Issue #391: default webcam fps

dennisvang:
- add condition to ensure default fps for webcam is *only* applied on macOS
- add note about supported webcam framerates in ffmpeg docstring

### DIFF
--- a/imageio/plugins/ffmpeg.py
+++ b/imageio/plugins/ffmpeg.py
@@ -22,7 +22,8 @@ fps : scalar
     The number of frames per second to read the data at. Default None (i.e.
     read at the file's own fps). One can use this for files with a
     variable fps, or in cases where imageio is unable to correctly detect
-    the fps.
+    the fps. In case of trouble opening camera streams, it may help to set an
+    explicit fps value matching a framerate supported by the camera.
 loop : bool
     If True, the video will rewind as soon as a frame is requested
     beyond the last frame. Otherwise, IndexError is raised. Default False.
@@ -122,6 +123,7 @@ import re
 import sys
 import time
 import logging
+import platform
 import threading
 import subprocess as sp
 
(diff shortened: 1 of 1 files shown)


imageio/plugins/ffmpeg.py
```diff
@@ -293,8 +295,8 @@ def _open(
             regex_match = re.match(r"<video(\d+)>", self.request.filename)
             if regex_match:
                 self.request._video = self.request.filename
-            # Specify input framerate?
-            if self.request._video:
+            # Specify input framerate? (only on macOS)
+            if self.request._video and platform.system().lower() == "darwin":
                 if "-framerate" not in str(self._arg_input_params):
                     self._arg_input_params.extend(["-framerate", str(float(fps or 30))])
```
almarklein, line 301: Let's set the framerate to 15 in this case, which seems better supported on MacOS. See  #525 and the snippet you posted.

almarklein:
This looks good. I wonder whether in the mean time, not specifying the framerate works on MacOS too. I cannot test this myself right now thought.

almarklein:
This PR relates to #391, #510 and #525

imageio/plugins/ffmpeg.py
dennisvang: Sounds good. I'll get right on it.

dennisvang:
> This looks good. I wonder whether in the mean time, not specifying the framerate works on MacOS too. I cannot test this myself right now thought.

Can't test this myself either. Don't have any apples.

almarklein (approved):

FirefoxMetzger:
My girlfriend has a Mac ... I will see if I can give this a shot during the weekend.

@almarklein I just realized that our PR templates don't work, so we don't get reminders for prefixing/tagging squash-commit messages with `FEAT:` (new feature) or `BUG:` (bugfix) so that the CD can pick it up and know what to do with the commit. There are other tags, too, which would be neatly documented in the PR template:
https://github.com/imageio/imageio/blob/master/.github/PULL_REQUEST_TEMPLATE/default.md

buuut ... that doesn't seem to be working right now. I'll add investigating that to the TODO list.

Issue #703, opened 2021-12-06
`tiff` can not handle the given uri "<bytes>"


Qwlouse:
Starting from version `2.11.0`  I get an error when trying to write a `tiff` image to a bytes object. 

The following code reproduces the error (up to version `2.10.5` it works fine):
``` python
import imageio 
imageio.imwrite("<bytes>", [[0]], format="tiff")
```

The error:

```
---------------------------------------------------------------------------
InitializationError                       Traceback (most recent call last)
.../python3.7/site-packages/imageio/core/imopen.py in imopen(uri, io_mode, plugin, legacy_mode, **kwargs)
    148         try:
--> 149             return config.plugin_class(request, **kwargs)
    150         except InitializationError as class_specific:

.../python3.7/site-packages/imageio/config/plugins.py in partial_legacy_plugin(request)
    107             def partial_legacy_plugin(request):
--> 108                 return LegacyPlugin(request, legacy_plugin)
    109 

.../python3.7/site-packages/imageio/core/legacy_plugin_wrapper.py in __init__(self, request, legacy_plugin)
     68                 raise InitializationError(
---> 69                     f"`{self._format.name}`"
     70                     f" can not write to `{self._request.raw_uri}`."

InitializationError: `TIFF` can not write to `<bytes>`.

The above exception was the direct cause of the following exception:

RuntimeError                              Traceback (most recent call last)
<ipython-input-5-f8f0708e8d2d> in <module>
----> 1 imageio.imwrite("<bytes>", [[0]], format="tiff")

.../python3.7/site-packages/imageio/core/functions.py in imwrite(uri, im, format, **kwargs)
    193         raise ValueError("Image must be 2D (grayscale, RGB, or RGBA).")
    194 
--> 195     with imopen(uri, "wi", plugin=format) as file:
    196         return file.write(im, **kwargs)
    197 

.../python3.7/site-packages/imageio/core/imopen.py in imopen(uri, io_mode, plugin, legacy_mode, **kwargs)
    165 
    166         request.finish()
--> 167         raise err_type(err_msg) from err_from
    168 
    169     # fast-path based on file extension

RuntimeError: `tiff` can not handle the given uri.
```

FirefoxMetzger:
Thanks for the report @Qwlouse . I tried it locally and could reproduce it. #704 should fix this and added your example as a regression test :)

PR #704, opened 2021-12-07
BUG: allow TIFF to write bytes and file objects

FirefoxMetzger:
Closes: https://github.com/imageio/imageio/issues/703

This PR makes `Tifffile._can_write` return `True` for bytes and FileObjects.


FirefoxMetzger:
@almarklein RTR 🚀  If you are happy with the changes then you can hit the big green button :)

almarklein (approved):

PR #705, opened 2021-12-07
Use pytest fixture and markers for a simpler test control (closes #700)

### LINKED ISSUE #700: Simplifying pytest markers/skip conditions for "no internet"
I'm trying to enable test runs in the conda-forge builds of this package, but I'm finding it hard to get the implementation right.

One of the reasons for this is that this package implements a way to check for internet access which is a bit unusual for a framework based on pytest.  As far as I could understand, if the value of the environment variable `IMAGEIO_NO_INTERNET` assumes a true value, then certain tests should be skipped.  This in turn is used inside `imageio/testing.py` to issue a `pytest.skip("No internet")` within a function called `need_internet()`, which is in turned used within each test inside `tests/`, which would require internet access otherwise.

Would it be possible to consider a simplification of this testing in which, instead of this system, you mark the tests that require internet with something like:

```python
@pytest.mark.requires_internet
def test_bla():
    ...
```

Then, you select tests with or without internet access with:

```sh
$ pytest -m "requires_internet" ... # all tests that require internet
$ pytest -m "not requires_internet" ... # all tests that do not require internet

```

anjos:
Closes: https://github.com/imageio/imageio/issues/700

This PR includes the following changes:

* replace the use of `need_internet()` by a marker, called `needs_internet`.  All tests that require internet access can now be easily skipped with `pytest -m 'not needs_internet'`.  Please note that, during testing, the environment variable `IMAGEIO_NO_INTERNET` is not required anymore, but that is still used by `imageio.core.fetching`.  I'm not sure it is still relevant.
* make `get_test_dir()/clean_test_dir()` into a proper fixture, and add it to `tests/conftest.py`, ensure that all test units that require that to be setup explicitly invoke that fixture
* remove `run_tests_if_main()` from `imageio.testing`.  You may call `pytest` by hand to achieve the same effect.  Default options may go into a `pyproject.toml` if they wish to be preserved (however I did not do this in this PR)
* cleanup unused modules in some tests
* use `pytest.importorskip` to skip modules that require specific python libraries to be installed.  This is the case for osgeo/gdal, astropy, tifffile and imageio-ffmpeg.
* only execute `tests/test_ffmpeg.py` if psutil is also installed.
* replace global setup functions by equivalent test fixtures (e.g. `tests/test_dicom.py`)
* pass Black through out all touched files
* no tests require `imageio.testing` anymore, however, I left the file there since some of the CD/CI pipelines use this via invoke?  (I'm not sure of this, however, I think it would be cleaner if the `pytest` command-line appeared directly on these YAML files, so it serves as an example of how the package should be tested.

To run tests, you can simply do:

```sh
$ pytest -sv tests/
```

Tests that need running should be automatically detected, except the internet access.

To see what is skipped (and the reason) in a more explicit way, do:

```sh
$ pytest -sv -rs tests/
```

To suppress all tests that require internet access, just do:

```sh
$ pytest -sv -m 'not needs_internet' tests/
```

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -119,9 +119,7 @@ def pytest_collection_modifyitems(items):
         if ("image_cache" in getattr(item, "fixturenames", ())) or (
             "image_copy" in getattr(item, "fixturenames", ())
         ):
-            if item.config.getoption("imageio_binaries").startswith(
-                internet_protocols
-            ):
+            if item.config.getoption("imageio_binaries").startswith(internet_protocols):
                 item.add_marker("needs_internet")
 
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -26,7 +26,7 @@ def pytest_addoption(parser):
         action="store",
         metavar="GIT_REPO",
         default="https://github.com/imageio/imageio-binaries.git",
-        help="Repository to use for checking out binary data used at tests"
+        help="Repository to use for checking out binary data used for tests"
         " (default: %(default)s).  Use 'file:///path/to/imageio-binaries'"
         " to clone from a local repository and save testing bandwidth.",
     )
@@ -156,23 +156,21 @@ def invalid_file(tmp_path, request):
 def tmp_userdir(tmp_path):
 
     ud = tmp_path / "userdir"
-    os.makedirs(ud, exist_ok=True)
+    ud.mkdir(exist_ok=True)
 
     if sys.platform.startswith("win"):
-        if "LOCALAPPDATA" in os.environ:  # saves it
-            os.environ["OLD_LOCALAPPDATA"] = os.environ["LOCALAPPDATA"]
-        os.environ["LOCALAPPDATA"] = str(ud)
+        user_dir_env = "LOCALAPPDATA"
     else:
-        os.environ["IMAGEIO_USERDIR"] = str(ud)
+        user_dir_env = "IMAGEIO_USERDIR"
+    
+    old_user_dir = os.getenv(user_dir_env, None)
+    os.environ[user_dir_env] = str(ud)
 
     yield ud
 
-    if sys.platform.startswith("win"):
-        del os.environ["LOCALAPPDATA"]
-        if "OLD_LOCALAPPDATA" in os.environ:
-            os.environ["LOCALAPPDATA"] = os.environ["OLD_LOCALAPPDATA"]
-            del os.environ["OLD_LOCALAPPDATA"]
+    if old_user_dir is not None:
+        os.environ[user_dir_env] = old_user_dir
     else:
-        del os.environ["IMAGEIO_USERDIR"]
+        del os.environ[user_dir_env]
 
     shutil.rmtree(ud)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,10 +1,7 @@
 import os
 import sys
-import stat
 import shutil
 from pathlib import Path
-import contextlib
-import subprocess
 import contextlib
 
 import pytest
@@ -114,11 +111,11 @@ def image_files(test_images, tmp_path):
 
 def pytest_collection_modifyitems(items):
 
-    # all tests using the image_cache fixture, require internet automatically
+    # all tests using the test_images fixture, require internet automatically
     # iff they checkout from the remote repository - everything else does not
     internet_protocols = ("git", "http", "ssh")
     for item in items:
-        if ("image_cache" in getattr(item, "fixturenames", ())) or (
+        if ("test_images" in getattr(item, "fixturenames", ())) or (
             "image_copy" in getattr(item, "fixturenames", ())
         ):
             if item.config.getoption("imageio_binaries").startswith(internet_protocols):
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -162,7 +162,7 @@ def tmp_userdir(tmp_path):
         user_dir_env = "LOCALAPPDATA"
     else:
         user_dir_env = "IMAGEIO_USERDIR"
-    
+
     old_user_dir = os.getenv(user_dir_env, None)
     os.environ[user_dir_env] = str(ud)
 
@@ -172,5 +172,3 @@ def tmp_userdir(tmp_path):
         os.environ[user_dir_env] = old_user_dir
     else:
         del os.environ[user_dir_env]
-
-    shutil.rmtree(ud)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,8 +1,12 @@
-import pytest
 import os
+import pathlib
 import shutil
 from pathlib import Path
 import contextlib
+import subprocess
+import contextlib
+
+import pytest
 
 import imageio as iio
 
@@ -120,3 +124,12 @@ def invalid_file(tmp_path, request):
         file.write("Actually not a file.")
 
     return tmp_path / ("foo" + ext)
+
+
+@pytest.fixture()
+def tmp_userdir(tmp_path):
+    ud = tmp_path / "userdir"
+    os.makedirs(ud, exist_ok=True)
+    os.environ["IMAGEIO_USERDIR"] = str(ud)
+    yield ud
+    shutil.rmtree(ud)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -19,12 +19,12 @@ def pytest_configure(config):
 
 def pytest_addoption(parser):
     parser.addoption(
-        "--imageio-binaries",
+        "--test-images",
         action="store",
         metavar="GIT_REPO",
-        default="https://github.com/imageio/imageio-binaries.git",
-        help="Repository to use for checking out binary data used for tests"
-        " (default: %(default)s).  Use 'file:///path/to/imageio-binaries'"
+        default="https://github.com/imageio/test_images.git",
+        help="Repository containing the test images "
+        " (default: %(default)s).  Use `file:///path/to/test_images`"
         " to clone from a local repository and save testing bandwidth.",
     )
 
@@ -82,7 +82,8 @@ def test_images(request):
             pass  # dir exist, validate and fail later
         else:
             with working_directory(checkout_dir):
-                os.system("git clone https://github.com/imageio/test_images.git .")
+                repo_location = request.config.getoption("--test-images")
+                os.system(f"git clone {repo_location} .")
 
         request.config.cache.set("imageio_test_binaries", str(checkout_dir))
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -10,8 +10,7 @@ import imageio as iio
 def pytest_configure(config):
     config.addinivalue_line(
         "markers",
-        "needs_internet: mark tests that require internet access "
-        "(deselect with '-m \"not needs_internet\"'",
+        "(deselect with '-m \"not needs_internet\"').",
     )
 
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -159,7 +159,7 @@ def tmp_userdir(tmp_path):
     os.makedirs(ud, exist_ok=True)
 
     if sys.platform.startswith("win"):
-        if "LOCALAPPDATA" in os.environ:  #saves it
+        if "LOCALAPPDATA" in os.environ:  # saves it
             os.environ["OLD_LOCALAPPDATA"] = os.environ["LOCALAPPDATA"]
         os.environ["LOCALAPPDATA"] = str(ud)
     else:
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -82,7 +82,7 @@ def test_images(request):
     return checkout_dir
 
 
-@pytest.fixture
+@pytest.fixture()
 def image_files(test_images, tmp_path):
     """A copy of the test images
 
@@ -95,7 +95,7 @@ def image_files(test_images, tmp_path):
     yield tmp_path
 
 
-@pytest.fixture
+@pytest.fixture()
 def clear_plugins():
 
     old_extensions = iio.config.known_extensions.copy()
@@ -122,7 +122,7 @@ def invalid_file(tmp_path, request):
     return tmp_path / ("foo" + ext)
 
 
-@pytest.fixture(scope="module")
+@pytest.fixture()
 def test_dir():
     # Define dir
     from imageio.core import appdata_dir
@@ -130,5 +130,7 @@ def test_dir():
     d = os.path.join(appdata_dir("imageio"), "testdir")
     os.makedirs(d)
     os.makedirs(os.path.join(d, "images"))
+
     yield d
+
     shutil.rmtree(d)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,4 +1,5 @@
 import os
+import stat
 import shutil
 from pathlib import Path
 import contextlib
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -153,8 +153,12 @@ def invalid_file(tmp_path, request):
 
 @pytest.fixture()
 def tmp_userdir(tmp_path):
+
     ud = tmp_path / "userdir"
     os.makedirs(ud, exist_ok=True)
     os.environ["IMAGEIO_USERDIR"] = str(ud)
+
     yield ud
+
+    del os.environ["IMAGEIO_USERDIR"]
     shutil.rmtree(ud)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -120,17 +120,3 @@ def invalid_file(tmp_path, request):
         file.write("Actually not a file.")
 
     return tmp_path / ("foo" + ext)
-
-
-@pytest.fixture()
-def test_dir():
-    # Define dir
-    from imageio.core import appdata_dir
-
-    d = os.path.join(appdata_dir("imageio"), "testdir")
-    os.makedirs(d)
-    os.makedirs(os.path.join(d, "images"))
-
-    yield d
-
-    shutil.rmtree(d)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,5 +1,4 @@
 import os
-import pathlib
 import shutil
 from pathlib import Path
 import contextlib
@@ -16,6 +15,18 @@ def pytest_configure(config):
         "markers",
         "needs_internet: Marks a test that requires an active interent connection."
         " (deselect with '-m \"not needs_internet\"').",
+    )
+
+
+def pytest_addoption(parser):
+    parser.addoption(
+        "--imageio-binaries",
+        action="store",
+        metavar="GIT_REPO",
+        default="https://github.com/imageio/imageio-binaries.git",
+        help="Repository to use for checking out binary data used at tests"
+        " (default: %(default)s).  Use 'file:///path/to/imageio-binaries'"
+        " to clone from a local repository and save testing bandwidth.",
     )
 
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -28,6 +28,20 @@ def pytest_addoption(parser):
         " to clone from a local repository and save testing bandwidth.",
     )
 
+    parser.addoption(
+        "--github-token",
+        action="store",
+        default=None,
+        help="If set, this value will be used as a token to authenticate with the GitHub API.",
+    )
+
+    parser.addoption(
+        "--github-username",
+        action="store",
+        default=None,
+        help="If set, this value will be used as the username to authenticate with the GitHub API.",
+    )
+
 
 @contextlib.contextmanager
 def working_directory(path):
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -110,6 +110,21 @@ def image_files(test_images, tmp_path):
     yield tmp_path
 
 
+def pytest_collection_modifyitems(items):
+
+    # all tests using the image_cache fixture, require internet automatically
+    # iff they checkout from the remote repository - everything else does not
+    internet_protocols = ("git", "http", "ssh")
+    for item in items:
+        if ("image_cache" in getattr(item, "fixturenames", ())) or (
+            "image_copy" in getattr(item, "fixturenames", ())
+        ):
+            if item.config.getoption("imageio_binaries").startswith(
+                internet_protocols
+            ):
+                item.add_marker("needs_internet")
+
+
 @pytest.fixture()
 def clear_plugins():
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -51,7 +51,11 @@ def working_directory(path):
         os.chdir(prev_cwd)
 
 
-@pytest.fixture(scope="session")
+# uses the fixture marking workaround from:
+# https://github.com/pytest-dev/pytest/issues/1368#issuecomment-466339463
+@pytest.fixture(
+    scope="session", params=[pytest.param(0, marks=pytest.mark.needs_internet)]
+)
 def test_images(request):
     """A collection of test images.
 
@@ -107,19 +111,6 @@ def image_files(test_images, tmp_path):
             shutil.copy(item, tmp_path / item.name)
 
     yield tmp_path
-
-
-def pytest_collection_modifyitems(items):
-
-    # all tests using the test_images fixture, require internet automatically
-    # iff they checkout from the remote repository - everything else does not
-    internet_protocols = ("git", "http", "ssh")
-    for item in items:
-        if ("test_images" in getattr(item, "fixturenames", ())) or (
-            "image_copy" in getattr(item, "fixturenames", ())
-        ):
-            if item.config.getoption("imageio_binaries").startswith(internet_protocols):
-                item.add_marker("needs_internet")
 
 
 @pytest.fixture()
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -10,7 +10,8 @@ import imageio as iio
 def pytest_configure(config):
     config.addinivalue_line(
         "markers",
-        "(deselect with '-m \"not needs_internet\"').",
+        "needs_internet: Marks a test that requires an active interent connection."
+        " (deselect with '-m \"not needs_internet\"').",
     )
 
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -7,6 +7,14 @@ import contextlib
 import imageio as iio
 
 
+def pytest_configure(config):
+    config.addinivalue_line(
+        "markers",
+        "needs_internet: mark tests that require internet access "
+        "(deselect with '-m \"not needs_internet\"'",
+    )
+
+
 @contextlib.contextmanager
 def working_directory(path):
     """
@@ -112,3 +120,15 @@ def invalid_file(tmp_path, request):
         file.write("Actually not a file.")
 
     return tmp_path / ("foo" + ext)
+
+
+@pytest.fixture(scope="module")
+def test_dir():
+    # Define dir
+    from imageio.core import appdata_dir
+
+    d = os.path.join(appdata_dir("imageio"), "testdir")
+    os.makedirs(d)
+    os.makedirs(os.path.join(d, "images"))
+    yield d
+    shutil.rmtree(d)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,4 +1,5 @@
 import os
+import sys
 import stat
 import shutil
 from pathlib import Path
@@ -156,9 +157,22 @@ def tmp_userdir(tmp_path):
 
     ud = tmp_path / "userdir"
     os.makedirs(ud, exist_ok=True)
-    os.environ["IMAGEIO_USERDIR"] = str(ud)
+
+    if sys.platform.startswith("win"):
+        if "LOCALAPPDATA" in os.environ:  #saves it
+            os.environ["OLD_LOCALAPPDATA"] = os.environ["LOCALAPPDATA"]
+        os.environ["LOCALAPPDATA"] = str(ud)
+    else:
+        os.environ["IMAGEIO_USERDIR"] = str(ud)
 
     yield ud
 
-    del os.environ["IMAGEIO_USERDIR"]
+    if sys.platform.startswith("win"):
+        del os.environ["LOCALAPPDATA"]
+        if "OLD_LOCALAPPDATA" in os.environ:
+            os.environ["LOCALAPPDATA"] = os.environ["OLD_LOCALAPPDATA"]
+            del os.environ["OLD_LOCALAPPDATA"]
+    else:
+        del os.environ["IMAGEIO_USERDIR"]
+
     shutil.rmtree(ud)
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -24,7 +24,7 @@ import imageio as iio
 import imageio.core.imopen as imopen_module
 from imageio import core
 from imageio.core import Request
-from imageio.core import get_remote_file, IS_PYPY
+from imageio.core import IS_PYPY, get_remote_file
 from imageio.core.request import Mode, InitializationError
 from imageio.config.plugins import PluginConfig
 
@@ -55,25 +55,23 @@ class BrokenDummyPlugin:
 
 
 @pytest.mark.needs_internet
-def test_fetching(tmp_path):
+def test_fetching(tmp_userdir):
     """Test fetching of files"""
 
-    # Clear image files
-
     # This should download the file (force download, because local cache)
-    fname1 = get_remote_file("images/chelsea.png", tmp_path, True)
+    fname1 = get_remote_file("images/chelsea.png", tmp_userdir, True)
     mtime1 = os.path.getmtime(fname1)
     # This should reuse it
-    fname2 = get_remote_file("images/chelsea.png", tmp_path)
+    fname2 = get_remote_file("images/chelsea.png", tmp_userdir)
     mtime2 = os.path.getmtime(fname2)
     # This should overwrite
-    fname3 = get_remote_file("images/chelsea.png", tmp_path, True)
+    fname3 = get_remote_file("images/chelsea.png", tmp_userdir, True)
     mtime3 = os.path.getmtime(fname3)
     # This should too (update this if imageio is still around in 1000 years)
-    fname4 = get_remote_file("images/chelsea.png", tmp_path, "3014-01-01")
+    fname4 = get_remote_file("images/chelsea.png", tmp_userdir, "3014-01-01")
     mtime4 = os.path.getmtime(fname4)
     # This should not
-    fname5 = get_remote_file("images/chelsea.png", tmp_path, "2014-01-01")
+    fname5 = get_remote_file("images/chelsea.png", tmp_userdir, "2014-01-01")
     mtime5 = os.path.getmtime(fname4)
     #
     assert os.path.isfile(fname1)
@@ -93,7 +91,7 @@ def test_fetching(tmp_path):
     _chunk_read = core.fetching._chunk_read
     #
     with pytest.raises(IOError):
-        get_remote_file("this_does_not_exist", tmp_path)
+        get_remote_file("this_does_not_exist", tmp_userdir)
     #
     try:
         core.fetching.urlopen = None
@@ -116,21 +114,19 @@ def test_fetching(tmp_path):
     assert "0 bytes" == core.fetching._sizeof_fmt(0)
 
 
+@pytest.mark.skip(reason="always skip, is tested implicitly anyway")
 def test_findlib1():
 
     # Lib name would need to be "libc.so.5", or "libc.so.6", or ...
     # Meh, just skip
-    skip("always skip, is tested implicitly anyway")
-
     if not sys.platform.startswith("linux"):
         skip("test on linux only")
 
     # Candidate libs for common lib (note, this runs only on linux)
-    dirs, paths = core.findlib.generate_candidate_libs(["libc"])
+    _, paths = core.findlib.generate_candidate_libs(["libc"])
     assert paths
 
 
-@pytest.mark.needs_internet
 def test_findlib2():
 
     if not sys.platform.startswith("linux"):
@@ -140,7 +136,9 @@ def test_findlib2():
     fi_dir = os.path.join(core.appdata_dir("imageio"), "freeimage")
     if not os.path.isdir(fi_dir):
         os.mkdir(fi_dir)
-    dirs, paths = core.findlib.generate_candidate_libs(["libfreeimage"], [fi_dir])
+    dirs, paths = core.findlib.generate_candidate_libs(
+        ["libfreeimage"], [fi_dir]
+    )
     # assert fi_dir in dirs -> Cannot test: lib may not exist
 
     open(os.path.join(fi_dir, "notalib.test.so"), "wb")
@@ -156,7 +154,7 @@ def test_findlib2():
 
 
 @pytest.mark.needs_internet
-def test_request(tmp_path):
+def test_request(image_cache, tmp_userdir):
     """Test request object"""
 
     # Check uri-type, this is not a public property, so we test the private
@@ -171,7 +169,7 @@ def test_request(tmp_path):
     R = Request(imageio.RETURN_BYTES, "wi")
     assert R._uri_type == core.request.URI_BYTES
     #
-    fname = get_remote_file("images/chelsea.png", tmp_path)
+    fname = image_cache / "images" / "chelsea.png"
     R = Request(fname, "ri")
     assert R._uri_type == core.request.URI_FILENAME
     R = Request("~/filethatdoesnotexist", "wi")
@@ -238,19 +236,17 @@ def test_request(tmp_path):
 
 
 @pytest.mark.needs_internet
-def test_request_read_sources(tmp_path):
+def test_request_read_sources(image_cache, tmp_userdir):
 
     # Make an image available in many ways
-    fname = "images/chelsea.png"
-    filename = get_remote_file(fname, tmp_path)
-    bytes = open(filename, "rb").read()
+    fname = os.path.join("images", "chelsea.png")
+    filename = image_cache / fname
+    bytes = open(str(filename), "rb").read()
     #
     burl = "https://raw.githubusercontent.com/imageio/imageio-binaries/master/"
-    zipfilename = tmp_path / "test1.zip"
+    zipfilename = tmp_userdir / "test1.zip"
     with ZipFile(zipfilename, "w") as zf:
         zf.writestr(fname, bytes)
-
-    has_inet = os.getenv("IMAGEIO_NO_INTERNET", "") not in ("1", "yes", "true")
 
     # Read that image from these different sources. Read data from file
     # and from local file (the two main plugin-facing functions)
@@ -265,8 +261,7 @@ def test_request_read_sources(tmp_path):
                 memoryview(bytes),
                 open(filename, "rb"),
             ]
-            if has_inet:
-                uris.append(burl + fname)
+            uris.append(burl + fname)
 
             for uri in uris:
                 R = Request(uri, "ri")
@@ -284,18 +279,16 @@ def test_request_read_sources(tmp_path):
                     assert all_bytes.startswith(first_bytes)
 
 
-@pytest.mark.needs_internet
-def test_request_save_sources(tmp_path):
+def test_request_save_sources(image_cache, tmp_path):
 
     # Get test data
-    fname = "images/chelsea.png"
-    filename = get_remote_file(fname, tmp_path)
+    filename = image_cache / "test-images" / "chelsea.png"
     with open(filename, "rb") as f:
         bytes = f.read()
     assert len(bytes) > 0
 
     # Prepare destinations
-    fname2 = fname + ".out"
+    fname2 = filename.with_suffix(".out").name
     filename2 = tmp_path / fname2
     zipfilename2 = tmp_path / "test2.zip"
     file2 = None
@@ -614,17 +607,16 @@ def test_util_has_has_module():
     assert core.has_module("sys")
 
 
-@pytest.mark.needs_internet
-def test_functions(tmp_path):
+def test_functions(image_cache, tmp_path):
     """Test the user-facing API functions"""
 
     # Test help(), it prints stuff, so we just check whether that goes ok
     imageio.help()  # should print overview
     imageio.help("PNG")  # should print about PNG
 
-    fname1 = get_remote_file("images/chelsea.png", tmp_path)
-    fname2 = fname1[:-3] + "jpg"
-    fname3 = fname1[:-3] + "notavalidext"
+    fname1 = image_cache / "test-images" / "chelsea.png"
+    fname2 = tmp_path / fname1.with_suffix(".jpg").name
+    fname3 = tmp_path / fname1.with_suffix(".notavalidext").name
     open(fname3, "wb")
 
     # Test read()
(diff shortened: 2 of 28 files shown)


FirefoxMetzger:
Thanks for the PR :) I am letting CI run, but I won't have time for a review today. I hope to get around to it on Thursday :)

anjos:
I just pushed some fixes to the workflow.  Please consider accepting it to run before your review.

anjos:
The above patch only fixes the linting phase.  I'll have to find a Windows machine to fix the Windows tests.

anjos:
@FirefoxMetzger: I think I have fixed all remaining problems.  Can you please allow the run of the CI suite?

anjos:
I've fixed some of the linting errors with the latest commit (everything else is green).  However, I'm doubtful of the following linting errors:

1. `./tests/test_fits.py:15:5: F401 'astropy.io.fits' imported but unused` 
2. `./tests/test_gdal.py:14:5: F401 'osgeo.gdal' imported but unused`

Is the import required so that the plugin works?  I can I just suppress those imports?

FirefoxMetzger:
(Disclaimer: this is without looking at any code.) These are the backends that our gdal and fits plugins are using. It's a bit odd that we don't make use of them in their respective unit tests.

@anjos You can comment them out for now (since flake says they are unused) and I will check what looks like a reasonable approach for them once I get around to doing a full review.

anjos:
@FirefoxMetzger: very well, latest patches pushed to remove the remainder of the linting issues.  Please allow the build to re-run.

tests/test_freeimage.py
```diff
@@ -413,7 +428,10 @@ def test_gif():
     )
 
 
-def test_animated_gif():
+@pytest.mark.needs_internet
+def test_animated_gif(tmp_path, get_library):
+
+    fnamebase = os.path.join(tmp_path, "test")
```
FirefoxMetzger, line 434:
```suggestion
def test_animated_gif(tmp_path):

    fnamebase = tmp_path / "test"
```

tests/test_freeimage.py
```diff
@@ -537,12 +562,17 @@ def test_multi_icon_ico():
...
+@pytest.mark.needs_internet
+def test_mng(get_library):
+    imageio.imread(get_remote_file("images/mngexample.mng"))
 
 
-def test_pnm():
+@pytest.mark.needs_internet
+def test_pnm(tmp_path, get_library):
+
+    fnamebase = os.path.join(tmp_path, "test")
```
FirefoxMetzger, line 574:
```suggestion
def test_pnm(tmp_path):

    fnamebase = tmp_path / "test"
```

tests/test_pillow_legacy.py
```diff
@@ -87,17 +85,20 @@ def test_pillow_format():
...
     W.set_meta_data({"foo": 3})
-    assert raises(RuntimeError, W.append_data, im0)
+    assert pytest.raises(RuntimeError, W.append_data, im0)
 
 
-def test_png():
+@pytest.mark.needs_internet
+def test_png(tmp_path):
+
+    fnamebase = os.path.join(tmp_path, "test")
```
FirefoxMetzger, line 101:
```suggestion
    fnamebase = tmp_path / "test"
```

tests/test_pillow_legacy.py
```diff
@@ -162,17 +163,19 @@ def test_png():
...
         "https://raw.githubusercontent.com/imageio/"
         + "imageio-binaries/master/images/astronaut.png"
     )
     assert im.shape == (512, 512, 3)
 
 
-def test_jpg():
+def test_jpg(tmp_path):
+
+    fnamebase = os.path.join(tmp_path, "test")
```
FirefoxMetzger, line 178:
```suggestion
    fnamebase = tmp_path / "test"
```

tests/test_pillow_legacy.py
```diff
@@ -186,41 +189,43 @@ def test_jpg():
...
-    raises(ValueError, imageio.imsave, fnamebase + ".jpg", im, quality=120)
+    pytest.raises(ValueError, imageio.imsave, fnamebase + ".jpg", im, quality=120)
 
 
-def test_jpg_more():
-    need_internet()
+@pytest.mark.needs_internet
+def test_jpg_more(tmp_path):
+
+    fnamebase = os.path.join(tmp_path, "test")
```
FirefoxMetzger, line 215:
```suggestion
    fnamebase = tmp_path / "test"
```

tests/test_pillow_legacy.py
```diff
@@ -186,41 +189,43 @@ def test_jpg():
...
 
     # Test broken JPEG
     fname = fnamebase + "_broken.jpg"
     open(fname, "wb").write(b"this is not an image")
-    raises(Exception, imageio.imread, fname)
+    pytest.raises(Exception, imageio.imread, fname)
```
FirefoxMetzger, line 220:
```suggestion
    with pytest.raises(Exception):
        imageio.imread(fname)
```

tests/test_pillow_legacy.py
```diff
@@ -186,41 +189,43 @@ def test_jpg():
...
     #
     img = get_ref_im(3, 0, 0)
     bb = imageio.imsave(imageio.RETURN_BYTES, img, "JPEG-PIL")
     with open(fname, "wb") as f:
         f.write(bb[:400])
         f.write(b" ")
         f.write(bb[400:])
-    raises(Exception, imageio.imread, fname)
+    pytest.raises(Exception, imageio.imread, fname)
```
FirefoxMetzger, line 228:
```suggestion
    with pytest.raises(Exception):
        imageio.imread(fname)
```

tests/test_pillow_legacy.py
```diff
@@ -238,7 +243,9 @@ def test_jpg_more():
     assert im.meta.EXIF_MAIN
 
 
-def test_gif():
+def test_gif(tmp_path):
+    fnamebase = os.path.join(tmp_path, "test")
```
FirefoxMetzger, line 247:
```suggestion
    fnamebase = tmp_path / "test"
```

tests/test_pillow_legacy.py
```diff
@@ -257,11 +264,14 @@ def test_gif():
                 assert_close(rim * mul, im, 1.1)  # lossless
 
     # Parameter fail
-    raises(TypeError, imageio.imread, fname, notavalidkwarg=True)
-    raises(TypeError, imageio.imsave, fnamebase + "1.gif", im, notavalidk=True)
+    pytest.raises(TypeError, imageio.imread, fname, notavalidkwarg=True)
+    pytest.raises(TypeError, imageio.imsave, fnamebase + "1.gif", im, notavalidk=True)
```
FirefoxMetzger, line 268:
```suggestion
    with pytest.raises(TypeError):
        imageio.imread(fname, notavalidkwarg=True)

    with pytest.raises(TypeError):
        imageio.imsave(fnamebase + "1.gif", im, notavalidk=True)
```

tests/test_pillow_legacy.py
```diff
@@ -257,11 +264,14 @@ def test_gif():
...
+
 
+@pytest.mark.needs_internet
+def test_animated_gif(tmp_path):
 
-def test_animated_gif():
+    fnamebase = os.path.join(tmp_path, "test")
```
FirefoxMetzger, line 274:
```suggestion
    fnamebase = tmp_path / "test"
```

tests/test_pillow_legacy.py
```diff
@@ -369,10 +379,10 @@ def test_gamma_correction():
         assert im.shape == (512, 768, 3) and im.dtype == "uint8"
 
 
-def test_inside_zipfile():
-    need_internet()
+@pytest.mark.needs_internet
+def test_inside_zipfile(tmp_path):
 
-    fname = os.path.join(test_dir, "pillowtest.zip")
+    fname = os.path.join(tmp_path, "pillowtest.zip")
```
FirefoxMetzger, line 385:
```suggestion
    fname = tmp_path / "pillowtest.zip"
```

tests/test_pillow_legacy.py
```diff
@@ -381,15 +391,16 @@ def test_inside_zipfile():
         imageio.imread(fname + "/" + name)
 
 
-def test_bmp():
-    need_internet()
-    fname = get_remote_file("images/scribble_P_RGB.bmp", test_dir)
+@pytest.mark.needs_internet
+def test_bmp(tmp_path):
+    fname = get_remote_file("images/scribble_P_RGB.bmp", tmp_path)
```
FirefoxMetzger, line 396: We got another one! I think at this point 50 of my comments must be on the amazing `test_images` fixture :D

tests/test_simpleitk.py
```diff
@@ -27,11 +23,11 @@
 
 
 @pytest.mark.skipif("itk is None")
-def test_simpleitk_reading_writing():
+def test_simpleitk_reading_writing(tmp_path):
     """Test reading and saveing tiff"""
     im2 = np.ones((10, 10, 3), np.uint8) * 2
 
-    filename1 = os.path.join(test_dir, "test_tiff.tiff")
+    filename1 = os.path.join(tmp_path, "test_tiff.tiff")
```
FirefoxMetzger, line 30:
```suggestion
    filename1 = tmp_path / "test_tiff.tiff"
```

tests/test_swf.py
```diff
@@ -5,30 +5,25 @@
...
 
 
-def test_format_selection():
-
-    need_internet()
+@pytest.mark.needs_internet
+def test_format_selection(tmp_path):
 
-    fname1 = get_remote_file("images/stent.swf", test_dir)
+    fname1 = get_remote_file("images/stent.swf", tmp_path)
```
FirefoxMetzger, line 26: `test_images` fixture anyone?

tests/test_pillow_legacy.py
```diff
@@ -313,10 +323,10 @@ def test_animated_gif():
...
     assert W._writer.opt_palette_size == 128
     # Fail
-    assert raises(IndexError, R.get_meta_data, -1)
-    assert raises(ValueError, imageio.mimsave, fname, ims, palettesize=300)
-    assert raises(ValueError, imageio.mimsave, fname, ims, quantizer="foo")
-    assert raises(ValueError, imageio.mimsave, fname, ims, duration="foo")
+    assert pytest.raises(IndexError, R.get_meta_data, -1)
+    assert pytest.raises(ValueError, imageio.mimsave, fname, ims, palettesize=300)
+    assert pytest.raises(ValueError, imageio.mimsave, fname, ims, quantizer="foo")
+    assert pytest.raises(ValueError, imageio.mimsave, fname, ims, duration="foo")
```
FirefoxMetzger, line 329:
```suggestion
    with pytest.raises(IndexError):
        R.get_meta_data(-1)

    with pytest.raises(ValueError):
        imageio.mimsave(fname, ims, palettesize=300)

    with pytest.raises(ValueError):
        imageio.mimsave(fname, ims, quantizer="foo")

    with pytest.raises(ValueError):
        imageio.mimsave(fname, ims, duration="foo")
```

tests/test_swf.py
```diff
@@ -39,11 +34,10 @@ def test_format_selection():
...
 
 
-def test_reading_saving():
+@pytest.mark.needs_internet
+def test_reading_saving(tmp_path):
 
-    need_internet()
-
-    fname1 = get_remote_file("images/stent.swf", test_dir)
+    fname1 = get_remote_file("images/stent.swf", tmp_path)
```
FirefoxMetzger, line 40: And one last time: `test_images` fixture

tests/test_swf.py
```diff
@@ -60,8 +54,8 @@ def test_reading_saving():
     # Seek
     assert (R.get_data(3) == ims1[3]).all()
     # Fails
-    raises(IndexError, R.get_data, -1)  # No negative index
-    raises(IndexError, R.get_data, 10)  # Out of bounds
+    pytest.raises(IndexError, R.get_data, -1)  # No negative index
+    pytest.raises(IndexError, R.get_data, 10)  # Out of bounds
```
FirefoxMetzger, line 58:
```suggestion
    with pytest.raises(IndexError):
        R.get_data(-1)  # No negative index

    with pytest.raises(IndexError):
        R.get_data(10)  # Out of bounds
```

tests/test_swf.py
```diff
@@ -131,51 +125,48 @@ def test_reading_saving():
         fname4,
     )
 
-    with open(os.path.join(test_dir, "test_swf.html"), "wb") as f:
+    with open(os.path.join(tmp_path, "test_swf.html"), "wb") as f:
```
FirefoxMetzger, line 128:
```suggestion
    with open(tmp_path / "test_swf.html", "wb") as f:
```

tests/test_swf.py
```diff
@@ -131,51 +125,48 @@ def test_reading_saving():
...
 
+@pytest.mark.needs_internet
 def test_read_from_url():
 
-    need_internet()
-
     burl = "https://raw.githubusercontent.com/imageio/imageio-binaries/master/"
     url = burl + "images/stent.swf"
 
     ims = imageio.mimread(url)
```
FirefoxMetzger, line 139: This would also be covered by the `test_images` fixture.

tests/test_swf.py
```diff
@@ -131,51 +125,48 @@ def test_reading_saving():
...
 
 
-def test_invalid():
+@pytest.mark.needs_internet
+def test_invalid(tmp_path):
 
-    need_internet()
-
-    fname1 = get_remote_file("images/stent.swf", test_dir)
+    fname1 = get_remote_file("images/stent.swf", tmp_path)
```
FirefoxMetzger, line 146: `test_images` fixture 🤖

tests/test_swf.py
```diff
@@ -131,51 +125,48 @@ def test_reading_saving():
...
     fname2 = fname1[:-4] + ".invalid.swf"
 
     # Empty file
     with open(fname2, "wb"):
         pass
     assert not imageio.formats.search_read_format(core.Request(fname2, "rI"))
-    raises(RuntimeError, imageio.mimread, fname2, "swf")
+    pytest.raises(RuntimeError, imageio.mimread, fname2, "swf")
```
FirefoxMetzger, line 153:
```suggestion
    with pytest.raises(RuntimeError):
        imageio.mimread(fname2, "swf")
```

tests/test_swf.py
```diff
@@ -131,51 +125,48 @@ def test_reading_saving():
...
 
     # File with BS data
     with open(fname2, "wb") as f:
         f.write(b"x" * 100)
     assert not imageio.formats.search_read_format(core.Request(fname2, "rI"))
-    raises(RuntimeError, imageio.mimread, fname2, "swf")
+    pytest.raises(RuntimeError, imageio.mimread, fname2, "swf")
```
FirefoxMetzger, line 159:
```suggestion
    with pytest.raises(RuntimeError):
        imageio.mimread(fname2, "swf")
```

tests/test_swf.py
```diff
@@ -131,51 +125,48 @@ def test_reading_saving():
...
 def test_lowlevel():
 
-    need_internet()
-
     # Some tests from low level implementation that is not covered
     # by using the plugin itself.
     _swf = imageio.plugins.swf.load_lib()
     tag = _swf.Tag()
-    raises(NotImplementedError, tag.process_tag)
+    pytest.raises(NotImplementedError, tag.process_tag)
```
FirefoxMetzger, line 169:
```suggestion
    with pytest.raises(NotImplementedError):
        tag.process_tag()
```

tests/test_swf.py
```diff
@@ -193,11 +184,10 @@ def test_lowlevel():
...
 
 
-def test_types():
-
-    need_internet()
+@pytest.mark.needs_internet
+def test_types(tmp_path):
 
-    fname1 = get_remote_file("images/stent.swf", test_dir)
+    fname1 = get_remote_file("images/stent.swf", tmp_path)
```
FirefoxMetzger, line 190: Let's try and not miss any places where a `test_images` fixture might be useful :D

tests/test_tifffile.py
```diff
@@ -23,14 +20,13 @@ def test_tifffile_format():
...
+@pytest.mark.needs_internet
+def test_tifffile_reading_writing(tmp_path):
     """Test reading and saving tiff"""
 
-    need_internet()  # We keep a test image in the imageio-binary repo
-
     im2 = np.ones((10, 10, 3), np.uint8) * 2
 
-    filename1 = os.path.join(test_dir, "test_tiff.tiff")
+    filename1 = os.path.join(tmp_path, "test_tiff.tiff")
```
FirefoxMetzger, line 29:
```suggestion
    filename1 = tmp_path / "test_tiff.tiff"
```

tests/test_tifffile.py
```diff
@@ -81,11 +77,11 @@ def test_tifffile_reading_writing():
     # meta = R.get_meta_data()
     # assert meta['orientation'] == 'top_left'  # not there in later version
     # Fail
-    raises(IndexError, R.get_data, -1)
-    raises(IndexError, R.get_data, 3)
+    pytest.raises(IndexError, R.get_data, -1)
+    pytest.raises(IndexError, R.get_data, 3)
```
FirefoxMetzger, line 81:
```suggestion
    with pytest.raises(IndexError)
        R.get_data(-1)

    with pytest.raises(IndexError):
        R.get_data(3)
```

tests/test_tifffile.py
```diff
@@ -81,11 +77,11 @@ def test_tifffile_reading_writing():
...
 
     # Ensure imread + imwrite works round trip
-    filename3 = os.path.join(test_dir, "test_tiff2.tiff")
+    filename3 = os.path.join(tmp_path, "test_tiff2.tiff")
```
FirefoxMetzger, line 84:
```suggestion
    filename3 = tmp_path / "test_tiff2.tiff"
```

tests/test_tifffile.py
```diff
@@ -94,7 +90,7 @@ def test_tifffile_reading_writing():
     assert (im1 == im3).all()
 
     # Ensure imread + imwrite works round trip - volume like
-    filename3 = os.path.join(test_dir, "test_tiff2.tiff")
+    filename3 = os.path.join(tmp_path, "test_tiff2.tiff")
```
FirefoxMetzger, line 93:
```suggestion
    filename3 = tmp_path / "test_tiff2.tiff"
```

tests/test_tifffile.py
```diff
@@ -133,6 +129,8 @@ def test_tifffile_reading_writing():
 
 def test_imagej_hyperstack(tmp_path):
     # create artifical hyperstack
+    import tifffile
+
```
FirefoxMetzger, line 133: Do we need this here? We have the `importorskip` behavior at the top.

FirefoxMetzger (changes requested):
That's some large amount of work. Thanks @anjos 🚀 

Most of my comments are about code styling. The two major ones are (1) I suggest to add a `test_images` fixture to avoid repeatedly downloading images during test execution, and (2) I refactored one of the fixtures that you added to use pytest's `tmp_path_factory`.

Regarding the two unused imports that you mentioned in the conversation, I think your current solution works fine and we can keep things as they are in this PR.

Otherwise, it looks pretty good :) Depending on your thoughts on the bigger points I raised we can probably merge this in the next round or round thereafter.

almarklein:
Looks good to me in general. I can't view the whole diff, because it seems it cant fully load it :D  I will have another try once @FirefoxMetzger's suggestions are resolved.

anjos:
Hello both of you - thanks for the review.  I'll have a look at this ASAP, but no before next week as I'm busy with other matters presently.  I apologize fo the delay.

almarklein:
No need to apologise - I think it applies to each of us that it's sometimes hard to find the time :)

anjos:
> Most of my comments are about code styling. The two major ones are (1) I suggest to add a `test_images` fixture to avoid repeatedly downloading images during test execution, and (2) I refactored one of the fixtures that you added to use pytest's `tmp_path_factory`.

I accepted all suggestions.  I'll have a look at the major suggestions next.
 
> Regarding the two unused imports that you mentioned in the conversation, I think your current solution works fine and we can keep things as they are in this PR.

OK.

> Otherwise, it looks pretty good :) Depending on your thoughts on the bigger points I raised we can probably merge this in the next round or round thereafter.

Let's continue.

FirefoxMetzger:
@anjos I pushed a few changes to refactor some of the tests that apparently don't support `pathlib` yet. It seemed easier to do it this way than to leave another round of reviews that would include minor changes.

There is still the question of adding a `test_images` fixture to avoid repeatedly downloading images into temporary folders. Let me know how you like that idea.

Finally, the pypy test matrix breaks currently because the `pypy-3.6` target doesn't seem to exist any longer; this doesn't seem related to this PR, so I will have a look at how to resolve this. Python 3.6 reaches end-of-life this year, so we will have to see how that affects us and also what is involved in ending support for it. I suspect that this is the cause of pypy-3.6 breaking here.

anjos:
Thanks.  I agree that a session-based downloader for all data should be cleaner than individual ones.

FirefoxMetzger:
@anjos I guess you are on vacation now and as such work on this PR will continue in the new year. 

I will leave this here as a note to pick up when you return:

- Could you merge master into this branch to resolve the current merge conflict and to get the latest CI config for pypy (should resolve the currently breaking tests)?
- Will you add the session-level fixture that downloads images or how should we handle it?

Have a good start into 2020 :)

anjos:
Hello there!  Indeed, I already started working on the session level downloading, but I started wondering on some aspects.  I’ll rebase and resolve conflicts.

FirefoxMetzger:
> [...] but I started wondering on some aspects

Sure. If you have questions feel free to ask them here and we can see what the best solution is.

anjos:
Here it goes a few points:

1.  Are we to download **all** required images at the start of the tests?  Some of the tests are only run if special dependencies are installed.  Shall we still download images related to those tests even if they don't get executed?  (The full list of required images is attached).  A different approach would be to cache at the begin of each individual module.  In this case, we would only download once, based on what gets executed.  It is a bit more complex, so I don't know if you prefer this route.
2. I'm assuming that tests that test the downloading itself should not be touched.  Please confirm.
3. What is the diffference between [test-images](https://github.com/imageio/imageio-binaries/tree/master/test-images) and [images](https://github.com/imageio/imageio-binaries/tree/master/images)?  Files like `chelsea.png` are available on both and sometimes downloaded from one or the other.  May be we should have just one list?
4. git vs `get_remote_file()`: I think we should use one method only to cache images.  Note that requring "git" for testing adds an extra dependence into this build phase.  The "complication" is that we need to download just a subdirectory from the `imageio-binaries` which then triggers the rather sophisticated git usage inside `conftest.py`.  If the repo would be completely downloaded once, then we can do this using a zip archive technique instead of actually requiring git to be available.  Would this be simpler?

## Required images from [test-images](https://github.com/imageio/imageio-binaries/tree/master/test-images) (these are "git downloaded" blindly at the begin):

```
README.md
chelsea.bmp
chelsea.bsdf
chelsea.jpg
chelsea.npy
chelsea.png
chelsea.zip
chelsea_jpg.npy
imageio_issue245.gif
imageio_issue246.png
kodim03.png
newtonscradle.gif
newtonscradle_rgb.npy
newtonscradle_rgba.npy
chelsea.noext
```

## Required images from [images](https://github.com/imageio/imageio-binaries/tree/master/images) (these are downloaded as required from `get_remote_file()`):

```
stent.npz
newtonscradle.gif
newtonscradle2.npz
dicom_sample1.zip
dicom_sample2.zip
dicom_file01.dcm
dicom_file03.dcm
dicom_file04.dcm
dicom_file90.dcm
dicom_file91.dcm
dicom_file02.dcm
fei-sem-rbc.tif
rommel.jpg
kodim03.png
'Ankylosaurus_&_Stegosaurus.LFR'
lenslet_whiteimage.RAW
Guitar.lfp
IMG_0001__frame.raw
chelsea.png
'Ankylosaurus_&_Stegosaurus_Thumbnail.jpg'
lenslet_whiteimage.TXT
IMG_0001__frame.json
imageio_issue245.gif
imageio_issue246.png
scribble_P_RGB.bmp
test_000_.SPE
stent.swf
stent.out.html
stent.out.swf
stent.compressed.swf
stent.out2.swf
stent.invalid.swf
stent.out3.swf
```

FirefoxMetzger:
> Are we to download all required images at the start of the tests? Some of the tests are only run if special dependencies are installed. Shall we still download images related to those tests even if they don't get executed? (The full list of required images is attached). A different approach would be to cache at the begin of each individual module. In this case, we would only download once, based on what gets executed. It is a bit more complex, so I don't know if you prefer this route.

Yes, I think it makes the most sense to download all required images in one go at the start of the test session (session-level fixture). A significant amount of slowdown comes from repeatedly opening and closing a connection with the remote server to download the images, so if we can avoid that it should speed things up. That said, I don't mind being sophisticated about choosing the actual images to download. We could check at runtime which plugins are available (or are selected as you suggest) and then only download those images that have formats supported by installed backends/plugins.

Another option could be to add a permanent local cache that gets created/updated once per session. Here git could come in handy to make sure the images stay in sync with the remote by simply doing a `git pull` inside a session-level fixture if the local cache already exists. If you like this idea, we could discuss where this should live.

> I'm assuming that tests that test the downloading itself should not be touched. Please confirm.

Yes, it's fine to not touch them. Though, if you want to mock those downloads, feel free to go ahead.

> What is the diffference between test-images and images? Files like chelsea.png are available on both and sometimes downloaded from one or the other. May be we should have just one list?

The idea now is that *images* should only contain the standard images (available via `iio.imread("imageio:<standard_image>"`) and that *test-images* contains all the images needed for tests (which are more plentiful than just the standard images).

*images* used to be a collection of both [standard images](https://imageio.readthedocs.io/en/stable/getting_started/standardimages.html) and images used for testing. I then introduced *test-images* a while ago to eventually separate standard images and test images. However, I didn't yet have the time to go through the old tests and change them to use the images from test-images.

> git vs get_remote_file(): I think we should use one method only to cache images. Note that requring "git" for testing adds an extra dependence into this build phase. The "complication" is that we need to download just a subdirectory from the imageio-binaries which then triggers the rather sophisticated git usage inside conftest.py. If the repo would be completely downloaded once, then we can do this using a zip archive technique instead of actually requiring git to be available. Would this be simpler?

There are two downsides to using `get_remote_file` in the test suite:

1. It is part of the library we are currently testing, and thus shouldn't be trusted to work correctly. (However, this is a more theoretical point.)
2. It requires explicit knowledge of the file to download and there is no "download all" option.

The first point is more of a nice to have. If we can avoid using it in a fixture then we should prefer to do so, but if it makes our life too complicated, then - from my side - it is okay to use it.

The second point is the bigger deal-breaker. We would have to maintain an explicit list of the files used by the test suite and update it whenever things change. This is not very intuitive, and easy to forget. Of course, we could write tooling to do this for us, but at this stage, it seems like overkill.

On the flip side, there are downsides to using `git` in the test suite:

1. It introduces git as a dependency. (Though I don't see this as too concerning, because 99.8% of people will get the ImageIO source via git and have that dependency already satisfied.)
2. The full repo in question (imageio-binaries) is really big (in the hundreds of MB), and we need to rely on a partial checkout, which makes the solution a bit hacky and unwieldy.

Just like before, the first point isn't that big an issue in my mind. Git is almost universally available in the development world, and outside of that, I don't think many people will wish to run the tests.

The second point is annoying, I agree, but I don't know how much can be done about it. In addition to the images `imageio-binaries` contains the binaries for FFmpeg, freeimage, etc. for all platforms and versions, which makes it bloated. I don't think it will be wise to download them all every time somebody executes a test, which rules out a normal git clone or zip download of the full repo.

To sum up the above, I think we probably have to stick with the git partial checkout approach. It seems to be the lesser of two evils. Then again, I'd be very happy if we could replace it with a more elegant solution, provided it can download all the required test images in one go and make them available to the test suite.

anjos:
Thanks for the detailed answer.

I'm taking away that we should keep git for the (partial) checkout.  Then, we should checkout all images in `test-images` **and** `images` once at start, and refrain from using `get_remote_file()` anywhere, except of course, for testing that function.

FirefoxMetzger:
> I'm taking away that we should keep git for the (partial) checkout. Then, we should checkout all images in test-images and images once at start, and refrain from using get_remote_file() anywhere, except of course, for testing that function.

Yes, that would be a workable solution. 

In the future, I will then copy all the standard images we need for the tests from *images* into *test-images* and clean up the *images* folder. However, this will happen after more pressing issues are resolved :D.

FirefoxMetzger:
Nice @anjos ! Let me know when you are happy with the updates and I will find some time to review the changes 🚀

anjos:
@FirefoxMetzger: here is a first attempt to implement what you asked.  Most tests now use a session-level fixture that shallow clones the two directories from image-binaries as agreed.  As a result, only a handful of tests actually (directly) require internet now.

Tests that download data (and cache it) use a temporary `$IMAGEIO_USERDIR`, so internet access is actually tested at each iteration - instead of relying on cached versions.  To disable those, it suffices to turn-off tests with the `needs_internet` marker.

anjos:
@FirefoxMetzger: I just fixed the last issue, I think.  Could you please let it run?

anjos:
@FirefoxMetzger: on the latest additions I fixed the linting errors and introduced a test option:

```sh
$ git clone https://github.com/imageio/imageio-binaries
$ pytest -sv tests --imageio-binaries file://`pwd`/imageio-binaries
```

Should allow for caching locally without compromising test reproducibility.  I still need to figure out the windows errors...

anjos:
@FirefoxMetzger: The latest commit seems to have solved also the problems under windows.  Please let the CI run for a last check!

anjos:
@FirefoxMetzger: a bit more to go before a complete green test.  We have a small issue with `test_freeimage`.

These tests are a bit crooked as all of them indirectly depend on a libfreeimage download from the imageio-binaries repository.  First of all, they all `needs_internet` because of this and then, we don't really test if it works against the installed version of freeimage instead, if one is available.

In this context can we possibly consider to change this to follow a similar policy as of the rest of the "cached" images?  That is, we download the `freeimage` directory somewhere during the session-level initialization, together with the rest of the other images.  We then rename the function `get_library()` at `tests/test_freeimage.py` to `setup_library()`.  There, we just try to find the system-wide installed version of libfreeimage, and test with that if available.  Otherwise, we use the version on the "image_cache".  The download test remains unchanged and will continue to `needs_internet`.

anjos:
The question is, of course, how to implement this correctly if we take this direction.

anjos:
@FirefoxMetzger: please let the CI run on the latest rebase.

anjos:
@FirefoxMetzger: there is another small set of fixed to resolve the broken builds.  Let's see if it finally gets green.

FirefoxMetzger:
@anjos Absolutely. By the way, tests will start automatically for contributors, so once your first PR is merged you will see the results immediately (plus/minus execution time ;) ).

I am attending a PhD Defense today, but I am still free afterward, so my plan was to review this PR in the afternoon/evening unless something urgent comes up. Otherwise it will happen tomorrow. 

Also, I just noticed that the CI is already done while I am still typing this. Your PR appears to make the tests extremely fast 🚀 Thanks for that 🎉

tests/test_freeimage.py
```diff
@@ -3,35 +3,65 @@
...
+    if use_imageio_binary:
+
+        # Setup from image_cache/freeimage
+        ud = tmp_path_factory.getbasetemp() / "userdir"
+        if sys.platform.startswith("win"):
+            if "LOCALAPPDATA" in os.environ:  # saves it
+                os.environ["OLD_LOCALAPPDATA"] = os.environ["LOCALAPPDATA"]
+            os.environ["LOCALAPPDATA"] = str(ud)
+        else:
+            os.environ["IMAGEIO_USERDIR"] = str(ud)
```
FirefoxMetzger, line 40: The same changes I suggested in `conftest.py` apply here.

tests/test_freeimage.py
```diff
@@ -3,35 +3,65 @@
...
-        # We cannot download; skip all freeimage tests
-        need_internet()
-
-
-def teardown_module():
-    # Set back to normal
+    if use_imageio_binary:
+
+        # Setup from image_cache/freeimage
+        ud = tmp_path_factory.getbasetemp() / "userdir"
```
FirefoxMetzger, line 34:
```suggestion
        ud = tmp_path_factory.mktemp("userdir")
```

this way pytest will take care of the cleanup.

tests/test_freeimage.py
```diff
@@ -3,35 +3,65 @@
...
+            del os.environ["LOCALAPPDATA"]
+            if "OLD_LOCALAPPDATA" in os.environ:
+                os.environ["LOCALAPPDATA"] = os.environ["OLD_LOCALAPPDATA"]
+                del os.environ["OLD_LOCALAPPDATA"]
+        else:
+            del os.environ["IMAGEIO_USERDIR"]
+
+        # on windows, loaded DLLs may refuse to get deleted
+        # there is no clean way to release those, so we leave it there
+        shutil.rmtree(ud, ignore_errors=True)
```
FirefoxMetzger, line 62:
```suggestion
```

DLLs won't get deleted, because they are still loaded. You can force them to unload, but I don't know how to do that in python (never had to). Either way, we can let pytest take care of cleaning up :)

tests/test_pillow.py
```diff
@@ -50,15 +50,17 @@ def test_write_single_frame(image_files: Path, im_npy: str, im_out: str, im_comp
         # ("newtonscradle_rgba.npy", "iio.gif", "pil.gif"),
     ],
 )
-def test_write_multiframe(image_files: Path, im_npy: str, im_out: str, im_comp: str):
+@pytest.mark.needs_internet
+def test_write_multiframe(image_cache, tmp_path, im_npy, im_out, im_comp):
```
FirefoxMetzger, line 54: @anjos I noticed this in other places within this file, too, but I will only ask here: How come you ended up removing the type annotations from the tests?

tests/test_pillow_legacy.py
```diff
@@ -369,56 +398,62 @@ def test_gamma_correction():
...
+def test_inside_zipfile(image_cache, tmp_path):
 
-    fname = os.path.join(test_dir, "pillowtest.zip")
+    fname = str(tmp_path / "pillowtest.zip")
     with ZipFile(fname, "w") as z:
-        z.writestr("x.png", open(get_remote_file("images/chelsea.png"), "rb").read())
-        z.writestr("x.jpg", open(get_remote_file("images/rommel.jpg"), "rb").read())
+        z.writestr(
+            "x.png",
+            open(image_cache / "test-images" / "chelsea.png", "rb").read(),
```
FirefoxMetzger, line 407:
```suggestion
            (image_cache / "test-images" / "chelsea.png).read_bytes(),
```

tests/test_pillow_legacy.py
```diff
@@ -369,56 +398,62 @@ def test_gamma_correction():
...
+        )
+        z.writestr(
+            "x.jpg",
+            open(image_cache / "images" / "rommel.jpg", "rb").read(),
```
FirefoxMetzger, line 411:
```suggestion
            (image_cache / "images" / "rommel.jpg").read_bytes(),
```

tests/test_tifffile.py
FirefoxMetzger: @anjos Same as earlier in the review; wouldn't `importskip` at the top suffice?

tests/conftest.py
```diff
@@ -1,62 +1,119 @@
...
-        os.system("git sparse-checkout set test-images")
-    else:
-        # sparse checkout on GH Actions
-        os.system(
-            "git clone --sparse --filter=blob:none https://github.com/imageio/imageio-binaries.git ."
-        )
-        os.system("git sparse-checkout init --cone")
-        os.system("git sparse-checkout set test-images")
-    os.chdir(current_path)
+def image_cache(pytestconfig, tmp_path_factory):
```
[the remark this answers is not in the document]
anjos: Sure, please go ahead and push these to the branch.

tests/conftest.py
```diff
@@ -1,62 +1,119 @@
...
-    yield data_dir
 
-    shutil.rmtree(data_dir)
+def pytest_collection_modifyitems(items):
 
+    # all tests using the image_cache fixture, require internet automatically
+    # iff they checkout from the remote repository - everything else does not
+    internet_protocols = ("git", "http", "ssh")
+    for item in items:
+        if ("image_cache" in getattr(item, "fixturenames", ())) or (
```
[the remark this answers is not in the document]
anjos: OK, a fair point.  Please go ahead and push these changes.

tests/conftest.py
```diff
@@ -81,3 +138,29 @@ def invalid_file(tmp_path, request):
         file.write("Actually not a file.")
 
     return tmp_path / ("foo" + ext)
+
+
+@pytest.fixture()
+def tmp_userdir(tmp_path):
+
+    ud = tmp_path / "userdir"
+    os.makedirs(ud, exist_ok=True)
```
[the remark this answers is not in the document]
anjos: Fair.

tests/conftest.py
```diff
@@ -81,3 +138,29 @@ def invalid_file(tmp_path, request):
...
+
+    yield ud
+
+    if sys.platform.startswith("win"):
+        del os.environ["LOCALAPPDATA"]
+        if "OLD_LOCALAPPDATA" in os.environ:
+            os.environ["LOCALAPPDATA"] = os.environ["OLD_LOCALAPPDATA"]
+            del os.environ["OLD_LOCALAPPDATA"]
+    else:
+        del os.environ["IMAGEIO_USERDIR"]
```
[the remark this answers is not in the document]
anjos: Frankly speaking, I hate environment variables - these are glorified globals...  Moreover, I'm not a Windows expert.  I have a VM lying around for these tests.  Let's remove it and I'll put it back if needs be.

tests/test_core.py
```diff
@@ -57,29 +54,24 @@ def __init__(self, request):
...
-    fname3 = get_remote_file("images/chelsea.png", test_dir, True)
+    fname3 = get_remote_file("images/chelsea.png", tmp_userdir, True)
     mtime3 = os.path.getmtime(fname3)
     # This should too (update this if imageio is still around in 1000 years)
-    fname4 = get_remote_file("images/chelsea.png", test_dir, "3014-01-01")
+    fname4 = get_remote_file("images/chelsea.png", tmp_userdir, "3014-01-01")
     mtime4 = os.path.getmtime(fname4)
     # This should not
-    fname5 = get_remote_file("images/chelsea.png", test_dir, "2014-01-01")
+    fname5 = get_remote_file("images/chelsea.png", tmp_userdir, "2014-01-01")
```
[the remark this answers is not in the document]
anjos: I tested it with `tmp_path` instead and still passes, so we should simplify it!

tests/test_core.py
```diff
@@ -140,8 +132,6 @@ def test_findlib2():
     if not sys.platform.startswith("linux"):
         skip("test on linux only")
 
-    need_internet()  # need our own version of FI to test this bit
```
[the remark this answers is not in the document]
anjos: Does it need internet?  From the code, it does not look like it.  Please confirm it does.

tests/test_ffmpeg.py
```diff
@@ -9,35 +9,33 @@
...
 
-test_dir = get_test_dir()
+pytest.importorskip("psutil", reason="ffmpeg support cannot be tested without psutil")
+
+pytest.importorskip("imageio_ffmpeg", reason="imageio-ffmpeg is not installed")
 
 
 def get_ffmpeg_pids():
+    import psutil
+
```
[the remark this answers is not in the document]
anjos: No, this is not required for as long as we attribute the output of `pytest.importorskip()` like `psutil = pytest.importorskip("psutil")`.  I'll do that.

tests/test_ffmpeg.py
```diff
@@ -9,35 +9,33 @@
...
 
 
-def setup_module():
-    pass
-
-
 def test_get_exe_installed():
     if platform.machine() == "aarch64":
-        skip("Skip for aarch64")
+        pytest.skip("Skip for aarch64")
```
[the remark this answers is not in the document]
anjos: Overlooked this one.  It is better as you suggested.  I'll add it.

tests/test_ffmpeg.py
```diff
@@ -489,15 +492,15 @@ def test_webcam():
...
+        with pytest.raises(ValueError):
             imageio.read(path)
 
 
 def test_webcam_get_next_data():
     try:
         reader = imageio.get_reader("<video0>")
     except IndexError:
-        skip("no webcam")
+        pytest.skip("no webcam")
```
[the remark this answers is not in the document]
anjos: I preserved the previous code here.  I can change it to xfail instead.

tests/test_ffmpeg.py
```diff
@@ -558,7 +561,7 @@ def test_webcam_process_termination():
         assert reader._read_gen is None
         assert not get_ffmpeg_pids().difference(pids0)
     except IndexError:
-        skip("no webcam")
+        pytest.skip("no webcam")
```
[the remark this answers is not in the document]
anjos: Will push this in a bit.

tests/test_fits.py
```diff
@@ -54,12 +44,10 @@ def test_fits_format():
...
 
-@pytest.mark.skipif("astropy is None")
+@pytest.mark.needs_internet
 def test_fits_reading():
     """Test reading fits"""
 
-    need_internet()  # We keep the fits files in the imageio-binary repo
-
     if IS_PYPY:
         return  # no support for fits format :(
```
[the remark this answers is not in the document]
anjos: Let's do it this way.  It is simpler and therefore better.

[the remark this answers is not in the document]
anjos: I'll push these changes in a bit.

tests/test_fits.py
```diff
@@ -94,12 +82,12 @@ def test_fits_reading():
     assert im.shape == (2042, 3054)
 
 
-@pytest.mark.skipif("astropy is None")
 @pytest.mark.skipif("IS_PYPY", reason="pypy doesn't support astropy.fits.")
 def test_fits_get_reader(normal_plugin_order, tmp_path):
     """Test reading fits with get_reader method
     This is a regression test that closes GitHub issue #636
     """
+    import astropy.io.fits
```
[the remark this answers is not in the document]
anjos: OK.

tests/test_format.py
```diff
@@ -1,27 +1,21 @@
 from pytest import raises
 import pytest
-from imageio.testing import run_tests_if_main, get_test_dir
 
-import os
 import gc
 import shutil
+import pathlib
```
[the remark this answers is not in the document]
anjos: OK.

tests/test_freeimage.py
```diff
@@ -3,35 +3,65 @@
...
-test_dir = get_test_dir()
+@pytest.fixture(scope="module")
+def setup_library(tmp_path_factory, image_cache):
 
+    # Checks if freeimage is installed by the system
+    from imageio.plugins.freeimage import fi
+
+    use_imageio_binary = False
+    if not fi.has_lib():
+        use_imageio_binary = True
```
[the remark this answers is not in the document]
anjos: And better, of course.

tests/test_freeimage.py
anjos: Excellent.

tests/test_core.py
```diff
@@ -121,17 +114,16 @@ def test_fetching():
...
     # Meh, just skip
-    skip("always skip, is tested implicitly anyway")
-
     if not sys.platform.startswith("linux"):
         skip("test on linux only")
 
     # Candidate libs for common lib (note, this runs only on linux)
-    dirs, paths = core.findlib.generate_candidate_libs(["libc"])
+    _, paths = core.findlib.generate_candidate_libs(["libc"])
     assert paths
```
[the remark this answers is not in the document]
anjos: I'll leave this unresolved for now.

tests/test_freeimage.py
anjos: OK, will apply.

tests/test_freeimage.py
```diff
@@ -3,35 +3,65 @@
...
+
+    if use_imageio_binary:
+
+        if sys.platform.startswith("win"):
+            del os.environ["LOCALAPPDATA"]
+            if "OLD_LOCALAPPDATA" in os.environ:
+                os.environ["LOCALAPPDATA"] = os.environ["OLD_LOCALAPPDATA"]
+                del os.environ["OLD_LOCALAPPDATA"]
+        else:
+            del os.environ["IMAGEIO_USERDIR"]
```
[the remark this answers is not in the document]
anjos: OK.

tests/test_freeimage.py
anjos: Very well - I'd prefer that as well.

anjos: I'm removing all code that cleans up.

tests/test_tifffile.py
anjos: Yes, will fix.

tests/test_pillow_legacy.py
anjos: Noice!

tests/conftest.py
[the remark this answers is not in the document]
FirefoxMetzger: perfect. We did another change to that fixture in #736 which is causing the current merge conflict. 

@anjos To avoid bumping into each other, let me know when you are done with making changes here. I will then go and resolve the merge conflict and apply the changes discussed here.

tests/test_core.py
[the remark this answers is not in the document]
FirefoxMetzger: Yep, good catch. It indeed doesn't use internet. In this case, there is nothing to do :)

tests/test_pillow.py
anjos: Frankly, I'm not sure anymore on this one.  Let me check again.

tests/conftest.py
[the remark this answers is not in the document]
anjos: Please go ahead.  I think I'm done for this round.

FirefoxMetzger:
This should resolve most of the merge conflict. I need to have a closer look at the freeimage tests in the next few days, because we need to find a different way to download the binaries, but the other tests should run fine.

anjos:
OK, let me know of anything that comes to mind.

FirefoxMetzger:
@anjos I think I do have a solution actually. I've added a new module-level fixture that downloads the freeimage library where needed. This uses the GitHub API under the hood, which has quite heavy rate limits; however, we only download a few files using it, so it should be fine.

Let's see what CI says.

anjos:
Except for the relative drop in test coverage, everything else is green:  Thank you @FirefoxMetzger.  I'm OK with the current state of things.

FirefoxMetzger:
I, too, am (almost) happy. There is one thing left that I just noticed when looking everything over:

@anjos You added a `--imageio-binaries` option to the test suite for which I may have unknowingly removed any functionality that it once had. Do you have a particular use for this, or shall we remove it?

anjos:
The function of this was to allow the user to provide their own copy of the test repo if needed be.  I was not particularly attached to it, but avoided the hard-coding of the repo address on the code itself, which I thought was nicer.  If you prefer hard-coding, I'm fine with it.

FirefoxMetzger:
@anjos Alright. Then let's bring it back, but let's make it an undocumented feature.

FirefoxMetzger:
I've hooked things back up again. Though, I renamed the variable to `--test-images` to better reflect what it actually does. 

I will leave the freeimage binaries hard-coded for now. I feel a bit uneasy exposing a direct way to change the location from where to fetch executable binaries; at least without thinking deeply about why this is a sane thing to do.

anjos:
Right, looks great as is already.  Thank you for the openness and work on this!

FirefoxMetzger (approved):

FirefoxMetzger:
Once CI is happy (minus the coverage change) I will merge :)

FirefoxMetzger:
Thanks, @anjos for this impressive piece of work 🚀. I think this just made our test suite 100x better 🎉

anjos:
Noice 👍🏻

Issue #697, opened 2021-11-26
ffmpeg.Reader._get_cam_inputname causes "ResourceWarning: unclosed file"


dennisvang:
### Description

In my tests I keep seeing `ResourceWarning: unclosed file <_io.BufferedWriter name=3>`

With `PYTHONTRACEMALLOC=1`, here's what I get:

```
...\lib\site-packages\imageio\plugins\ffmpeg.py:303: ResourceWarning: unclosed file <_io.BufferedWriter name=3>
  self._filename = self._get_cam_inputname(index)
Object allocated at (most recent call last):
  File "...\lib\subprocess.py", lineno 935
    self.stdin = io.open(p2cwrite, 'wb', bufsize)
...\lib\site-packages\imageio\plugins\ffmpeg.py:303: ResourceWarning: unclosed file <_io.BufferedReader name=4>
  self._filename = self._get_cam_inputname(index)
Object allocated at (most recent call last):
  File "...\lib\subprocess.py", lineno 941
    self.stdout = io.open(c2pread, 'rb', bufsize)
...\lib\site-packages\imageio\plugins\ffmpeg.py:303: ResourceWarning: unclosed file <_io.BufferedReader name=5>
  self._filename = self._get_cam_inputname(index)
Object allocated at (most recent call last):
  File "...\lib\subprocess.py", lineno 946
    self.stderr = io.open(errread, 'rb', bufsize)
```

### How to reproduce

On a Windows 10 laptop with built-in webcam (`<video0>`), using Python 3.9.7, `imageio` 2.11.0, `imageio-ffmpeg` 0.4.5, run the following test:

```python
from imageio import get_reader
import unittest

class ReaderTests(unittest.TestCase):
    def test_get_reader(self):
        reader = get_reader(uri='<video0>')
```

### Cause

Looking at the [source of _get_cam_inputname](https://github.com/imageio/imageio/blob/v2.12.0/imageio/plugins/ffmpeg.py#L209), and looking at the trace above, I suspect the issue is caused by the `stdin`, `stdout`, and `stderr` streams not being closed properly after use in `subprocess.Popen`.

### Proposed solution

Use `subprocess.Popen` as a context manager, to ensure streams are always properly closed, as suggested e.g. [here](https://stackoverflow.com/a/58696973).

Replace this ([source](https://github.com/imageio/imageio/blob/v2.12.0/imageio/plugins/ffmpeg.py#L228)):
```python
...
proc = sp.Popen(
[...]

PR #711, opened 2021-12-17
Fix issue 697: close standard streams properly, using sp.run

dennisvang:
This should fix issue #697.

The test has a temporary workaround, which could be removed after the acceptance of https://github.com/imageio/imageio-ffmpeg/pull/62 and subsequent release. Not sure how you would like to handle that.

### DIFF
--- a/imageio/plugins/ffmpeg.py
+++ b/imageio/plugins/ffmpeg.py
@@ -231,8 +231,12 @@ class FfmpegFormat(Format):
                 # using `-list_devices` (or `-list_options`), even if the
                 # command is successful, so we set `check=False` explicitly.
                 completed_process = sp.run(
-                    cmd, stdout=sp.PIPE, stderr=sp.PIPE, encoding="utf-8",
-                    shell=True, check=False
+                    cmd,
+                    stdout=sp.PIPE,
+                    stderr=sp.PIPE,
+                    encoding="utf-8",
+                    shell=True,
+                    check=False,
                 )
 
                 # Return device name at index
--- a/imageio/plugins/ffmpeg.py
+++ b/imageio/plugins/ffmpeg.py
@@ -231,8 +231,8 @@ class FfmpegFormat(Format):
                 # using `-list_devices` (or `-list_options`), even if the
                 # command is successful, so we set `check=False` explicitly.
                 completed_process = sp.run(
-                    cmd, capture_output=True, encoding="utf-8", shell=True,
-                    check=False)
+                    cmd, capture_output=True, encoding="utf-8", shell=True, check=False
+                )
 
                 # Return device name at index
                 try:
--- a/imageio/plugins/ffmpeg.py
+++ b/imageio/plugins/ffmpeg.py
@@ -224,18 +224,19 @@ class FfmpegFormat(Format):
                     "-i",
                     "dummy",
                 ]
-                # Set `shell=True` in sp.Popen to prevent popup of a command line
-                # window in frozen applications. Note: this would be a security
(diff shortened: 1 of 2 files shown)


almarklein:
Is this an alternative for the fix in imageio-ffmpeg, or do they work in conjunction?

FirefoxMetzger (approved):
@dennisvang Thanks for this 🚀 Looking good from my end.

@almarklein Yes, the changes are in addition to the changes in `ffmpeg-imageio`, since we are leaking file handles in both places. Once when checking available cameras (handled here) and then later on when closing the camera resource (handled in ffmpeg-imageio).

almarklein (approved):

FirefoxMetzger:
@dennisvang Could you merge this branch with the latest master (rebase also works depending on the workflow you prefer)? I fixed the pypy problem and a merge should allow the full matrix to run on the PR.

dennisvang:
@FirefoxMetzger Sure, no problem.

Still failing on macos, I see. I guess the neatest solution would be to drop the [temporary part](https://github.com/imageio/imageio/pull/711/files#diff-e4e9cfdd4597c5c621ea46cab4445fbb747400f26b5c56c2a80fa0fe3cd1eeabR582) of that test.

Or just skip the assertion until the next release of `imageio_ffmpeg`, something like the following?

```python
def test_webcam_resource_warnings():
    try:
        with warns(None) as warnings:
            with imageio.get_reader("<video0>"):
                pass
    except IndexError:
        skip("no webcam")

    # todo: remove version check when imageio_ffmpeg issue #61 has been fixed
    import imageio_ffmpeg

    if tuple(int(v) for v in imageio_ffmpeg.__version__.split('.')) > (0, 4, 5):
        # there should be no warnings, but show warning messages if there are
        assert not [w.message for w in warnings]
```

FirefoxMetzger:
@dennisvang Yeah, though we did progress by one step since we get to run the actual tests now instead of failing in the setup 😃 

The CPython MacOS tests pass fine, so I am surprised to see them fail on pypy. There is some difference in the memory management model though, so it could be that there is no leaking of `imageio-ffmpeg` on pypy?

There is also the failure on Windows in CPython that I haven't seen before. Not sure if this is a flakey test or if something happened. I will trigger a re-run of the CI and then we'll know more.

FirefoxMetzger:
Okay, the windows tests still break with the same error messages; however #714 that was running through CI in parallel finished cleanly. This could be evidence that something is going wrong in this PR, but I don't have the time to dive into this right now. :(

Regarding the MacOS pypy test, I noticed that the linux builds all skip this test (because there is now webcam installed). The Windows and MacOS tests on the other hand run it but fail. I'm not sure if the GH Actions CI servers have webcams installed, so this could be an angle to investigate.

dennisvang:
@FirefoxMetzger 

- Modified the `sp.run()` call to make it python 3.6 compatible. 
- Modified the *temporary* test so it allows "at most" one warning (from `imageio_ffmpeg`). 

I'm seeing all green now, except for the codecov/project check. Not sure what to do about that.

FirefoxMetzger:
@dennisvang Thanks for that :) 

The codecov test is fine. We are removing a few lines that were covered before which slightly reduces the total number of lines in the project but keeps the number of uncovered lines the same. Thus, the coverage slightly decreases, but I don't mind that here.

FirefoxMetzger:
@dennisvang I just went over the PR again, and it still looks good. Do you want to add anything more, or can I merge it?

dennisvang:
@FirefoxMetzger That's great, thanks. :-) I have nothing more to add.

Issue #708, opened 2021-12-14
Correct way to write a multi-layered tiff file for reading in Photoshop?


KeygenLLC:
Hello again

I've been trying to use imageio to write a multi-layered TIFF file that can be read by Photoshop. I prepared a variable of multiple numpy arrays (the images read in by imageio).

If I use `mimwrite` I seem to only get one layer in Photoshop. It is RGB and at the correct bit depth. I'm using tifffile as the plugin driver.

By _LAYERS_, I mean individual, editable layers that each contain their own independent RGB data in Photoshop. If layered files are written out of Photoshop as TIFF without flattening, they can be read back in just fine, so not sure why this is not working.

When searching for an answer, most people seem to refer to color channels as layers (R, G, B, Alpha1, Alpha2) when working with images in Python. They refer to "Multipage" TIFF, so maybe that's the terminology. I didn't see anything in the imageio docs about writing Multipage, only a patch note saying Multipage was supported when reading TIFF files. 

The data _seems_ like it's in the file because I merged 3 files that were 1.1 MB each and the resulting TIFF is 3.2MB, but Photoshop only seems to recognize the first file layer of RGB data that was read in. There are only 3 channels in the file (R,G,B).

Not sure if I'm missing something, if I have to use some kind of loop to create the file and then add each opened image as a layer, or if there's some extra data that needs to be added to the file for Photoshop to recognize them properly.

Here's the test code (applicable part):
```python
image_data = []

for idx, f in enumerate(filepaths):
    image = imageio.imread(f)
    image_data.append(image)
    
image_array = np.array(image_data)
imageio.mimwrite(outtif, image_array, format="tif")

# I tried this also with same result, since it seems to be a numpy array when opened
for idx, f in enumerate(filepaths):
    image = imageio.imread(f)
    image_data.append(image)

imageio.mimwrite(outtif, image_data, format="tif")
```
Printing the shape of `image_array` produces:
```plain
(3, 300, 300, 3)
```

Here's the expected results in Photoshop:
<img width="257" alt="proper_layers" src="https://user-images.githubusercontent.com/92394639/146094749-0908143c-4ebb-4e38-82e9-b040ed0c3aa1.png">

Actual result:
<img width="257" alt="no_layers" src="https://user-images.githubusercontent.com/92394639/146094533-ff580ce5-7593-4442-87d5-0d397d884497.png">

Would appreciate a point in the right direction.

Thanks

cgohlke:
[...]

PR #714, opened 2021-12-21
BUG: clean up PillowPlugin destructor

FirefoxMetzger:
I was looking through the files in https://github.com/imageio/imageio/issues/708 and noticed that the v3 PillowPlugin can produce a warning/exception in the deconstructor. For me, this happened when the plugin was called with invalid arguments (and hence wasn't constructed properly). This PR addresses this and also streamlines one error message.


FirefoxMetzger:
@almarklein RTR :)

Issue #715, opened 2021-12-21
Reconsider release frequency policy


kmilos:
It's great that we have more regular releases after such a long pause (thank you!), but can we please do them in a more meaningful way, i.e. real feature or bug fix, and perhaps when a couple of them accumulate? Like a monthly rather than a weekly/daily basis that seems to be done now via some automation as soon as a line changes somewhere in the repository?

I don't want to have to be reminded to re-package imageio for distribution constantly and to have to check changes just to find an URL update for example and then decide to skip it...

FirefoxMetzger:
Hey @kmilos. Yes, a monthly release cycle is possible though I personally like a daily cycle because it means that there is little delay between having a bug fixed and having the fix propagate to whoever reported it. I like to imagine that this encourages contribution because you can benefit from your own work more easily.

The type of release is determined by a leading tag in a commit message FEAT/BUG/MAINT/DOC/... which also determines if a release should be made in the first place. For example, DOC releases won't trigger a release and simply accumulates until the next version-changing change is committed. What is currently missing in this setup, however, is a tag for dev/tooling changes, that also just accumulates. This is because, in my infinite wisdom (lol), I set up the `DEV:` commit to bump the version minor. This, however, can be fixed upstream, or we could run our own version parser locally (which may be the smart thing to do since we won't get surprised by upstream changing things).

Whichever path we choose, it is probably useful to revisit what should happen for each commit message tag. There was little external input when I first wrote it, so I am very happy to discuss if you have suggestions for improvement :). The current config resides here: https://github.com/relekang/python-semantic-release/blob/master/semantic_release/history/parser_scipy.py#L82-L99

So in sum, I think we can probably solve the "empty releases" problem by being smarter about the above config, but - independently of that - we can also discuss a less frequent release cycle if there is a need for that. For the release cycle I'd also like to hear @almarklein s thoughts 😃.

almarklein:
I think @kmilos makes a good point. If you follow the release, then daily releases will feel like spam.

My suggestion would be an automated monthly release, with the option to trigger a new release in case urgency is required (only for security issues, I suppose).

As for people contributing, the fact that we can tell the user exactly when their change will be available in the new release is already a big benefit. I mean, as opposed to saying "I will make a release soon" and then forgetting about it (I plead guilty).

kmilos:
@FirefoxMetzger That system you have built is pretty neat indeed! I'd just like it to be a little bit less trigger-happy ;)

[...]

PR #716, opened 2021-12-31
MAINT: reduce release frequency to weekly

FirefoxMetzger:
Related to: https://github.com/imageio/imageio/issues/715

This is the first step to closing the above-mentioned issue. It reduces the frequency with which we check for release-worthy changes in the repo from once a day to once a week.


FirefoxMetzger:
@almarklein RTR :)

PR #717, opened 2022-01-03
MAINT: reorganize commit tags

FirefoxMetzger:
Closes: https://github.com/imageio/imageio/issues/715

This PR reorganizes the resolution of commit messages to version bumps to make the library less chatty. For this, we add our own commit message parser which is very similar to the scipy commit message parser that we already use, except that we avoid bumping the version for most commits. In combination with a cap on the release cycle (weekly), I think this will result in a much lower frequency at which new versions are released.


almarklein (approved):
I'm not familiar with this code, but it looks good to me :)

PR #718, opened 2022-01-03
DOC: v3 API docs

FirefoxMetzger:
Closes: https://github.com/imageio/imageio/issues/640

Long overdue, but let's get this started.

This PR adds documentation of the ImageIO v3 API and migration instructions to navigate the breaking changes when moving from v2 to v3. I will put the documentation for writing plugins into a separate PR to try and not make this one too big.

### DIFF
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -28,37 +28,43 @@ If the image is a GIF:
 
     import imageio.v3 as iio
     
+    # index=None means: read all images in the file and stack along first axis
     frames = iio.imread("imageio:newtonscradle.gif", index=None)
     # ndarray with (num_frames, height, width, channel)
-    print(frames.shape)  # (36, 150, 200, 3)
-    
-If the GIF is stored in memory:
-
-.. code-block:: python
-
-    import imageio.v3 as iio
-
-    # create bytes containing an encoded GIF
-    frames = iio.imread("imageio:newtonscradle.gif", index=None)
-    bytes_image = iio.imwrite("<bytes>", frames, format_hint=".gif")
-    
-    # read GIF from bytes
-    frames = iio.v3.imread(bytes_image, index=None)
-    
+    print(frames.shape)  # (36, 150, 200, 3)   
 
 Read from fancy sources
 -----------------------
 
-Imageio can read from filenames, file objects, http, zipfiles and bytes.
+Imageio can read from filenames, file objects.
 
 .. code-block:: python
 
     import imageio.v3 as iio
-    import matplotlib.pyplot as plt
+    import io
 
-    im = iio.imread('http://upload.wikimedia.org/wikipedia/commons/d/de/Wikipedia_Logo_1.0.png')
-    plt.imshow(im, cmap="gray")
-    plt.show()
+
+    # from HTTPS
+    web_image = "https://upload.wikimedia.org/wikipedia/commons/d/d3/Newtons_cradle_animation_book_2.gif"
+    frames = iio.imread(web_image, index=None)
+
+    # from bytes
+    bytes_image = iio.imwrite("<bytes>", frames, format_hint=".gif")
+    frames = iio.imread(bytes_image, index=None)
+
+    # from byte streams
+    byte_stream = io.BytesIO(bytes_image)
+    frames = iio.imread(byte_stream, index=None)
+   
+    # from file objects
+    class MyFileObject:
+        def read(size:int=-1):
+            return bytes_image
+
+        def close():
+            return  # nothing to do
+
+    frames = iio.imread(MyFileObject())
 
 Read all Images in a Folder
 ---------------------------
@@ -73,7 +79,7 @@ Read all Images in a Folder
         if not file.is_file():
             continue
 
-        images.append(iio.v3.imread(file))
+        images.append(iio.imread(file))
 
 Note, however, that ``Path().iterdir()`` does not guarantees the order in which
 files are read.
@@ -97,38 +103,38 @@ Grab screenshot or image from the clipboard
 
 .. code-block:: python
 
-    import imageio as iio
+    import imageio.v3 as iio
 
-    im_screen = iio.v3.imread('<screen>')
-    im_clipboard = iio.v3.imread('<clipboard>')
+    im_screen = iio.imread('<screen>')
+    im_clipboard = iio.imread('<clipboard>')
 
 
 Grab frames from your webcam
 ----------------------------
 
-Use the special ``<video0>`` uri to read frames from your webcam (via the ffmpeg
-plugin). You can replace the zero with another index in case you have multiple
-cameras attached. You need to ``pip install imageio[ffmpeg]`` in order to use
-this plugin.
+.. note::
+    For this to work, you need to install the ffmpeg backend::
+
+        pip install imageio[ffmpeg]
 
 .. code-block:: python
 
-    import imageio as iio
-    import visvis as vv
+    import imageio.v3 as iio
+    import numpy as np
 
-    with iio.v3.imopen("<video0>", "r") as reader:
-        t = vv.imshow(reader.read(index=0), clim=(0, 255))
-        for im in reader.iter():
-            vv.processEvents()
-            t.SetData(im)
+    for idx, frame in enumerate(iio.imiter("<video0>")):
+        print(f"Frame {idx}: avg. color {np.sum(frame, axis=-1)}")
 
+Note: You can replace the zero with another index in case you have multiple
+devices attached.
 
 Convert a short movie to grayscale
 ----------------------------------
 
-Here we take a movie and convert it to gray colors. Of course, you could apply
-any kind of (image) processing to the image here. You need to ``pip install
-imageio-ffmpeg`` in order to use the ffmpeg plugin.
+.. note::
+    For this to work, you need to install the ffmpeg backend::
+
+        pip install imageio[ffmpeg]
 
 .. code-block:: python
 
@@ -136,17 +142,18 @@ imageio-ffmpeg`` in order to use the ffmpeg plugin.
     import numpy as np
 
     # read the video (it fits into memory)
-    with iio.v3.imopen("imageio:cockatoo.mp4", "r") as reader:
-        frames = reader.read(index=None)
-        fps = reader.get_meta()["fps"]
-
-    # convert the video
+    # Note: this will open the image twice. Check the docs (advanced usage) if
+    # this is an issue for your use-case
+    metadata = iio.immeta("imageio:cockatoo.mp4", exclude_applied=False)
+    frames = iio.imread("imageio:cockatoo.mp4", index=None)
+    
+    # manually convert the video
     gray_frames = np.dot(frames, [0.2989, 0.5870, 0.1140])
     gray_frames = np.round(gray_frames).astype(np.uint8)
     gray_frames_as_rgb = np.stack([gray_frames] * 3, axis=-1)
 
     # write the video
-    iio.v3.imwrite("cockatoo_gray.mp4", gray_frames_as_rgb, fps=fps)
+    iio.imwrite("cockatoo_gray.mp4", gray_frames_as_rgb, fps=metadata["fps"])
 
 
 Read medical data (DICOM)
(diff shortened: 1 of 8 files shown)


docs/reference/core_v3.rst
```diff
@@ -0,0 +1,122 @@
...
+.. note::
+    For migration instructions from the v2 API check :ref:`our migration guide
+    <migration_from_v2>`.
+
+This API provides a functional interface to all of ImageIOs backends. It is
+split into two parts: :ref:`Everyday Usage <v3_basic_usage>` and :ref:`Advanced
+Usage <v3_advanced_usage>`. The first part (everyday usage) focusses on sensible
+defaults and automation to cover everyday tasks such as plain reading or writing
+of images. The second part (advanced usage) focusses on providing fully-featured
+and performant access to all the a particular backend.
```
almarklein, line 16: I *think* you meant this?
```suggestion
and performant access to all the features of a particular backend.
```

docs/reference/core_v3.rst
```diff
@@ -0,0 +1,122 @@
...
+
+If we do need to customize or tweak this process we can pass additional keyword
+arguments (`kwargs`) to overwrite any defaults ImageIO uses when reading or
+writing. The two `kwargs` that are always available are:
+
+- ``plugin``: The name of the plugin to use for reading/writing.
+- ``index`` (reading only): The index of the ndimage to read, if the format
+  supports storing more than one (GIF, multi-page TIFF, video, etc.).
+
+Additional `kwargs` are specific to the plugin being used and we can find a list
```
almarklein, line 60:
```suggestion
Additional `kwargs` are specific to the plugin being used and you can find a list
```

docs/reference/core_v3.rst
```diff
@@ -0,0 +1,122 @@
...
+        pass # do something with the current frame
+
+Just like imread, imiter accepts additional `kwargs` to overwrite any defaults
+used by ImageIO.
+
+Low-Level Access
+^^^^^^^^^^^^^^^^
+
+Sometimes we may wish for low-level access to a plugin or file, for example,
+because
```
almarklein, line 100:
```suggestion
because:
```

docs/reference/core_v3.rst
```diff
@@ -0,0 +1,122 @@
...
+
+- we wish to have fine-grained control over when it is opened/closed.
+- we need to perform multiple IO operations and don't want to open the file
+  multiple times.
+- a plugin/backend offers unique features not otherwise exposed by the v3 API.
+
+For these cases, the v3 API offers :func:`iio.v3.imopen
+<imageio.core.imopen.imopen>`. It provides a context managed that initializes
```
almarklein, line 108:
```suggestion
<imageio.core.imopen.imopen>`. It provides a context manager that initializes
```

docs/reference/core_v3.rst
```diff
@@ -0,0 +1,122 @@
...
+the plugin and openes the file for reading (``"r"``) or writing (``"w"``),
+similar to the built-in ``open``::
```
almarklein, line 110: Avoid confusion that "built-in" might refer to something built into imageio.
```suggestion
similar to the Python built-in function ``open``::
```

docs/reference/userapi.rst
```diff
@@ -70,3 +75,129 @@ that best suits that file-format.
...
+
+If you have an old script where you do not wish to migrate the API, you can
+explicitly import the old v2 API and avoid migration. Note that, while this will
+keep the old (v2) API, it will use on the new (v3) plugins and backends, which
+may contain other backwards incompatible changes. As such, you should prefer a
+full migration to v3, and should avoid using the v2 API in new code. To import
+the v2 API, depending on your script/code, use::
+
+    import imageio... as iio
+    import imageio... as imageio
```
almarklein, line 100: This is invalid syntax and looks confusing. Do you mean `import imageio.v2 as iio`?

docs/reference/userapi.rst
```diff
@@ -70,3 +75,129 @@ that best suits that file-format.
...
+``iio.mimwrite``, ``iio.volwrite``, and ``iio.mvolwrite`` have all dissapeared.
+The same goes for their aliases ``iio.mimsave``, ``iio.volsave``,
+``iio.mvolsave``, and ``iio.imsave``. They are now all covered by
+``iio.imwrite``.
+
+To reproduce similar behavior to the v2 API behavior use one of the following
+snippets::
+
+    # img = iio.v2.imwrite(image_resource, ndimage)
+    # img = iio.v2.imsave(image_resource, ndimage)
```
almarklein, line 143: Maybe don't mention the aliases here to avoid verbosity.

docs/reference/userapi.rst
```diff
@@ -70,3 +75,129 @@ that best suits that file-format.
...
+
+Metadata
+--------
+
+The v3 API makes handling of metadata much more consistent and explicit.
+Previously in v2, metadata was provided as a ``dict`` that was attached to the
+image by returning a custom subclass of ``np.ndarray``. Now metadata is
+provided separately from pixel data::
+
+    # metadata = iio.imread(image_resource)._meta
```
almarklein, line 191:
```suggestion
    # metadata = iio.imread(image_resource).meta
```

docs/reference/userapi.rst
```diff
@@ -70,3 +75,129 @@ that best suits that file-format.
...
+    metadata = iio.immeta(image_resource)
+
+Further, ImageIO now provides a curated set of standardized metadata which is
+called ImageProperties in addition to above metadata. The difference between the
+two is as follows: The metadata dict contains metadata using format-specific
+keys to the extent the reading plugin supports them. The ImageProperties
+dataclass contains metadata using standardized attribute names. They are the
+same name for all plugins, and - if the plugin or format doesn't provide a field
+- they are set to a (sensible) default value; usually ``None``. To access
```
almarklein, line 200:
```suggestion
dataclass contains metadata using standardized attribute names. Each plugin
provides the same set of properties, and if the plugin or format doesn't provide a field
it is set to a (sensible) default value. To access
```

almarklein:
I see that in most examples you now do something like
```py
import imageio as iio

iio.v3.xxxx(...)
```

That seems very verbose, and I don't think we want this to be the future way to use imageio, right? This discussion is very related to the deprecation path to adopt. Let's discuss in https://github.com/imageio/imageio/issues/725

docs/reference/core_v3.rst
FirefoxMetzger: Hm, do we want to use `we`, `you`, or passive tense in the docs? 

I tend to mix and match `we` and `you`, but should be consistent with one or the other. Personally, I dislike the "old academic style" of putting passive tense everywhere.

docs/reference/userapi.rst
FirefoxMetzger: Yes, I agree. This is still a TODO.  https://github.com/imageio/imageio/pull/720 is not yet merged, and its exact layout determines how this should be imported here.

Will fix it once we are clear on our transition policy.

docs/reference/userapi.rst
FirefoxMetzger: Will it be confusing for people that use the alias to not find it in the list? Or will us mentioning the aliases in the text above suffice?

docs/reference/core_v3.rst
almarklein: The passive tense can indeed become a bit "artificial". Otherwise it's a somewhat personal choice, but I find "we" a bit odd in docs like this.

docs/reference/userapi.rst
almarklein: I think they'll get the point because the pattern is quite clear, and we mention the aliases above.

docs/reference/core_v3.rst
FirefoxMetzger: Sure, then I will change the docs to use "you".

docs/reference/userapi.rst
FirefoxMetzger: I changed it to `import imageio.v2 as iio` for now. If we end up agreeing on something else in #720 I will update again.

FirefoxMetzger:
@almarklein RTR :) I think once this (and the pillow PR) are merged, we only need deprecation notices and then we are ready to move to v3 🎉

FirefoxMetzger:
@jni would you happen to have time to review this PR? It adds documentation for the ImageIO v3 API including the handling of metadata that we've been discussing a while back.

On top of the diffs in GitHub, we also build temporary docs for each PR, and you can find the ones for this PR here:
https://imageio--718.org.readthedocs.build/en/718/

docs/examples.rst
```diff
@@ -105,89 +112,93 @@ Grab screenshot or image from the clipboard
 Grab frames from your webcam
 ----------------------------
 
-Use the special ``<video0>`` uri to read frames from your webcam (via
-the ffmpeg plugin). You can replace the zero with another index in case
-you have multiple cameras attached. You need to ``pip install imageio-ffmpeg``
-in order to use this plugin.
+.. note::
+    For this to work, you need to install the ffmpeg backend::
```
almarklein, line 116: or pyav? ;)

docs/reference/userapi.rst
```diff
@@ -70,3 +76,129 @@ that best suits that file-format.
...
+Reader and Writer
+-----------------
+
+Plugins now longer have nested ``Reader`` and ``Writer`` classes and they are no
+longer global singletons. Instead, a new plugin object is instantiated in mode ``r``
+(reading) or  ``w`` (writing) for each `ImageResource` being opened.
+As such, the functions ``iio.get_reader``, ``iio.get_writer``, ``iio.read``, and
+``iio.save`` have become obsolete and the plugin object is used directly
+instead. Using these is considered advanced usage, happens inside a context
+manager, and uses `iio.imopen`::
```
almarklein, line 164: The mentioning of nested classes and singletons is not relevant for normal users, so let's omit that.

```suggestion
Previously, the `reader` and `writer` objects provided an advanced way to read/write data with more control. These are now removed. Likewise, the functions ``iio.get_reader``, ``iio.get_writer``, ``iio.read``, and
``iio.save`` have become obsolete.

Instead, the plugin object is used directly, which is instantiated in mode ``r`` (reading) or  ``w`` (writing).
To create a plugin object, call `iio.imopen` and use it as a context manager::
```

docs/reference/userapi.rst
```diff
@@ -70,3 +76,129 @@ that best suits that file-format.
...
+    metadata = iio.immeta(image_resource)
+
+Further, ImageIO now provides a curated set of standardized metadata, which is
+called :class:`ImageProperties <imageio.core.v3_plugin_api.ImageProperties>`, in
+addition to above metadata. The difference between the two is as follows: The
+metadata dict contains metadata using format-specific keys to the extent the
+reading plugin supports them. The ImageProperties dataclass contains generally
+available metadata using standardized attribute names. Each plugin provides the
+same set of properties, and if the plugin or format doesn't provide a field it
+is set to a (sensible) default value. To access ImageProperties use::
```
almarklein, line 202: or to None, still TBD

docs/reference/userapi.rst
```diff
@@ -70,3 +76,129 @@ that best suits that file-format.
...
+
+    # metadata = iio.imread(image_resource).meta
+    metadata = iio.immeta(image_resource)
+
+Further, ImageIO now provides a curated set of standardized metadata, which is
+called :class:`ImageProperties <imageio.core.v3_plugin_api.ImageProperties>`, in
+addition to above metadata. The difference between the two is as follows: The
+metadata dict contains metadata using format-specific keys to the extent the
+reading plugin supports them. The ImageProperties dataclass contains generally
+available metadata using standardized attribute names. Each plugin provides the
```
almarklein, line 200: Does this link work? Links in rst always confuse me.
```suggestion
available metadata using standardized attribute names (check the `ImageProperties docs <imageio.core.v3_plugin_api.ImageProperties>`_). Each plugin provides the
```

almarklein (commented):
Just some minor comments

docs/examples.rst
FirefoxMetzger: I would actually go with `imageio-ffmpeg` for now and then update once pyav is merged and had some time to prove itself.

docs/reference/userapi.rst
FirefoxMetzger: True. My thinking here was that "sensible default value" admits to the interpretation of: "it is always None if not provided", so we we are correct in either case x)

docs/reference/userapi.rst
FirefoxMetzger: Yes it does. Broken links trigger a warning in sphinx, and we have both a CI job that builds the docs (with the same settings as CD) as well as RTD building PR docs we can inspect. I've configured both to fail on warnings so that we can quickly spot things like broken links ... useful feature :D

docs/reference/userapi.rst
almarklein: Ok, but would still be good to mention that values can be `None` so the user can take that into account. If we go that route.

FirefoxMetzger: agreed. Does this mean we have a decision for https://github.com/imageio/imageio/issues/754? 

Either way, I have changed this around and added the phrase "; usually `None`". I think this is as close as we get until we make the definite decision to return `None` when the value is not set.

almarklein: Yes, I think it makes sense.

FirefoxMetzger: Perfect. That was the last open point in this review. Any other points or shall we merge?

almarklein:
🚀

PR #720, opened 2022-01-04
MAINT: dedicated namespaces for APIs

FirefoxMetzger:
This PR is a prelude to the v3 API docs. It

- ~refactors `core.imopen` into `core.v3_api` (since imopen is now a normal function and doesn't require it's own module anymore)~ (this will cause problems with circular imports. `imopen` is also core to all available APIs so we keep it separate.)
- adds the `imageio.v2` and moves all the v2 API functions into it.
- adds the `imageio.v3` and moves all the v3 API functions into it. This allows to `import imageio.v3 as iio` instead of the current `import imageio.core.v3_api as iio`. It also cleans up an inconsistency that snuck in during a previous commit: `imageio.v3` did exist before, however it was an attribute of `imageio` instead of being a dedicated module.
- adds `__all__` to the top-level `imageio` module to explicitly declare what the module exports.
- imports v2 and v3 API functions into top-level nainspace.
- [ ] adds a `imread` function to the top-level namespace that allows emitting a deprecation warning and eventual swapping of the function to v3.


imageio/__init__.py
```diff
@@ -19,47 +19,53 @@
...
 
-# Load all the plugins
-from . import plugins
 
 # expose the show method of formats
 show_formats = formats.show
 
-# Clean up some names
-del FormatManager
+__all__ = [
```
almarklein, line 44: Maybe this can be `v2.__all__ + ["v2", "v3", "config", "plugins"]

imageio/v3.py
```diff
@@ -0,0 +1,6 @@
+from .core.v3_api import imopen
+from .core.v3_api import imread
+from .core.v3_api import imwrite
+from .core.v3_api import imiter
+
+__all__ = ["imopen", "imread", "imwrite", "imiter"]
```
almarklein, line 6: Why not rename `v3_api` to `v3` and include the `__all__` there :)

almarklein (commented):
Nice! Just a few comments.

imageio/v3.py
FirefoxMetzger: Sure, we could do that. Alternatively, I can also move the functions themself from `core.v3_api` into `v3`. Which organizational structure do you think makes more sense?

almarklein: The end-result should be the same, right? :)

FirefoxMetzger: Looking at this now (about 1 month later), I think it actually makes sense to keep them separate or move `imopen` back into its own file.

Both the v2 and the v3 functions use `imopen` internally, so - in the light of the results of #725 - I am rethinking how to structure the library to accommodate this long term. Haven't settled for anything specific yet though.

FirefoxMetzger: I ended up moving the API functions into v3.py ; however, I have reintroduced `imopen.py` since it is a shared dependency between the v2 API and the v3 API. This now works nicely without any odd dependency problems.

FirefoxMetzger:
@almarklein From my side, I am done with this PR. Have a look if this reflects the import structure that you have understood from our discussion in #725 . If we are on the same page, let's merge this in.

From there the next steps are (in that order):

- https://github.com/imageio/imageio/pull/739
- https://github.com/imageio/imageio/pull/718
- add deprecation warnings
- https://github.com/imageio/imageio/pull/745

almarklein (approved):

imageio/__init__.py
```diff
@@ -49,17 +54,68 @@
...
+    "v3",
+    "config",
+    "plugins",
+    # v3 API
+    "imopen",
+    "imread",
+    "imwrite",
+    "imiter",
+    # v2 API
+    "imread",
```
FirefoxMetzger, line 100:
```suggestion
```

Remove duplicate entry (also in v3)

Issue #721, opened 2022-01-04
saving boolean images with pillow produces wrong results


FirefoxMetzger:
A colleague found some odd behavior in ImageIO:

```python
import numpy as np
import imageio as iio
from PIL import Image

# test image
foo = np.arange(255*255).reshape((255, 255)) % 2 == 0

# ensure pillow round-trip works
Image.fromarray(foo).save("pil.png")     
bar_pil = np.asarray(Image.open("pil.png"))
assert np.allclose(bar_pil, foo)

# imageIO v3 round trip
iio.v3.imwrite("iio.png", foo, mode="1") 
bar_v3 = iio.v3.imread("iio.png")
assert np.allclose(bar_v3, foo)  # Fails

# imageIO v2
iio.imwrite("iio.png", foo, pilmode="1")  # ValueError image not numeric
```

Given that the pillow round-trip works smoothly, I assume that this is related to our pillow plugin(s). Currently, I don't have a good solution. Once I know where the problem originates, I will submit a PR to address this. If, in the meantime, anyone else has a clue where things derail, I'm all ears :)

PR #722, opened 2022-01-05
FEAT: Allow mode=None when writing with pillow

FirefoxMetzger:
Closes: #721 

This PR refactors the v3 pillow plugin to rely on pillow's exposed modes (`PIL.ImageMode`) instead of maintaining our own list that needs updating whenever pillow changes something. After this refactor, we should automatically stay in sync.

It also adds the ability (new feature) to use `mode=None` when writing an ndimage with pillow. This will use pillow's logic to lookup an appropriate mode based on the ndimage's shape and dtype. It also changes the default mode to `None` to match
pillows `Image.save()`. (Although this is a tad experimental because I don't know if our current unit tests cover all the important scenarios)

As a side-effect this will address the problem reported in #721 which comes from some pillow weirdness:

```python
expected = np.arange(256 * 256).reshape((256, 256)) % 2 == 0

# saves the image correctly because the rawmode is determined by pillow
Image.fromarray(expected).save(tmp_path / "iio.png")

# saves a corrupt image, because mode="1" means 1-bit boolean (bit-packed) 
# whereas numpy uses 1-byte boolean (byte-packed).
# It appears to be impossible to specify mode=?? explicitly in this case.
Image.fromarray(expected, mode="1").save(tmp_path / "iio.png")
```

The PR also adds two tests to cover against regression for the fix of #721.

### DIFF
--- a/imageio/plugins/pillow.py
+++ b/imageio/plugins/pillow.py
@@ -57,32 +57,16 @@ def _is_multichannel(mode: str) -> bool:
 
     """
     multichannel = {
-        "1": False,
-        "L": False,
-        "P": False,
-        "RGB": True,
-        "RGBA": True,
-        "CMYK": True,
-        "YCbCr": True,
-        "LAB": True,
-        "HSV": True,
-        "I": False,
-        "F": False,
-        "LA": True,
-        "PA": True,
-        "RGBX": True,
-        "RGBa": True,
-        "La": False,
-        "I;16": False,
-        "I:16L": False,
-        "I;16N": False,
         "BGR;15": True,
         "BGR;16": True,
         "BGR;24": True,
         "BGR;32": True,
     }
 
-    return multichannel[mode]
+    if mode in multichannel:
+        return multichannel[mode]
+
+    return Image.getmodebands(mode) > 1
 
 
 def _exif_orientation_transform(orientation, mode):
@@ -255,7 +239,7 @@ class PillowPlugin(object):
 
         return image
 
-    def write(self, image, *, mode="RGB", format=None, **kwargs):
+    def write(self, image: np.ndarray, *, mode=None, format=None, **kwargs):
         """
         Write an ndimage to the URI specified in path.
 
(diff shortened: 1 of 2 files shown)


imageio/plugins/pillow.py
```diff
@@ -288,31 +273,54 @@ def write(self, image, *, mode="RGB", format=None, **kwargs):
...
-            image = image[None, ...]
-            save_all = False
+        if mode is None:
+            # there is no explicit function in pillow to determine the mode.
+            # However, pillow does it in Image.fromarray, and we imitate that
+            # here
+
+            # pillow supports 1 - 4 color channels depending on the format
+            # and all are channel-last
+            if image.shape[-1] in [2, 3, 4]:
```
almarklein, line 283: Prevent this `if` from being triggered on very narrow greyscale images:
```suggestion
            if image.ndim ==3 and image.shape[-1] in [2, 3, 4]:
```

imageio/plugins/pillow.py
```diff
@@ -288,31 +273,54 @@ def write(self, image, *, mode="RGB", format=None, **kwargs):
...
+                shape_key = (1, 1) + image.shape[-1:]
+            else:
+                shape_key = (1, 1)
+
+            try:
+                typekey = shape_key, image.dtype.str
+            except KeyError as e:
+                raise TypeError("Cannot handle this data type") from e
+            try:
+                lookup_mode, _ = Image._fromarray_typemap[typekey]
```
almarklein, line 293: This is a bit scary, because it uses a private function, which means that Pillow can change its name or behaviour at any time. I agree that the benefits are probably worth the risk, but a comment would be good.
```suggestion
                # Note: uses private function, so in theory this may break someday.
                lookup_mode, _ = Image._fromarray_typemap[typekey]
```

almarklein (commented):
LGTM, just a few minor points/suggestions.

imageio/plugins/pillow.py
FirefoxMetzger: Fair point regarding the use of a private function. Re-reading the code I think I can do a refactor to avoid using it.

almarklein (approved):

imageio/plugins/pillow.py
almarklein: LGTM!

Issue #719, opened 2022-01-03
Pillow v9.0.0 breaks test suite


FirefoxMetzger:
Pillow v9 introduced a change that broke our GIF reading test in the legacy pillow plugin:

https://github.com/imageio/imageio/blob/51f552ba5ed4aa771f205a2798e1fd3859fb2942/tests/test_pillow_legacy.py#L264-L333

It fails with:

```
============================= test session starts =============================
platform win32 -- Python 3.8.8, pytest-6.2.2, py-1.10.0, pluggy-0.13.1
rootdir: c:\Users\Sebastian\Documents\Coding-Projects\imageio
plugins: cov-2.11.1, localserver-0.5.0
collected 1 item

tests\test_pillow_legacy.py F                                            [100%]

================================== FAILURES ===================================
______________________________ test_animated_gif ______________________________

    def test_animated_gif():
    
        # Read newton's cradle
        ims = imageio.mimread("imageio:newtonscradle.gif")
        assert len(ims) == 36
        for im in ims:
>           assert im.shape == (150, 200, 4)
E           assert (150, 200, 3) == (150, 200, 4)
E             At index 2 diff: 3 != 4
E             Use -v to get the full diff

tests\test_pillow_legacy.py:270: AssertionError
- generated xml file: C:\Users\SEBAST~1\AppData\Local\Temp\tmp-22744GqVwFx95evBK.xml -
=========================== short test summary info ===========================
FAILED tests/test_pillow_legacy.py::test_animated_gif - assert (150, 200, 3) ...
============================== 1 failed in 0.46s ==============================
```

The new (v3) pillow plugin is unaffected.

At the time of this writing, I don't know what causes this problem; however, if I find the time this week, I will check how to resolve this.

Edit: Looking at the AssertionError, it appears that this is because pillow now returns a RGB image instead of returning a RGBA image.

FirefoxMetzger:
I did some digging. The change in pillow that introduced this problem is mentioned in their changelog here:

https://pillow.readthedocs.io/en/stable/releasenotes/9.0.0.html#convert-subsequent-gif-frames-to-rgb-or-rgba

I opened https://github.com/python-pillow/Pillow/issues/5929 to ask if this can be made more consistent and the way we should interact with this here depends a bit on the result of the discussion there. 

[...]

PR #724, opened 2022-01-07
BUG: handle pillow9 GIFs correctly in v2 pillow plugin

FirefoxMetzger:
Closes: #719 

I figured out a fix that doesn't involve a huge amount of changes, nor involves skipping pillow9. Not sure why I didn't come up with this one sooner.

I think https://github.com/imageio/imageio/pull/722 is actually affected by the pillow9 changes, too, but somehow we don't have a breaking test here. I will investigate and add a test as necessary.


almarklein:
Nice work!

PR #730, opened 2022-01-18
fix: add TiffFileError to the internal tifffile package

bojiang:
This barely cause any issues in general usage, but since `TiffFileError` is required [here](https://github.com/imageio/imageio/blob/master/imageio/plugins/tifffile.py#L367), it's better to include it in.

### DIFF
--- a/imageio/plugins/_tifffile.py
+++ b/imageio/plugins/_tifffile.py
@@ -4561,6 +4561,9 @@ class TiffTag(object):
         return line
 
 
+# Added to produce cleaner exceptions if tifffile unexpectedly fails to open the
+# file. See this comment (and the following) for details:
+# https://github.com/imageio/imageio/commit/bdbe699bbcda4223b0b6bd4d7474f84bbe34af09#r64068747
 class TiffFileError(ValueError):
     pass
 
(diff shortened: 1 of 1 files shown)


FirefoxMetzger:
@almarklein I decided to not add a unit-test here considering how small the change is. Further, our CI has already updated to installing tifffile instead of using the vendored one. As such, it will never encounter this problem anyway. Are you happy with this, too?

See here for a discussion of the problem that this PR solves: https://github.com/imageio/imageio/commit/bdbe699bbcda4223b0b6bd4d7474f84bbe34af09#r64068747

almarklein:
Should we perhaps update to the latest tiffile while we're at it?

almarklein (approved):

almarklein:
I'm not sure if its necessary to bother users with a deprecation warning. We could at least make tifffile a necessary dependency first, so that the warning is only shown in rare cases.

FirefoxMetzger:
Sure, I suppose this would be a dedicated PR then? 

@almarklein Do you think we should merge this as is then? (I don't know when I get the time to come back around to updating this.)

almarklein:
Yeah, let's merge it now.

almarklein:
Thanks @bojiang !

Issue #731, opened 2022-01-19
TIFF reader: ResolutionUnit, XResolution, YResolution not accessible


lfriedri:
Documentation states that TIFF resolution (pixel scaling) can be read:
https://imageio.readthedocs.io/en/stable/reference/_backends/imageio.plugins.tifffile.html#module-imageio.plugins.tifffile

However, resolution_unit is not available in meta data after reading.
Example:

```
import numpy as np
import imageio

filename = 'test.tif'
data = np.zeros((200, 100), dtype=N.uint8)
dpi = (100, 200)

writer = imageio.get_writer(filename)
writer.append_data(data, dict(resolution=dpi))
writer.close()

read = imageio.imread(filename)
print(read.meta)
```

It seems that tifffile provides only certain tags as attributes:
https://github.com/cgohlke/tifffile/blob/master/tifffile/tifffile.py#:~:text=def%20TAG_ATTRIBUTES()%3A,%7D

imageio on the other hand only provides those tags, that are attributes:
https://github.com/imageio/imageio/blob/master/imageio/plugins/tifffile.py#:~:text=def%20_get_meta_data(self,pass

Additionally, in the list of the reading tags in imageio, "resolution_unit" appears without comma, is that intended?
https://github.com/imageio/imageio/blob/master/imageio/plugins/tifffile.py#:~:text=READ_METADATA_KEYS%20%3D%20(,)

lfriedri:
Note that the resolution of the generated 'test.tif' can be successfully read with ImageJ.

FirefoxMetzger:
@lfriedri This should get fixed in next monday's release. You should then be able to access the resolution under the metadata key "resolution".

lfriedri:
thank you, that was fast. I will try to test the release in this regard.

PR #732, opened 2022-01-19
FEAT: Make tifffile read `resolution` metadata

FirefoxMetzger:
Closes: #731 

This PR implements the ability for tifffile to report the `resolution_unit` tag. This was documented before but appears to have not been implemented. I took the liberty of changing its behavior to match the behavior of the ResolutionUnit tag in the TIFF spec.

Further, this PR implements a new metadata item called `resolution` that matches the behavior of the `resolution` metadata item in write. aka: the ability to read dpi/dpc if specified in the TIFF.


imageio/plugins/tifffile.py
```diff
@@ -42,8 +42,20 @@
...
     'contig': last dimension contains samples.
     'planar': third last dimension contains samples.
-resolution_unit : (float, float) or ((int, int), (int, int))
-    X and Y resolution in dots per inch as float or rational numbers.
+resolution_unit : int
+    The resolution unit stored in the TIFF tag. Usually 1 means no/unknown unit,
+    2 means dpi (inch), 3 means dpc (centimeter).
+resolution : ((int, int), (int, int), str)
+    A tuple formatted as (X_resolution, Y_resolution, unit). Both resolutions
+    are given as rational numbers, i.e. as a (numerator, denominator) tuple. Unit
```
almarklein, line 50: Would it make more sense to return these as floats, or as [real rational numbers](https://docs.python.org/3/library/fractions.html)?

tests/test_tifffile.py
```diff
@@ -144,6 +144,34 @@ def test_imagej_hyperstack(tmp_path):
...
+    dpi = (100, 200)
+
+    writer = imageio.get_writer(tmp_path / "test.tif")
+    writer.append_data(data, dict(resolution=dpi))
+    writer.close()
+
+    read_image = iio.imread(tmp_path / "test.tif")
+
+    assert read_image.meta["resolution"] == (*dpi, "INCH")
+    assert read_image.meta["resolution_unit"] == 2
```
almarklein, line 160: I'm confused, here the numbers are not rationals?

FirefoxMetzger: True. Yesterday, it seemed like a good idea to reduce a rational like `(200, 1)` to `200`, i.e., if the denominator is 1, drop it. Today, I'm not so sure anymore how good this idea actually is, considering that it makes the return type inconsistent.

What do you think?

almarklein: I think we should at least be consistent. So either we always return `(int, int)` or we always return `float`. My gut says floats but I also see the value in just passing what tifffile returns. Your call :D

almarklein: I cannot imagine that the precision matters *that* much. I think I'm leaning towards floats too; the int-tuple is confusing.

FirefoxMetzger: lets go with `float` then. I made the necessary changes.

almarklein (approved):

Issue #686, opened 2021-11-17
make ffmpeg's can_read more sophisticated


FirefoxMetzger:
Details in https://github.com/imageio/imageio/pull/680#issuecomment-970142963

Currently, there is a discrepancy between what our FFmpeg plugin can actually read and what it claims it can read. It claims to be able to read less than it actually does. The cause of this is that `_can_read` makes a decision based on the file's extension alone instead of querying the file's metadata/header (e.g., via ffprobe). This is problematic because it will reject reading from `bytes`/`BytesIO` or reading any format that we have not manually registered.

Kulikovpavel:
Temporary fix
```
from imageio.plugins.ffmpeg import FfmpegFormat 
FfmpegFormat.can_read = lambda x, y: True
```

JacksonMaxfield:
Bumping this after your comment @FirefoxMetzger!

This also affects us over at `aicsimageio`. Until one of our own users requires a change it is a lower priority for us but will update this issue if something comes up.

FirefoxMetzger:
For future readers, you can use the new (v2.15.0) `format_hint=` kwarg in v3 to help the FFmpeg backend and read the bytes/BytesIO:

```python
import imageio as iio

frame = iio.v3.imread(video_bytes, format_hint=".mkv")
frame = iio.v3.imread(video_bytesio, format_hint=".mp4")
```

On older versions of ImageIO, a workaround is to set the extension for the stream manually. Note, however, that this is more of a hack, because there is no guarantee that the (private) `_extension` will be around forever:

```python
import imageio as iio

req = iio.core.Request(video_bytes, "r")
req._extension = ".mkv"  # your format extension here
frame = iio.imread(req)
```

PR #734, opened 2022-01-24
FEAT: add format_hint to v3 API

FirefoxMetzger:
I didn't create an explicit issue for this PR, because I thought demonstrating it is easier than describing it.

The PR adds a new keyword argument (called `format_hint`) to `imopen` (and subsequentially to `imread`, `imwrite`, and `imiter` in the v3 API. It allows a new fast path for ImageResources without an explicit extension (e.g., streams, bytes and FileLike Objects) or for files for which the extension doesn't match the actual content of the file. It works pretty much like the fast path we have for file extensions, i.e., try plugins that are associated with that extension first. Hints are provided using the format's file extension, e.g. `imread(something, format_hint=".jpg")`.

In addition, it will overwrite/set the extension of a Request if the extension for that request was previously `None`. This is useful to elevate some of the pains tracked in #686. With this, ffmpeg streams can now be read via

```python
# this is a video as bytes
video_blob = Path("path/to/video.mp4).read_bytes()

# raises IOError, because FFMPEG's can_read is too naive
# ndimage = imageio.v3.imread(video_blob)

# with the hint, FFMPEG can read blobs directly again
ndimage = imageio.v3.imread(video_blob, format_hint=".mp4")
```

This is, ofc, only a workaround, but I think its already much better than not being able to read them anymore.

### DIFF
--- a/imageio/core/imopen.py
+++ b/imageio/core/imopen.py
@@ -204,7 +204,7 @@ def imopen(
         raise err_type(err_msg) from err_from
 
     # fast-path based on format_hint
-    if format_hint is not None:
+    if request.format_hint is not None:
         for candidate_format in known_extensions[format_hint]:
             for plugin_name in candidate_format.priority:
                 config = known_plugins[plugin_name]
--- a/imageio/core/imopen.py
+++ b/imageio/core/imopen.py
@@ -227,7 +227,9 @@ def imopen(
 
                 return plugin_instance
         else:
-            raise IOError(f"The ImageResource can not be opened as a `{format_hint}` file.")
+            raise IOError(
+                f"The ImageResource can not be opened as a `{format_hint}` file."
+            )
 
     # fast-path based on file extension
     if request.extension in known_extensions:
--- a/imageio/core/imopen.py
+++ b/imageio/core/imopen.py
@@ -66,6 +66,7 @@ def imopen(
     io_mode: str,
     *,
     plugin: Union[str, Any] = None,
+    format_hint: str = None,
     legacy_mode: bool = True,
     **kwargs,
 ) -> Any:
@@ -100,7 +101,15 @@ def imopen(
 
     plugin : str, Plugin, or None
         The plugin to use. If set to None (default) imopen will perform a
-        search for a matching plugin.
+        search for a matching plugin. If not None, this takes priority over
+        the provided format hint.
+    format_hint : str
+        A format hint to help optimize plugin selection given as the format's
+        extension, e.g. ".png". This can speed up the selection process for
+        ImageResources that don't have an explicit extension, e.g. streams, or
+        for ImageResources where the extension does not match the resource's
+        content. If the ImageResource lacks an explicit extension, it will be
+        set to this format.
     legacy_mode : bool
         If true (default) use the v2 behavior when searching for a suitable
         plugin. This will ignore v3 plugins and will check ``plugin``
@@ -117,6 +126,13 @@ def imopen(
     Passing a ``Request`` as the uri is only supported if ``legacy_mode``
     is ``True``. In this case ``io_mode`` is ignored.
 
+    Using the kwarg ``format_hint`` does not enforce the given format. It merely
+    provides a `hint` to the selection process and plugin. The selection
+    processes uses this hint for optimization; however, a plugin's decision how
+    to read a ImageResource will - typically - still be based on the content of
+    the resource.
+
+
     Examples
     --------
 
@@ -129,16 +145,23 @@ def imopen(
 
     """
 
+    if format_hint is not None and format_hint[0] != ".":
+        raise ValueError(
+            "`format_hint` should be a file extension starting with a `.`,"
+            f" but is `{format_hint}`."
+        )
+
     if isinstance(uri, Request) and legacy_mode:
         request = uri
         uri = request.raw_uri
         io_mode = request.mode.io_mode
+        request.format_hint = format_hint
     else:
-        request = Request(uri, io_mode)
+        request = Request(uri, io_mode, format_hint=format_hint)
 
     source = "<bytes>" if isinstance(uri, bytes) else uri
 
-    # plugin specified, no search needed
+    # fast-path based on plugin
     # (except in legacy mode)
     if plugin is not None:
         if isinstance(plugin, str):
@@ -180,6 +203,32 @@ def imopen(
         request.finish()
         raise err_type(err_msg) from err_from
 
+    # fast-path based on format_hint
+    if format_hint is not None:
+        for candidate_format in known_extensions[format_hint]:
+            for plugin_name in candidate_format.priority:
+                config = known_plugins[plugin_name]
+
+                # v2 compatibility; delete in v3
+                if legacy_mode and not config.is_legacy:
+                    continue
+
+                try:
+                    candidate_plugin = config.plugin_class
+                except ImportError:
+                    # not installed
+                    continue
+
+                try:
+                    plugin_instance = candidate_plugin(request, **kwargs)
+                except InitializationError:
+                    # file extension doesn't match file type
+                    continue
+
+                return plugin_instance
+        else:
+            raise IOError(f"The ImageResource can not be opened as a `{format_hint}` file.")
+
     # fast-path based on file extension
     if request.extension in known_extensions:
         for candidate_format in known_extensions[request.extension]:
@@ -205,7 +254,7 @@ def imopen(
                 return plugin_instance
 
     # error out for read-only special targets
-    # this is again hacky; can we come up with a better solution for this?
+    # this is hacky; can we come up with a better solution for this?
     if request.mode.io_mode == IOMode.write:
         if isinstance(uri, str) and uri.startswith(SPECIAL_READ_URIS):
             err_type = ValueError if legacy_mode else IOError
--- a/imageio/core/imopen.py
+++ b/imageio/core/imopen.py
@@ -11,6 +11,7 @@ from .request import (
 from ..config.plugins import PluginConfig
 from ..config import known_plugins
 from ..config.extensions import known_extensions
+import warnings
 
 
 def _get_config(plugin: str, legacy_mode: bool) -> PluginConfig:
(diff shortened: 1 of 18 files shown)


imageio/core/imopen.py
```diff
@@ -100,7 +101,15 @@ def imopen(
...
-        search for a matching plugin.
+        search for a matching plugin. If not None, this takes priority over
+        the provided format hint.
+    format_hint : str
+        A format hint to help optimize plugin selection given as the format's
+        extension, e.g. ".png". This can speed up the selection process for
+        ImageResources that don't have an explicit extension, e.g. streams, or
+        for ImageResources where the extension does not match the resource's
+        content. If the ImageResource lacks an explicit extension, it will be
+        set to this format.
```
almarklein, line 112: What about if `plugin` starts with a dot, it means an extension, and thus a format hint? This is how it worked previously.

FirefoxMetzger: > What about if plugin starts with a dot, it means an extension, and thus a format hint?

Yes. I separated them, because plugins and formats are different things. This is ultimately how we ended up with hybrid format names like `PNG-FI`, `PNG-PIL`, but also with format names (that are actually plugins) `FFMPEG` or `SPE`. Separating them into `plugin=` (only plugin names allowed here) and `format=` (only formats allowed here) makes this cleaner. 

> This is how it worked previously.

Indeed, this is something we introduced to support the v2 API through imopen. It won't work in the v3 API, where the value of `plugins` is checked against ... well plugins (`plugin in config.known_plugins`). If we do end up merging this, I will refactor the internals of the v2 API to separate plugins and formats and send them to the appropriate arguments. This will keep the behavior local to the v2 API and allow us to have a cleaner backend.

What's more is that calling `PNG-PIL` doesn't actually guarantee that the data is treated as a PNG. The v3 pillow plugin currently has no notion of format (it hands a file-like object to pillow and, if that fails, concludes that pillow can't read it). The v2 plugin does have a notion of format (enforced via file extension), but if that fails (e.g. for bytes that have no extension) it also just does the same thing as the v3 plugin: call the backend and see if things fail. Notably, neither plugin restricts the formats pillow may use internally (this could be done though), so at the moment iio.imread(some-jpg-bytes, format="PNG-PIL") will look like it loads a PNG from a JPEG, but it actually just uses pillows JPEG decoder and ignores the `format=` part silently. Because of this, the format is a _hint_ and not enforced. 

I felt like this should be documented as such, in an effort to be more transparent and, while at it, I figured it would be useful to have a similar functionality in v3.

Brownie points: we no longer have to figure out which format supports which extension and create strange virtual formats like `PNG-PIL` and alike. `plugins` is everything listed in `config.known_plugins` and `format_hint` is everything listed in `config.known_extensions` and the rest is mix and match. If the format being hinted is actually not what is in the file, that's totally fine; it's a hint anyway, and all it does is make the selection process slower, because we try unsuitable plugins first due to the wrong hint.

almarklein: Good point. I was a bit reluctant about making the API more complex by introducing a new argument. My initial reaction was that we can still use a single argument, that is a format hint if it's a string starting with a dot, and a plugin selection otherwise, but indeed this becomes a bit weird, since the former is a hint and the latter a hard selection.

almarklein (approved):

imageio/core/imopen.py
```diff
@@ -180,6 +203,34 @@ def loader(request, **kwargs):
...
+                    plugin_instance = candidate_plugin(request, **kwargs)
+                except InitializationError:
+                    # file extension doesn't match file type
+                    continue
+
+                return plugin_instance
+        else:
+            raise IOError(
+                f"The ImageResource can not be opened as a `{format_hint}` file."
+            )
```
FirefoxMetzger, line 232: Actually, as it is now, the PR would only check plugins that are known to read the hinted format and fail otherwise. We can remove this part (or make it a warning) which would fall back to checking all plugins if the expected plugins don't work out.

@almarklein Which of these two approaches do you think is cleaner?

imageio/core/v3_api.py
```diff
@@ -24,7 +26,16 @@ def imread(uri, *, index: int = None, plugin: str = None, **kwargs) -> np.ndarra
         If the URI contains multiple ndimages, select the index-th ndimage
         from among them and return it. The exact behavior is plugin dependent.
     plugin : {str, None}
-        The plugin to be used. If None, performs a search for a matching plugin.
+        The plugin to use. If set to None (default) imopen will perform a
```
FirefoxMetzger, line 29:
```suggestion
        The plugin to use. If set to None (default) imread will perform a
```

imageio/core/v3_api.py
```diff
@@ -24,7 +26,16 @@ def imread(uri, *, index: int = None, plugin: str = None, **kwargs) -> np.ndarra
...
+        search for a matching plugin. If not None, this takes priority over
+        the provided format hint.
```
FirefoxMetzger, line 31:
```suggestion
        the provided format hint  (if present).
```

imageio/core/v3_api.py
```diff
@@ -24,7 +26,16 @@ def imread(uri, *, index: int = None, plugin: str = None, **kwargs) -> np.ndarra
...
+    format_hint : str
+        A format hint to help optimize plugin selection given as the format's
+        extension, e.g. ".png". This can speed up the selection process for
+        ImageResources that don't have an explicit extension, e.g. streams, or
+        for ImageResources where the extension does not match the resource's
+        content. If the ImageResource lacks an explicit extension, it will be
+        set to this format.
```
FirefoxMetzger, line 38:
```suggestion
        content.
```

imageio/core/v3_api.py
```diff
@@ -54,8 +67,16 @@ def imiter(uri, *, plugin: str = None, **kwargs) -> Iterator[np.ndarray]:
         The resource to load the image from, e.g. a filename, pathlib.Path,
         http address or file object, see the docs for more info.
     plugin : {str, None}
-        The plugin to be used. If None, performs a search for a matching
-        plugin.
+        The plugin to use. If set to None (default) imopen will perform a
```
FirefoxMetzger, line 70:
```suggestion
        The plugin to use. If set to None (default) imiter will perform a
```

imageio/core/v3_api.py
```diff
@@ -54,8 +67,16 @@ def imiter(uri, *, plugin: str = None, **kwargs) -> Iterator[np.ndarray]:
...
+        search for a matching plugin. If not None, this takes priority over
+        the provided format hint.
```
FirefoxMetzger, line 72:
```suggestion
        the provided format hint (if present).
```

imageio/core/v3_api.py
```diff
@@ -90,8 +113,16 @@ def imwrite(uri, image: np.ndarray, *, plugin: str = None, **kwargs) -> Optional
     image : np.ndarray
         The image to write to disk.
     plugin : {str, None}
-        The plugin to be used. If None, performs a search for a matching
-        plugin.
+        The plugin to use. If set to None (default) imopen will perform a
```
FirefoxMetzger, line 116:
```suggestion
        The plugin to use. If set to None (default) imwrite will perform a
```

imageio/core/v3_api.py
```diff
@@ -90,8 +113,16 @@ def imwrite(uri, image: np.ndarray, *, plugin: str = None, **kwargs) -> Optional
...
+        search for a matching plugin. If not None, this takes priority over
+        the provided format hint.
```
FirefoxMetzger, line 118:
```suggestion
        the provided format hint (if present).
```

imageio/core/imopen.py
almarklein: I'm not sure what you mean, but I'd expect this:
* if plugin is given, that is the only plugin that will be used, and will fail otherwise.
* if format-hint is given, it will try to find a plugin that satisfies that format, and if it fails, will try harder, with a warning,
* if both are given, only the selected plugin is used, and again the format is a hint.

But why should the format be a hint? What about:
* if plugin is given, that is the only plugin that will be used, and will fail otherwise.
* if format is given, it will use a plugin that can read that format, and it will be read as that format, or it will fail.
* if both are given, it *has* to be the given format *and* plugin.

FirefoxMetzger:
Looks like there is a new black version and that it reformats a bunch of files. I'll sneak those updates into this PR.

imageio/core/imopen.py
FirefoxMetzger: > But why should the format be a hint? [...] and it will be read as that format, or it will fail.

It is currently a hint because we would struggle to enforce/guarantee that the ImageResource is indeed read as the desired format. 

From an interface design standpoint we can't control (or assert) if a backend actually opens the file using the desired format. Worse, some backends may not support enforcing a format, and then the plugin writer is stuck between a rock and a hard place, because the ImageIO API requires that you can force a format, but the backend doesn't permit this easily. The `plugin` argument on the other hand, we can (and do) enforce ourselves, so for that one we can give the user a guarantee that it will be that plugin or nothing.

From a practical perspective, you might also not get too much out of enforcing that a ImageResource is treated as a certain format. E.g., just because you enforce that the resource is treated as TIFF doesn't tell you anything about the data you will get (colorspace, shape, ndims, planar/interleaved aka. channel_axis). You'd have similar problems with videos (almost any container/format supports multiple codecs and pixel formats/colorspaces) and PNG (supports multiple colorspaces). Not sure about other formats. What I'm trying to say is that I'm not entirely sure what a user may gain from enforcing a format.

Further, if you do have a file with the wrong extension (and you don't know which one it should be/ what's inside the file) then ImageIO will already take care of that during fallback. It will first try the plugins associated with that extension and, if that fails, it will do a broader search considering all available plugins (which will read almost anything under the sun at this point).

> if format-hint is given, it will try to find a plugin that satisfies that format, and if it fails, will try harder, with a warning,

Okay, then I will change it to a warning (the other points are as you listed in what you'd expect). 

We can discuss the utility and desired behavior of enforcing a format further, and - based on the outcome - I will make further adjustments as needed.

almarklein: Ok, lets move on with that direction then.

FirefoxMetzger:
Okay, this should apply the changes as discussed.

@almarklein Will merge once CI is happy (minus PR coverage, because the black reformatting touches uncovered lines).

PR #736, opened 2022-01-25
Simplify initial image downloading

almarklein:
The approach that used Git proved to be fragile, and made it impossible for me to run tests on my Win11 machine.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,4 +1,3 @@
-import platform
 import pytest
 import os
 import shutil
@@ -6,7 +5,6 @@ from pathlib import Path
 import contextlib
 
 import imageio as iio
-import requests
 
 
 @contextlib.contextmanager
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,6 +1,7 @@
 import pytest
 import os
 import shutil
+import json
 
 import imageio as iio
 import requests
@@ -28,6 +29,9 @@ def tmp_dir(tmp_path_factory):
 
         # Download the images
         for image_info in image_info_dicts:
+            if isinstance(image_info, str):
+                print(info)
+                image_info = json.loads(image_info)
             filename = tmp_path / image_info["name"]
             r = requests.get(image_info["download_url"])
             with open(filename, "bw") as f:
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -29,11 +29,11 @@ def tmp_dir(tmp_path_factory):
             image_info_dicts = list(image_info_dicts.values())
 
         # Download the images
+        downloader = requests.Session()
         for image_info in image_info_dicts:
-            filename = tmp_path / image_info["name"]
-            r = requests.get(image_info["download_url"])
-            with open(filename, "bw") as f:
-                f.write(r.content)
+            response = downloader.get(image_info["download_url"])
+            response.raise_for_status()
+            (tmp_path / image_info["name"]).write_bytes(response.content)
 
     return tmp_path_factory.getbasetemp()
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,3 +1,4 @@
+import platform
 import pytest
 import os
 import shutil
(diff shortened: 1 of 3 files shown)


almarklein:
Why the hell does `r.json()` produce something different on MacOS? I checked on my Mac and it works fine there ...

tests/conftest.py
```diff
@@ -13,29 +13,27 @@ def tmp_dir(tmp_path_factory):
...
-        os.system("git sparse-checkout set test-images")
-    else:
-        # sparse checkout on GH Actions
-        os.system(
-            "git clone --sparse --filter=blob:none https://github.com/imageio/imageio-binaries.git ."
-        )
-        os.system("git sparse-checkout init --cone")
-        os.system("git sparse-checkout set test-images")
-    os.chdir(current_path)
+    if not os.path.isfile(tmp_path / "chelsea.png"):
```
FirefoxMetzger, line 16: I don't think this branch is necessary. This fixture is run once at the beginning of a test session (session scoped) and tmp_path will be a clean  directory.

tests/conftest.py
```diff
@@ -13,29 +13,27 @@ def tmp_dir(tmp_path_factory):
...
+        # We only need to download once (per test session)
+
+        # Get list of images. authenticate with our GH token so that the rate limit is per-repo
+        api_endpoint = "https://api.github.com/repos/imageio/imageio-binaries"
+        token = os.getenv("GITHUB_TOKEN")
+        headers = {}
+        if token:
+            headers["Authorization"] = f"token {token}"
+        r = requests.get(api_endpoint + "/contents/test-images", headers=headers)
+        image_info_dicts = r.json()
```
FirefoxMetzger, line 27:
```suggestion
        r = requests.get(api_endpoint + "/contents/test-images", headers=headers)
        r.raise_for_status()
        image_info_dicts = r.json()
```

Let's make sure the request succeeded :)

tests/conftest.py
```diff
@@ -13,29 +13,27 @@ def tmp_dir(tmp_path_factory):
...
+        if isinstance(image_info_dicts, dict):
+            image_info_dicts = list(image_info_dicts.values())
```
FirefoxMetzger, line 29: I believe this branch, too, will never happen since we will always have more than 1 test image.

tests/conftest.py
```diff
@@ -13,29 +13,27 @@ def tmp_dir(tmp_path_factory):
...
+
+        # Download the images
+        for image_info in image_info_dicts:
+            filename = tmp_path / image_info["name"]
+            r = requests.get(image_info["download_url"])
+            with open(filename, "bw") as f:
+                f.write(r.content)
```
FirefoxMetzger, line 36:
```suggestion
        # Download the images
        downloader = requests.Session()
        for image_info in image_info_dicts:
            response = downloader.get(image_info["download_url"])
            response.raise_for_status()
            (tmp_path / image_info["name"]).write_bytes(response.content)
```

I would refactor like this. A `Session` will keep the connection alive between request and avoid the overhead of re-establishing a connection for each download. Also, `Path` has a convenience wrappers so we can avoid some boilerplate. Here `write_bytes`.

FirefoxMetzger (commented):
I like the change from calling git under the hood (way too fragile) to using the GitHub API. I will also have a look into why things fail on Mac and see if I have any insight.

tests/conftest.py
almarklein: I did see it run more often once, but maybe that was because I put a `1/0` in it. Does it make sense that it will then try it again and again?

tests/conftest.py
almarklein: Oh, that's probably it :D

tests/conftest.py
almarklein: I figured that maybe it was encoded as `{1: {...}, 2:{...}, 3:{...}}`, but anyway, this was me being desperate.

tests/conftest.py
FirefoxMetzger: I think so. I never dug into it explicitly, but as far as I understand, a cache entry for a fixture is only cached if it executes successfully. If it fails (e.g. because of a div/0 exception) then it should be run repeatedly since there is no cache entry to reuse.

tests/conftest.py
FirefoxMetzger:
```suggestion
```

Then let's remove it :)

FirefoxMetzger:
@almarklein No idea what was the cause exactly, but my crude fix appears to make macos happy :)

FirefoxMetzger:
Yep. Here is the reason for our failures:

```
"requests.exceptions.HTTPError: 403 Client Error: rate limit exceeded for url: https://api.github.com/repos/imageio/imageio-binaries/contents/test-images"
```

FirefoxMetzger:
@almarklein I did some googling, and it turns out that the rate limit for the GH API is quite low, so this could be one reason why we run into rate limits. Even if it currently isn't, it's not very scalable if we have multiple PRs running CI at the same time.

So I reverted to `git clone`, but with a twist: I created a new repo at `imageio/test_images`, which contains test images only. For this, we don't need a fancy sparse checkout dance, but can do plain old `git clone <repo>`. I then migrated one of @anjos changes (a `working_directory` context manager) and added the persistent caching mechanism that I was proposing in #705 .

Long story short, I changed the fixture so that it neither needs sparse checkout nor the GitHub API. Further, (adding some sprinkles) test images will now be permanently stored in the project's folder (under `project_root/.test_images`) and reused for each test, saving the need to re-download.

tests/conftest.py
```diff
@@ -1,59 +1,90 @@
...
-            shutil.copy(item, data_dir / item.name)
+def image_files(test_images, tmp_path):
+    """A copy of the test images
 
-    yield data_dir
+    This is intended for tests that write to or modify the test images.
+    """
+    for item in test_images.iterdir():
+        if item.is_file():
+            shutil.copy(item, tmp_path / item.name)
```
almarklein, line 85: Out of curiosity, why are the images also copied to a temp folder? And are these temp folders cleared automaticlaly?

FirefoxMetzger: > And are these temp folders cleared automaticlaly?

Yes, `tmp_path` is a fixture provided by pytest itself which creates a new, empty temporary directory for you to mess around in.

> Out of curiosity, why are the images also copied to a temp folder?

Because while I was writing this I was, in my infinite wisdom, concerned that the tests might end up modifying the images and may end up corrupting the persistent cache so I thought it would be wise to make a defensive copy.

Now (6 months down the line) I kind of regret that decision and it is on my list of "let's refactor this at some point". Maybe we can again rely on git here and defensively revert to the test_images repo at the beginning of a test run; I haven't deeply thought into how to refactor in a good way yet.

FirefoxMetzger (approved):

almarklein:
Thanks for helping out!

I tested on MacOS: all good.

Tested on Win11: it hangs, then I interrupt, try again, dir is corrupt, so I delete it, try again, goto start :)  Sometimes it works, and from then on it is fine. I'm not sure whether the hang is just the cloning being slow, or whether it's a proper hang.

At least this is an improvement, so let's merge this.

FirefoxMetzger:
It is cloning being slow ... we have >100MB of image data in the repo and git isn't optimized for downloading blobs of that size :D

At some point, we could do with some blob storage where this data should move to. No idea if this is easy to come by for OOS projects...

almarklein:
> It is cloning being slow

Yeah, probably.

PR #737, opened 2022-01-25
Replace imread and imwrite with new style versions

almarklein:
This is a test to see what fails when we just replace imread and imwrite with their new versions. I wonder whether we can "catch" the cases that now break ...


FirefoxMetzger:
The main purpose was to test where we are backward-incompatible (which appears to be limited to imread only). @almarklein If there is no other use for this PR, should we close it to keep things clean?

Issue #610, opened 2021-03-14
Wrap python av 


FirefoxMetzger:
[av](https://pyav.org/docs/develop/#) is a python library that wraps ffmpeg and aims to provide pythonic access to it. It apprears quite similar to our `imageio-ffmpeg`, so it could be useful to create a wrapper for it instead of maintaining our own wrapper.

Of course, both could exist in parallel instead of being a replacement.

almarklein:
> Due to the complexity of the dependencies, PyAV is not always the easiest Python package to install from source. Since release 8.0.0 binary wheels are provided on PyPI for Linux, Mac and Windows linked against a modern FFmpeg. 

One of the main reasons for building imagio-ffmpeg instead of using pyav was that pyav was not trivial to install. It looks like this has changed.

Issue #220, opened 2017-02-06
Animated gif writer via ffmpeg


almarklein:
As mentioned in #219, ffmpeg can also export animated gif (ffmpeg 2.6+). Would be a nice addition. http://blog.pkh.me/p/21-high-quality-gif-with-ffmpeg.html

Advantages are speed and maybe quality, we'd have to do some testing to see how the paletting related to PIL's paletting. Disadvantage is that it needs the ffmpeg exe, which is not available by default (though easy enough to install using one function call).

almarklein:
Also see #115

FirefoxMetzger:
I submitted a PR to pyAV a few days back that allows sending EOF to a filter, so we will be able to use `palettegen` with the next pyAV release. From this point on it should be possible to generate GIF via ffmpeg, too.

FirefoxMetzger:
pyAV released v9 recently which includes my changes to send EOF to filters (among other goodies). I've updated our pyAV plugin, and now we can indeed create GIF using ffmpeg:

![test](https://user-images.githubusercontent.com/4402489/157633664-ccf0351f-6745-4901-bba7-f4ead280060e.gif)

(at the moment it's grayscale only until I work out how to handle frames with palette in pyav)

FirefoxMetzger:
And now we can do colored GIFs, too:

![test](https://user-images.githubusercontent.com/4402489/161819196-5354b320-949d-47df-b58f-528c7298b80b.gif)

Issue #504, opened 2020-01-28
How should imageio deal with colorspaces?


almarklein:
I think that at the moment, imageio generally returns either grayscale, RGB, or RGBA. Though I am less certain about Tiff files. 

So when imageio encounters data that is not RGB (e.g. HSV) what should it do?

* Return as RGB/RGBA. 
* Return as original colorspace, and make user aware via flag in meta data (see #382).

My inclination is to go with the former. That way, you can read data without knowing the source and work with it. For more control, we can consider adding an argument to the read functions to return another colorspace or the original colorspace.

ivoflipse:
How does OpenCV handle this?

FirefoxMetzger:
Looking at this 2 years later, we seem to have settled for option two.

Currently, plugins return the data in the native colorspace by default since this is the least surprising/assuming. Additionally, we offer a kwarg (`format` for pyav, `mode` for pillow, ...) that allows users to specify a colorspace that the image will be converted into if it isn't in that format already.

@almarklein Any objections to me closing this issue?

almarklein:
Might still be worth explicitly mentioning this in the docs. What do you think?

almarklein:
Actually, I think I am still in favor of returning RGB/RGBA by default, and provide an option to return as it is stored. One of the principals of imageio has been that you read any format and you get back a numpy array in the expected shape that you can just pass to a visualization tool.

FirefoxMetzger:
> One of the principals of imageio has been that you read any format and you get back a numpy array in the expected shape that you can just pass to a visualization tool.

Yes, I am aware and I think it is a nice idea to convert into a standard colorspace. Unfortunately, I think it only works as a first-order approximation, because there are quite a few situations where this breaks down:

- What about (binary) mask images? The expectation would be to get a boolean mask, and plugins like pillow indeed return a `bool` array. We could promote to RGB(A), but I would find that quite surprising. Maybe we want to make an exception for these.
- What about grayscale images? I think users would be quite surprised to get a color channel or have it promoted to RGB(A) by default. We could, again, make an exception for this colorspace.
- What about ImageJ hyperstacks? Those are stored channel-first and have usually no obvious relationship to RGB unless they happen to represent staining that uses those colors and that has planes in that order. I'm not sure there is a natural/ intuitive way to convert this to RGB. Again we could (must?) make an exception for these images.
[...]

PR #745, opened 2022-01-31
FEAT: New Plugin for pyAV

FirefoxMetzger:
Closes: #610, https://github.com/imageio/imageio/issues/220, https://github.com/imageio/imageio/issues/686

This PR adds a pyAV plugin to ImageIO using the v3 API. This is in parallel to our imageio-ffmpeg plugin, because (a) it doesn't have feature parity and (b) I still need to work out how well maintained pyAV is. One of the reasons why we still want to have a pyAV plugin is because pyAV is more performant and - depending on the task - I get a 3x -10x speedup by switching to this plugin.

Other features:
- sophisticated way of checking if ffmpeg can handle the ImageResource that is based on ffmpeg's logic instead of us manually checking extensions
- ImageProperties (shape+dtype) without reading pixel data 🎉 
- supports reading and writing of videos
- adjustable thread count when reading/decoding
- optimised threading model for bulk-reading (`index=None`)
- custom conversion between av.VideoFrame and np.ndarray to support formats that pyAV doesn't support yet (I think I will try to get this merged upstream at some point)
- support for filters and graphs (arbitrary ones when iterating, zero-delay ones writing writing or reading single frames)
- typing support

Planned features for upcoming PRs: 
- support for streams (needs rework of `core.Requests` to accept relevant schemes)
- Fast-path based on file extension. [This will happen in a separate PR because it is quite substantial (more than 150 supported formats) and I want to keep this PR focussed on code].
- support for libavdevice (again to keep this PR on core functionality)


FirefoxMetzger:
@hmaarrfk I saw your name pop up while reading through the pyAV issues. Do you have any interest in a plugin like this?

hmaarrfk:
(between .3x - 10x times faster in my naive timings).

Did you forget a 1 somewhere?

FirefoxMetzger:
> (between .3x - 10x times faster in my naive timings).
> 
> Did you forget a 1 somewhere?

True. it's 1.3x - 10x as fast. 1.3x (30% faster) for bulk reading of an entire video (where decoding dominates runtime), about 5x faster for first frame access (where setup cost dominates), and 10x faster for a random frame (where seeking dominates). Although the seeking depends heavily on the contents of the file and how keyframes are distributed, so that number should be taken with some salt.

hmaarrfk:
For what its worth, I've been trying to get the hardware decoders in reasonable versions of ffmpeg to work.

I think I got the intel ones in conda-forge.

I tried to get nvidia ones in, but I think there are some intriciacies with dependencies and driver versions.

So i am generally interested in speedups :D but not sure how to help at this point. I might be able to give a review if that is what you want?

hmaarrfk:
Is there a way to hide codecov from the PR? it is really annoyign to read it.

FirefoxMetzger:
> but not sure how to help at this point. I might be able to give a review if that is what you want?

Sure that would help :) Though, I wouldn't want to take up too much of your time at this stage. Maybe I can ping you once the PR is RTR (so that, among other things, the codecov messages are gone ;) ).

I primarily pinged you for awareness, because I thought you could be interested in a feature like this. 

I haven't yet figured out a good way to disseminate news around ImageIO nor a good way to estimate interest in new/upcoming features, so this is also partly me figuring out a good way to go about this. Ideas are welcome :D

> I tried to get nvidia ones in, but I think there are some intriciacies with dependencies and driver versions.

Interesting. I've always thought of hardware acceleration (especially via GPU) as something that encoding benefits from, but decoding not so much. Primarily, because sending data back and forth isn't exactly cheap. Do you know of a benchmark for this? (I did some searching yesterday, but didn't find any comparisons beyond people guesstimating)

On that note, you are probably aware of [this note](https://pyav.org/docs/develop/overview/about.html?highlight=hardware#unsupported-features) in pyAV that hardware acceleration won't be supported. Not sure how much we can do from here without adding new features to pyAV (e.g., to expose the required config).

hmaarrfk:
`-hwaccel` is one thing, but there are different encoders that can help speed things up.

Thanks for looking into this!

FirefoxMetzger:
> -hwaccel is one thing, but there are different encoders that can help speed things up.

@hmaarrfk Ah, so you are mainly interested in encoding/writing of video?

hmaarrfk:
Decoding too! It is very useful when dealing with large stacks of images!

FirefoxMetzger:
I am done with the bulk of the work for this. I still need to write some narrative docs and fix pypy (though I think it breaks because of the new astropy and not because of pyav), but the code is there and it is covered by tests.

@hmaarrfk Would you have the time in the coming days to review the code section of this PR (the file pyav.py)?

tests/test_pyav.py
```diff
@@ -0,0 +1,329 @@
...
+        assert props.is_batch is False
+
+        props = plugin.properties(index=None)
+        assert props.shape == (280, 3, 720, 1280)
+        assert props.dtype == np.uint8
+        assert props.is_batch is True
+
+
+def test_video_format_to_dtype():
+    fake_formats = [
```
hmaarrfk, line 103: I'm curious why these are fake formats. There are presumably many fake formats. Why choose these?

tests/test_pyav.py
```diff
@@ -0,0 +1,329 @@
...
+        index=None,
+        plugin="pyav",
+        filter_sequence=[
+            ("drawgrid", "w=iw/3:h=ih/3:t=2:c=white@0.5"),
+            ("scale", {"size": "vga", "flags": "lanczos"}),
+            ("tpad", "start=3"),
+        ],
+    )
+
+    assert frames.shape == (283, 3, 480, 640)
```
hmaarrfk, line 146: Do images really come out as `time, color, height, width`? Typically you would think of them as coming as `time, height, width, color`

tests/test_pyav.py
```diff
@@ -0,0 +1,329 @@
...
+            ("framestep", "5"),
+            ("scale", {"size": "vga", "flags": "lanczos"}),
+        ],
+    )
+    img_bytes = iio.imwrite(
+        "<bytes>",
+        img,
+        format_hint=".mp4",
+        plugin="pyav",
+        in_pixel_format="yuv444p",
```
hmaarrfk, line 176: By default, pyav returns images in `yuv444p`?  Shouldn't you want things as `rgb` by default in a `python` world?

imageio/plugins/pyav.py
```diff
@@ -0,0 +1,839 @@
...
+
+"""
+
+from math import ceil
+from typing import Any, Dict, List, Optional, Tuple, Union
+
+import av
+import av.filter
+import numpy as np
+from numpy.typing import ArrayLike
```
hmaarrfk, line 17: The typing module is really new. Do you need to bump numpy quite high?

hmaarrfk (commented):
Just a few comments for now.

tests/test_pyav.py
FirefoxMetzger: Maybe `fake_formats` is not a good name for them. All of these make sense, and they are typically related to hardware acceleration (which is currently unsupported in pyAV).

The reason I call them fake formats here is because we can instantiate them as a format (`av.VideoFormat(fake_format_name)` succeeds), but these formats have no components, so we can't deduce an appropriate dtype from these so we have to raise instead.

tests/test_pyav.py
FirefoxMetzger: @hmaarrfk Yes, because we don't modify the raw pixel format unless told to do so. 

`cockatoo.mp4` uses `yuv444p`, which is a planar format and it will thus come out as `time, channel, height, width`. Though, we could easily make it something more familiar by passing `format="rgb24"` which would then produce `time, height, width, channel`.

tests/test_pyav.py
FirefoxMetzger: @hmaarrfk By default, it returns the pixels in the same format as they were stored in the file. If we don't know how to return that format (e.g. because it's not trivally stridable), we raise an exception and ask the user to explicitly specify a suitable format via the `format=` kwarg, e.g., `format="rgb24"` to return RGB frames.

imageio/plugins/pyav.py
FirefoxMetzger: Yes, `ArrayLike` was introduced in numpy 1.20 and we depend on that since ImageIO v2.15.

imageio/plugins/pyav.py
```diff
@@ -0,0 +1,839 @@
...
+    -------
+    shape : Tuple[int, ...]
+        A tuple describing the shape of the image data in the frame.
+
+    Notes
+    -----
+
+    This function assumes that channel components will be byte-aligned upstream
+    if necessary. This applies to the formats monow, monob, rgb4, and rgb8.
+    (This will be converted to gray8/rgb24 upstream.)
```
FirefoxMetzger, line 60:
```suggestion
```

This is no longer the case.

tests/test_pyav.py
FirefoxMetzger: @hmaarrfk Yes, because we don't modify the raw pixel format unless told to do so. 

`cockatoo.mp4` uses `yuv444p`, which is a planar format and it will thus come out as `time, channel, height, width`. Though, we could easily make it something more familiar by passing `format="rgb24"` which would then produce `time, height, width, channel`.

hmaarrfk, line 146:
```suggestion
    # cockatoo.mp4 uses yuv444p, which is a planar format 
    # and it will thus come out as time, channel, height, width. 
    assert frames.shape == (283, 3, 480, 640)

    frames = iio.imread(
        test_images / "cockatoo.mp4",
        index=None,
        plugin="pyav",
        # passing format "rgb24" ensures the images come out as 
        # time, height, width, channel
        format="rgb24",
        filter_sequence=[
            ("drawgrid", "w=iw/3:h=ih/3:t=2:c=white@0.5"),
            ("scale", {"size": "vga", "flags": "lanczos"}),
            ("tpad", "start=3"),
        ],
    )

    assert frames.shape == (283, 480, 640, 3)
```

tests/test_pyav.py
hmaarrfk: understood. thank you.

tests/test_pyav.py
hmaarrfk: could you add a comment in the code to that effect
```suggestion
    # All of these make sense, and they are typically related to hardware acceleration,
    # which is  unsupported in pyAV at the time of writing (2022/03/06)

    # We can instantiate them as a format (av.VideoFormat(fake_format_name) succeeds)
    # but these formats have no components. Therefore we can't deduce an appropriate 
    # dtype from these so we have to raise instead.
    fake_formats = [
```

tests/test_pyav.py
almarklein: Regardless, the final image provided to the user would be `height, width, color`, right?

almarklein: Oh, it does not. IMO imageio should make sure that it is.

imageio/plugins/pyav.py
```diff
@@ -0,0 +1,839 @@
...
+
+            # hacky, but beats running our own format selection logic
+            # (since av_guess_format is not exposed)
+            try:
+                setattr(file_handle, "name", filename or "tmp" + extension)
+            except AttributeError:
+                pass  # read-only, nothing we can do
+
+            try:
+                self._container = av.open(file_handle, mode="w", format=container)
```
almarklein, line 157: Can we avoid opening twice?

imageio/plugins/pyav.py
```diff
@@ -0,0 +1,839 @@
...
+            examples.
+        filter_graph : (dict, List)
+            If not None, apply the given graph of FFmpeg filters to each
+            ndimage. The graph is given as a tuple of two dicts. The first dict
+            contains a (named) set of nodes, and the second dict contains a set
+            of edges between nodes of the previous dict.Check the (module-level)
+            plugin docs for details and examples.
+        constant_framerate : bool
+            If True assume the video's framerate is constant. This allows for
+            faster seeking in side the file. If False, the video is reset before
```
almarklein, line 206:
```suggestion
            faster seeking inside the file. If False, the video is reset before
```

almarklein: Since this can also be auto-detected, what is the need for making this settable with an arg?

imageio/plugins/pyav.py
```diff
@@ -0,0 +1,839 @@
...
+
+        Notes
+        -----
+        Accessing random frames repeatedly is costly (O(k), where k is the
+        average distance between two keyframes). You should do so only sparingly
+        if possible. In some cases, it can be faster to bulk-read the video (if
+        it fits into memory) and to then access the returned ndarray randomly.
+
+        The current implementation may cause problems for b-frames, i.e.,
+        bidirectionaly predicted pictures. I don't have any test videos this
```
almarklein, line 236:
```suggestion
        bidirectionaly predicted pictures. I don't have any test videos of this
```

imageio/plugins/pyav.py
```diff
@@ -0,0 +1,839 @@
...
+
+        self._seek(index, constant_framerate=constant_framerate)
+        desired_frame = next(self._container.decode(video=0))
+
+        # useful flags defined on the frame
+        metadata.update(
+            {
+                "key_frame": bool(desired_frame.key_frame),
+                "time": desired_frame.time,
+                "interlaced_frame": bool(desired_frame.interlaced_frame),
```
almarklein, line 793: I always find it helpful if names that represent flags start with `is_`, e.g.  `is_interlaced` and `is_key_frame`. What do you think?

imageio/plugins/pyav.py
```diff
@@ -0,0 +1,839 @@
...
+    if n_planes > 1:
+        shape = [n_planes] + shape
+
+    channels_per_plane = [0] * n_planes
+    for component in frame.format.components:
+        channels_per_plane[component.plane] += 1
+    n_channels = max(channels_per_plane)
+
+    if n_channels > 1:
+        shape = shape + [n_channels]
```
almarklein, line 90: I'm a bit confused. This suggests that its always `height, width, channels`, but one of the tests seems to suggest it does not. I'm probably missing something.

almarklein (commented):
Nice!

tests/test_pyav.py
FirefoxMetzger: > Oh, it does not. IMO imageio should make sure that it is.

Yes, `p` indicates planar (channel-first) formats. Well technically not channel-first, because you have weird formats like `nv24` which are planar and mix channel-first and channel-last, but yes. Usually, it means channel-first :D

Regarding the discussion of making sure that frames will be channel-last, I think we should continue that discussion in one place (here: https://github.com/imageio/imageio/issues/504)

imageio/plugins/pyav.py
FirefoxMetzger: I'm actually now debating if it would make sense to move all the type annotations into a separate repo and to relax the numpy dependency. Not 100% sure if we depend on anything else from `numpy>=1.20`...

imageio/plugins/pyav.py
FirefoxMetzger: Hm, maybe the name `channel` is a bit misleading, but I'm not 100% sure what good taxonomy for this should be. 

FFMpeg stores frames as a sequence of image planes and pixel components (`channels`) can be spread out between these planes. Some formats use one plane and a pixel's components are packed/contigous (aka, the format is channel-last). For these, `n_channels` counts how many color channels the format has. Other formats store one component per plane (aka the format is channel-first) and for those formats `n_planes` counts the number of channels the format has.

Finally, there are esoteric formats. `n12` and consorts store a different number of of components per plane; this is not trivially strideable so we can't easily create a ndarray from this and I'm raising for this format. There are also `bayer_rggb8` and consorts, which are stridable, but pixels aren't packed so the notion of `channels` inside `shape` doesn't make sense anymore. I'm currently raising for this format as well, but I'm very open for suggestions on how to support it (because it is possible, but the question is how to do so cleanly).

imageio/plugins/pyav.py
FirefoxMetzger: Oh? Where is the other open?

imageio/plugins/pyav.py
almarklein: Maybe just put what you just said in a comment :)

imageio/plugins/pyav.py
almarklein: A few lines up where we establish whether we can read the file.

imageio/plugins/pyav.py
FirefoxMetzger: The auto-detection checks for the container's general ability to store variable framerate. This doesn't say anything about the actual timings of frames inside a file so we allow the user to control this if needed (because it is _really_ slow to seek inside of files with non-constant framerate).

imageio/plugins/pyav.py
FirefoxMetzger: Ah. The open here is for writing and the open call above for reading. Do you mean to try and combine the reading and writing path where possible?

imageio/plugins/pyav.py
FirefoxMetzger: Sure. This here is simply 1:1 plumbing. Inside FFmpeg these flags are called `key_frame` and `interlaced_frame`, so I figured it might be intuitive for users coming from that background to have the same names. Should we add a `is_` variant on top of this or as a replacement?

almarklein: If we do it, we should replace it. 

Good point - using the same terminology as ffmpeg is also worth something, so I'm fine if you leave it as-is.

tests/test_pyav.py
FirefoxMetzger: @hmaarrfk Done. After discussion with @almarklein and with local colleagues I changed the default for reading in pyav to `rgb24`. Passing `format=None` will still read the native colorspace, but it will no longer be the default. I've added a test to make sure both scenarios are covered.

imageio/plugins/pyav.py
FirefoxMetzger: I created https://github.com/imageio/imageio/issues/763 to keep track of this. Since we are using `ArrayLike` in other places in the library this is a larger refactor and I will deal with this in a separate PR.

imageio/plugins/pyav.py
FirefoxMetzger: Done.

tests/test_pyav.py
FirefoxMetzger: done.

imageio/plugins/pyav.py
FirefoxMetzger: Then I will keep them as is, because my general understanding is to keep the `metadata` dict keys close to the names in the format/backend where possible.

FirefoxMetzger:
Nice! I just sorted out the pending suggestions from your reviews @almarklein , @hmaarrfk (thanks for taking the time!). I also sorted out the astropy failures on pypy [by no longer installing astropy, because I can't work out where/why it fails and I also can't reproduce the failure locally].

What's left?

FirefoxMetzger:
@daveisfera I can"t think of anything specific from the top of my head. The main thing that is missing is approval from @almarklein and - perhaps - some documentation, which I will look into after I finished https://github.com/imageio/imageio/pull/769).

If you want, you could help to proofread the docs and see if they are intelligible and user friendly, or - depending on your level of expertise, time, and interest - you can also review the PR itself and see if there is anything that you think should be added or changed.

almarklein (approved):

almarklein:
LGTM, but CI does not approve yet?

FirefoxMetzger:
One of the things it will trip over is the changes in pillow 9.1 which break our CI. We've merged the PR that fixes this already, but I still have to rebase this PR on them. Will do soon(TM).

FirefoxMetzger:
@almarklein @daveisfera @hmaarrfk I added some narrative documentation for the pyAV plugin. If one of you has the time, could you have a look at it and check for legibility and grammar? Once that is done, we can merge the PR :)

imageio/plugins/pyav.py
```diff
@@ -0,0 +1,977 @@
+"""Read/Write Videos (and images) using PyAV.
+
+Backend Library: `PyAV <https://pyav.org/docs/stable/>`_
+
+This plugin that wraps pyAV, a set of pythonic bindings for the FFMPEG library.
```
almarklein, line 5:
```suggestion
This plugin wraps pyAV, a pythonic binding for the FFMPEG library.
```

imageio/plugins/pyav.py
```diff
@@ -0,0 +1,977 @@
...
+It is similar to our famous FFMPEG plugin, but offers a more performant and
+robut interface and aims to superseed the FFMPEG plugin in the future.
```
almarklein, line 7:
```suggestion
It is similar to our FFMPEG plugin, but offers a more performant and
robust interface, and aims to supersede the FFMPEG plugin in the future.
```

imageio/plugins/pyav.py
```diff
@@ -0,0 +1,977 @@
...
+    single filter, you can do this by specifying a filter sequence with a single
+    entry.
+
+A ``filter_sequence`` is a list of filters, each defined through a 2-element
+tuple of the form ``(filter_name, filter_parameters)``. The first element of the
+tuple is the name of the filter. The second element are the filter parameters,
+which can be given either as a string or a dict. The string matches the same
+format that you would use when specifying the filter using the ffmpeg
+command-line tool and the dict has entries of the form ``parameter:value``. For
+example::
```
almarklein, line 74: This is cool, but also leaning towards video processing, which is not our core business. We should be careful not to "pollute" the docs too much. I guess it's ok here though, since it's the last section in theses docs.

imageio/plugins/pyav.py
```diff
@@ -0,0 +1,977 @@
...
+        self,
+        ndimage: Union[ArrayLike, List[ArrayLike]],
+        *,
+        codec: str,
+        is_batch: bool = True,
+        fps: int = 24,
+        in_pixel_format: str = "rgb24",
+        out_pixel_format: str = None,
+        filter_sequence: List[Tuple[str, Union[str, dict]]] = None,
+        filter_graph: Tuple[dict, List] = None,
```
almarklein, line 536: Maybe we should go easy on the type hints (e.g. using a string instead?), because this is not helping 😮  :

image

almarklein (approved):

imageio/plugins/pyav.py
almarklein: Are there docs in pyav that we can delegate some of this to? Perhaps some of this would fit as an actual example, instead of a code sample in the docs.

FirefoxMetzger: Unfortunately, no. Here are the docs on filters in pyav: https://pyav.org/docs/stable/api/filter.html . I worked out how to use them in pyav by digging through GH issues and source code. An item for the example page could, indeed, be nice, and I was already thinking of turning our GIF writing unit test into one. 

The downside of doing just an example is that the examples page is supposed to give various (general) usage examples of ImageIO, whereas the page here is supposed to discuss the features of the pyav plugin. The problem with that is that discussing/describing filter graphs is cumbersome, and a single example does more explaining than me writing a wall of text. That's essentially why I choose to put an example here. We could move them, but then we would need a decent (and clear) replacement for them here, because in the end we have to document available kwargs.

daveisfera:
> @almarklein @daveisfera @hmaarrfk I added some narrative documentation for the pyAV plugin. If one of you has the time, could you have a look at it and check for legibility and grammar? Once that is done, we can merge the PR :)

Looked good to me. I'm not totally sure how I'd use this when [iterating over a movie](https://imageio--745.org.readthedocs.build/en/745/examples.html?highlight=iter#iterate-over-frames-in-a-movie) (is it now the default or do I need to do something to use this new code?), but that's a detail and I don't have any concerns with these changes.

Looking forward to being able to use this!

FirefoxMetzger:
> I'm not totally sure how I'd use this when [iterating over a movie](https://imageio--745.org.readthedocs.build/en/745/examples.html?highlight=iter#iterate-over-frames-in-a-movie) (is it now the default or do I need to do something to use this new code?)

If I understand the question correctly then the answer is that, for now, you will have to explicitly select this plugin using the `plugin=` kwarg, e.g. via

```python
import imageio.v3 as iio

for frame in iio.imiter("imageio:cockatoo.mp4", plugin="pyav"):
    pass  # do stuff
```

because `ìmageio-ffmpeg` will remain the default for MP4. That is until this plugin had some time to prove itself and we have verified that it is indeed stable, feature-complete, and efficient enough to warrant switching the default.

imageio/plugins/pyav.py
almarklein: Mmm, I guess we can then best leave it like this for now.

FirefoxMetzger:
Alright, looks like this PR is ready to go. I will wait with merging it until #779 is ready and then I will merge both in the same week.

This way, they will both be released the same week and we can advertise this plugin with the plugin-specific default for `index` (which is to read all frames/images by default instead of just the first). Should be less confusing to users than to experience this plugin switching behaviors.

FirefoxMetzger:
The time has finally come. Once CI is happy, I will merge the pyav plugin, Thanks @almarklein @hmaarrfk @daveisfera for your feedback and discussion.

PR #747, opened 2022-02-01
DEV: make RTD fail on warnings during CI

FirefoxMetzger:
I noticed our CD failing while building the docs (it is set up to fail on warning to avoid broken links in the stable docs). This PR maches that behavior in our CI pipeline and also fixes the broken link (once I've tracked it down).

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -59,6 +59,7 @@ jobs:
   docs:
     name: Build Docs
     needs: initial_check
+    if: needs.initial_check.outputs.should_release == 'true'
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -59,13 +59,12 @@ jobs:
   docs:
     name: Build Docs
     needs: initial_check
-    if: needs.initial_check.outputs.should_release == 'true'
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
     - uses: actions/setup-python@v2
       with:
-        python-version: '3.8'
+        python-version: '3.9'
     - name: Install Dependencies
       run: |
         pip install -e .[docs]
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -65,7 +65,7 @@ jobs:
     - uses: actions/checkout@v2
     - uses: actions/setup-python@v2
       with:
-        python-version: '3.9'
+        python-version: '3.8'
     - name: Install Dependencies
       run: |
         pip install -e .[docs]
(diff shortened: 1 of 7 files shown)


FirefoxMetzger:
Hm. Either it got fixed by me changing the CD pipeline to use python 3.8 instead of 3.9 or I can't reproduce this. Either way, adding a CI job that mirrors the test that is done during CD is probably worthwhile and so is failing the docs on a warning.

@almarklein I'm happy with this one. Feel free to merge if you agree.

.github/workflows/ci.yml
```diff
@@ -17,6 +17,22 @@ on:
 
 
 jobs:
+  docs-GH:
+    name: Build Docs
+    needs: initial_check
+    if: needs.initial_check.outputs.should_release == 'true'
```
almarklein, line 23: Why not always run this doc test?

.github/workflows/cd.yml
```diff
@@ -65,7 +65,7 @@ jobs:
     - uses: actions/checkout@v2
     - uses: actions/setup-python@v2
       with:
-        python-version: '3.9'
+        python-version: '3.8'
```
almarklein, line 68: Why *reduce* the Python version?

almarklein (approved):
Looks good. Just two minor questions.

.github/workflows/ci.yml
FirefoxMetzger: Good point. It's a relic from copying the job from CI. I will remove this.

.github/workflows/cd.yml
FirefoxMetzger: Because it matches my local python version and I thought it might be related (no issues building locally for me). You did make a good point with the conditional above though. I will investigate this and see if we can't bump back up to python 3.9

almarklein: or 3.10 ;)

FirefoxMetzger:
@almarklein Okay, I can confirm that this is indeed a python 3.9 vs 3.8 issue (lol). I will need to make time to write a MWE and submit a bug report to sphinx.

FirefoxMetzger:
I created a sample repo to [reproduce the issue](https://github.com/FirefoxMetzger/sphinx-39-bug); in particular, it has a PR that shows things working in python <=3.8 and breaking in >3.8. It seems to be related to us dynamically importing plugins (but that shouldn't matter). I may have to poke the sphinx devs on this.

In the meantime, I vote for keeping the docs built on 3.8. Once we know what the cause is, we can fix it and bump up to 3.10.

@almarklein What are your thoughts on the timeline for this PR? Or CD will remain broken until this one gets merged, so I am motivated to have this resolved :D

almarklein:
Interesting Sphynx issue :)  - yeah, let's just use py38 then, no big deal!

PR #750, opened 2022-02-10
BUG: read metadata of current frame when iterating + APNG fast path

FirefoxMetzger:
This PR adds APNG to our list of extensions and enables a fast-path for it.

It also fixes a bug: When iterating over images in pillow (imiter), we need the frames metadata (to apply gamma or orientation), but we didn't extract frame-specific metadata before and instead extracted the metadata from the first frame each time. This PR fixes this behavior :)

### DIFF
--- a/imageio/config/extensions.py
+++ b/imageio/config/extensions.py
@@ -81,6 +81,12 @@ extension_list = [
         priority=["RAW-FI"],
     ),
     FileExtension(
+        name="Animated Portable Network Graphics",
+        external_link="https://en.wikipedia.org/wiki/APNG",
+        extension=".apng",
+        priority=["pillow"],
+    ),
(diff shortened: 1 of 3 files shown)


imageio/config/extensions.py
```diff
@@ -80,6 +80,12 @@ def reset(self) -> None:
         extension=".arw",
         priority=["RAW-FI"],
     ),
+    FileExtension(
+        name="Animated Portable Network Graphics",
+        external_link="https://en.wikipedia.org/wiki/APNG",
+        extension=".apng",
+        priority=["pillow"],
+    ),
```
almarklein, line 88: Nice how you can add formats like this now :)

almarklein (approved):

Issue #498, opened 2020-01-05
webp writer writes 0 bytes to BytesIO 


leoshine:
Toy code to reproduce the issue (tested on both linux and osx): 

```python
# webp writer writes 0 bytes to BytesIO 
with io.BytesIO() as bio:
    im_rgb = imageio.imread('src_input.jpg')  
    with imageio.get_writer(bio, format='webp', mode='i') as writer: # doesn't work for webp.
        writer.append_data(im_rgb)
    data = bio.getvalue()
    print(len(data))  # always return 0 here for 'webp' (but it is fine for jpg, png, ect)

```

almarklein:
Could you try `format='targa-FI'`? That would also use the default FreeImage plugin. Maybe it helps to determine where it goes wrong. I don't seem to be able to write `webp` on Windows.

Issue #604, opened 2021-02-27
Support WebP Animations


Skylion007:
WebP can support animations too (ie multiple frames). FFMPEG can generate these, but I am unsure how to add this PR since WebP is already supported for single images. What would be the best way to add this functionality? Would just adding webp to the support ffmpeg extensions solve the issue?

FirefoxMetzger:
It's actually already possible to write `webp` in `imageio` if you manually select the plugin. Here is my tinker in the new API (#574 ):

```python
import imageio as iio
im = iio.new_api.imread("random.gif", legacy_plugin=False, mode="RGB")

# FFMPEG isn't ported yet
with iio.imopen("test.webp", legacy_plugin=True, plugin="FFMPEG") as file:
    writer = file.legacy_get_writer(iio_mode="I", codec="libwebp", fps=1)
    for frame in im:
        writer.append_data(frame)
    writer.close()
```

which translates to the current (v2.9) API as something like

```
import imageio as iio
frames = iio.mimread("random.gif", pilmode="RGB")  # Note: frames is a list not an array
writer = iio.get_writer("test.webp", format="FFMPEG", mode="I", codec="libwebp", fps=25)
for frame in im:
    writer.append_data(frame)
writer.close()
```

This works, because specifying a `format=` skips the search for a compatible plugin and instead assumes that the user-specified plugin can handle the format. If you want to add "native" support for `webp` ala `iio.get_writer("test.webp")`, you will have to do two things:

1. Add `webm` to the list of supported formats (as you suggested), so that ffmpeg will be seen as a compatible plugin during the search.
2. Add logic to `FFmpegFormat` to switch to the `libwebp` codec when required. For writing, checking the file extension will be sufficient. 

I didn't investigate reading. @almarklein probably knows the correct options here, and can also tell you about the procedure for implementing `mimread`.

almarklein:
Shouldn't this work?
```py
iio.mimwrite("test.webp", format="FFMPEG", codec="libwebp", fps=25)
```

FirefoxMetzger:
Good catch. You can definitely use `mimread` as a one-liner for writing. I guess it should then have the image as second argument, i.e.,

```python
iio.mimwrite("test.webp", frames, format="FFMPEG", codec="libwebp", fps=25)
```

My question was about reading, because `codec` is currently not a supported argument for `mimread` with `imageio-ffmpeg`.

FirefoxMetzger:
[...]

Issue #324, opened 2018-03-23
Support for PDFs?


1313e:
I am actually very surprised that this question exists, given the number of image formats that are supported, but is there any reason why PDFs are not supported as an image format or are there any plans to include this in the future?

almarklein:
I think that this would be very interesting indeed. Do you know what options currently exist to do this from Python?

1313e:
Hmmm, I know that matplotlib fully supports PDFs, in that it can save anything in it. Since I always use imageio for working with images, I do not really know of any other packages that can maybe do this. I actually fully expected it to be capable of working with PDFs, given how many formats are supported and how common PDF is.

I can go and check if there are some options of loading PDFs into Python, if that is what you were asking? It would make my life a lot easier, since I prefer saving my figures in PDF over PNG, since PDFs have lower file sizes and save more details about the picture (and are easier to export).

almarklein:
It would be good to distinguish between vector graphics and raster graphics in this discussion. Matplotlib has an internal description of lines, text and other shapes that it can export to formats like eps and pdf, giving small file sizes and crisp images even when zoomed in. It can also export to e.g. PNG, in which case a "snapshot" is made at a certain resolution and dpi. (This generally results in a larger file size, unless your plot contains *a lot* of points.)

Imageio is targeted entirely on raster images. You read an image, and get a numpy array that contains the pixels (e.g. NxMx3 for RGB images). Similarly, you can use imageio to save such numpy arrays to a variety of formats.

So when you store an image with imageio to both png and pdf (when we have a plugin for it), the pdf will likely be *larger* than the png, because the PDF will simply embed the raster image bitmap.

The use-case for pfd support in imageio would be to make it easy to e.g. add images to a pdf presentation. And to e.g. read all images contained in a PDF document. In the latter case, its probably only feasable to read the images that are already in raster format, i.e. reading a figure exported by matplotlib will likely not work --- unless we make use of some sort of external program that can "render" the image.

1313e:
Hmm, okay, I understand.
I didn't know about the differences in types of graphics.

I am using imageio mostly for making movies out of large series of plots that I have created with matplotlib. I prefer to save these plots in PDF, because of the reasons you mentioned (also saving the lines and text, such that you have higher quality images when zoomed in).
Using that for a movie would be very useful, but I can see how that would become really hard to do.

[...]

PR #752, opened 2022-02-11
MAINT: Cleanly determine write format in pillow

FirefoxMetzger:
Closes: https://github.com/imageio/imageio/issues/498, https://github.com/imageio/imageio/issues/604
Solves 50% of: https://github.com/imageio/imageio/issues/324

This PR cleans up the way pillow determines if it can write a format or not and makes use of `format_hint` to choose the writing format when writing to file/likes.

It also adds a few extensions that pillow supports but that we didn't track explicitly yet. Most notably: PDF (writing only) and WEBP (reading and writing). I"ll add a snippet to both issues.

### DIFF
--- a/imageio/config/extensions.py
+++ b/imageio/config/extensions.py
@@ -97,6 +97,10 @@ extension_list = [
         priority=["RAW-FI"],
     ),
     FileExtension(
+        extension=".blp",
+        priority=["pillow"],
+    ),
+    FileExtension(
         name="Bitmap",
         extension=".bmp",
         priority=["pillow", "BMP-PIL", "BMP-FI", "ITK"],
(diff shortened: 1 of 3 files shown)


FirefoxMetzger:
@almarklein RTR :)

I also removed a helper function that we no longer need. We had it as a wrapper around a pillow function, because pillow was a bit inconsistent (the function didn't support all modes pillow officially mentioned). I added a PR upstream to fix that and now we can remove the helper here and simply match pillow's behavior.

FirefoxMetzger:
bumping this since it hasn't received any activity for a while. Are you able to review @almarklein? If not, I can find a different solution, just let me know :)

imageio/config/extensions.py
```diff
@@ -551,12 +559,12 @@ def reset(self) -> None:
     FileExtension(
         name="Moving Picture Experts Group",
         extension=".mpeg",
-        priority=["FFMPEG"],
+        priority=["pillow", "FFMPEG"],
```
almarklein, line 562: Wait, since when can pillow handle mpeg?

almarklein (approved):
LGTM. Just one question.

PR #757, opened 2022-02-23
Update "python_requires" to 3.7

### LINKED ISSUE #756: setup.py "python_requires" is still set to Python 3.5
https://github.com/imageio/imageio/blob/master/setup.py#L221

I'm using Python 3.6 and now with version 2.16.0 of `imageio` forcing the `numpy` requirement >= 1.20.0, pip is having a hard time resolving dependencies (only on Windows for some reason). pip thinks it can install `imageio` 2.16.0 because it says it's compatible with python 3.5 but `numpy` 1.20.0 requires Python 3.7+.

I'm using the latest version of pip (21.3.1) that is supported on Python 3.6.
For some reason, I don't have any issue resolving dependencies on Linux.

If the "python_requires" had been updated to 3.7+ for the 2.16.0 release of `imageio`, then pip would stick with using version 2.15.0 and there would be no problem.

Can you publish a hotfix to pypi for version 2.16.0 with the "python_requires" set to Python 3.7?

BTOdell:
Closes #756 

Please release v2.16.1 asap and yank v2.16.0 on pypi. See above issue for reasoning.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -218,7 +218,7 @@ setup(
     long_description=long_description.replace("__doc__", __doc__),
     platforms="any",
     provides=["imageio"],
-    python_requires=">=3.5",
+    python_requires=">=3.7",
     install_requires=install_requires,
     extras_require=extras_require,
     packages=["imageio", "imageio.core", "imageio.plugins", "imageio.config"],


FirefoxMetzger:
Thanks for the PR! As I mentioned in the issue, its good that you caught this and let us know about it.

Issue #766, opened 2022-03-14
Problem importing imageio_ffmpeg in frozen applications


paulmueller:
I see this issue with imageio 2.16.1
I don't see this issue with imageio 2.10.3

This might have been caused by https://github.com/imageio/imageio/commit/bdbe699bbcda4223b0b6bd4d7474f84bbe34af09 (which appeared first in imageio 2.11).

For some reason imageio cannot import the imageio_ffmpeg module in frozen applications anymore.

This is the imageio-related traceback that I get:

```
  File "imageio\core\functions.py", line 98, in get_reader
  File "imageio\core\imopen.py", line 332, in imopen
ValueError: Could not find a backend to open `C:\Users\Paul\Downloads\fmt-tdms_fl-image-bright_2017\M4_0.040000ul_s_imaq.avi`` with iomode `r?`.
Based on the extension, the following plugins might add capable backends:
  FFMPEG:  pip install imageio[ffmpeg]
```

I checked an imageio_ffmpeg got bundled into the application.

FirefoxMetzger:
Interesting. We recently added [a section on freezing](https://imageio.readthedocs.io/en/stable/user_guide/freezing.html) to the docs (thanks again @richardsheridan ). Did you account for what is written there? I.e., did you make sure that pyinstaller ends up including the ffmpeg plugin file? (Should be in imageio/plugins/ffmpeg.py)

paulmueller:
No, I didn't. Why don't you add those plugins to the imageio pyinstaller hook? From the docs I understand they are light-weight, so I guess this would be the best choice.

FirefoxMetzger:
> Why don't you add those plugins to the imageio pyinstaller hook?

Lack of knowledge from my side. If it's possible to add them automatically, I'm happy to have that added; however, I never used pyinstaller myself so I don't know what would be needed to accomplish that. I could - of course - spend the time and learn, but that time currently goes to other issues with higher priority.

That said if you would like to submit a PR that adds this, I'm happy to review.

paulmueller:
OK. I can do that. The PR would be with Pyinstaller. I will let you know here. Please keep this open in the meantime.

paulmueller:
[...]

PR #768, opened 2022-03-21
DOC: note on preferring venv

FirefoxMetzger:
Closes: https://github.com/imageio/imageio/issues/766

In https://github.com/pyinstaller/pyinstaller-hooks-contrib/pull/396 @paulmueller added dynamically imported plugins to the list of modules that are automagically collected when freezing imageio. A result of the discussions in that PR was that it is recommended to package apps inside a virtual environment and the PR here adds a note about this to our docs.


almarklein (approved):

almarklein:
I agree :)

PR #769, opened 2022-03-25
MAINT: Deprecation warnings to transition to v3

FirefoxMetzger:
This PR adds deprecation warnings to inform users about API changes that will happen once take the step to v3 and how to deal with them.

It also reformats our testsuite to deal with the warnings by

- switching from `import imageio` to `import imageio.v2 as iio` where applicable
- adding a `deprecated_test` fixture that suppresses warnings for tests that will be removed when deprecated functions disappear (mainly related to the old format manager)
- consistently using `iio` throughout the test-suite

### DIFF
--- a/imageio/plugins/pillow.py
+++ b/imageio/plugins/pillow.py
@@ -163,8 +163,6 @@ class PillowPlugin(PluginV3):
         else:
             iterator = self.iter(mode=mode, rotate=rotate, apply_gamma=apply_gamma)
             image = np.stack([im for im in iterator], axis=0)
-            if image.shape[0] == 1:
-                image = np.squeeze(image, axis=0)
             return image
 
     def iter(self, *, mode=None, rotate=False, apply_gamma=False) -> np.ndarray:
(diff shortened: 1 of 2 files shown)


imageio/plugins/pillow_legacy.py
```diff
@@ -213,6 +214,12 @@
         mode 'F'.
 """
 
+warnings.warn(
+    "The `PillowFormat` plugin is deprecated and will be removed in ImageIO v3."
+    " Use the new `PillowPlugin` instead.",
+    DeprecationWarning,
+)
```
almarklein, line 221: When does a user get this warning? Most users won't know how to select the `PillowPlugin` over the `PillowFormat`, so I'm not sure how useful this warning is ...

imageio/plugins/tifffile.py
```diff
@@ -175,6 +175,12 @@
 try:
     import tifffile as _tifffile
 except ImportError:
+    warnings.warn(
+        "ImageIO's vendored tifffile backend is deprecated and will be"
+        " removed in ImageIO v3. Install the tifffile directly:"
+        " `pip install imageio[tifffile]`", 
+        DeprecationWarning
+    )
```
almarklein, line 183: Dito, under what circumstances does this warning occur?

imageio/plugins/pillow_legacy.py
FirefoxMetzger: For users that sidestep our normal plugin selection mechanism and that access the format directly, e.g., via `iio.formats["PNG-PIL"]`. I expect this to be niche, but it is better to warn than to just drop it :)

At the moment this will also warn for some of our codepaths, but this is a thing that I have to change (and one of the reasons this is still a draft).

imageio/plugins/tifffile.py
FirefoxMetzger: This will occur if a user loads a TIFF file and doesn't have `tifffile` installed. 

I'm not 100% sure if we are making `tifffile` a primary dependency once we remove this. You will probably say we should, and the bioinformatics community will use it a lot, but for other scenarios, it is rather optional. In most cases it won't hurt to have it anyway but considering that there are use-cases with space concerns (e.g. the recently discussed freezing of apps) it might be better to keep things that can be optional optional.

FirefoxMetzger:
@almarklein Do you remember what we decided for the plugin resolution logic of v2 plugins? I remember that we wanted to serve v3 plugins through the v2 API, but I don't remember if we want to keep v2-unique resolution logic. 

Should we deprecate it and resolve plugins using the same rules as v3? (Sounds more maintainable to me)

almarklein:
> @almarklein Do you remember what we decided for the plugin resolution logic of v2 plugins? I remember that we wanted to serve v3 plugins through the v2 API, but I don't remember if we want to keep v2-unique resolution logic.

I *think* we wanted to keep the selection logic for backwards compatibility, right? There should not be much maintenance, since the `v2` module is on a deprecation path anyway.

FirefoxMetzger:
@almarklein This one is also RTR. 

I'm trying to find the parts we are missing to get the patch coverage to 100%, but since it touches files that aren't fully covered this is proving more laborsome than expected. I will update the ones I find, but am happy otherwise since overall coverage improves.

PR #770, opened 2022-03-28
DEV: build docs using python 3.10

FirefoxMetzger:
A while ago we had trouble building our docs in python 3.9+. This traced back to a surprising interaction between our dynamic plugin loader and sphinx that resulted in autosummary not building the plugin classes anymore.

This PR fixes that behavior using the solution from https://github.com/sphinx-doc/sphinx/issues/10182.


FirefoxMetzger:
@almarklein This PR should be ready to go. I will update further if any tests fail, otherwise feel free to review.

PR #779, opened 2022-04-05
FEAT: add plugin default sentinel

FirefoxMetzger:
Not sure if we should track this as a feature or a bugfix, because it is both an addition of new functionality as well as a fix of a previous regression (that has so far only been noticed by a colleague of mine).

This PR adds a custom sentinel (called `PLUGIN_DEFAULT`) and makes it the default value of the `index` kwargs in the v3 API. This delegates the default value of `index` to the plugin, instead of having a global default overwrite plugin decisions. This solves a few problems:

1. It solves a regression that I didn't notice. In early iterations of the API the default was `None`, which means that videos, GIF, and APNG would by default load all images in the file (the sensible thing to do). I changed the global default to `0` later on to closer match the behavior of v2 and didn't consider how that affects those formats. With this PR, we can bring the `None` behavior back for the aforementioned formats while still behaving similarly to v2 in other cases. (which I think is an overall improvement.)
2. pyAV (and other video plugins) would benefit from a `None` default, because a user would expect to get more then one frame when reading a video.
3. It allows more flexibility to write new plugins, because we can now have more fine-grained control over what constitutes a sensible default for `index`.
4. It allows plugins that support multiple formats (e.g. pillow) to specify format-specific defaults while still allowing for explicit `index=0` and `index=None` overwrite where desired.
5. We can have different behavior for the v2 and v3 API, which brings us one step closer to using v3 plugins with the v2 API. `v2.imread` can use `index=0` and make v3 plugins produce the expected result (read the first frame).

So yeah, overall I think a net improvement. It does come with the not-so-nice part of using a custom sentinel instead of the usual `None`; however, `None` is already taken and means "read all images in the file". If there is a better solution let me know and I'll update :D


almarklein:
Do you think we can change this so that `None` is the default, and we add a sentinel for `ALL`? Would that break backwards compat somewhere, and can we accept that (in the new v3 module)?

If we go this route now, we're stuck forever with a pretty odd and unpythonic way to set a default.

FirefoxMetzger:
Hm, I was thinking the other way around. If we use `None` as our internal sentinel then we force the user to import our `ALL` sentinel if they want to set it. On the other hand, a user will never have to explicitly set `PLUGIN_DEFAULT`, because that is already the default. 

Instead of introducing a dedicated sentinel object we could also use a string instead. I.e., we can allow `index: Union[int,str]` and ask users to use `index="all"` to read everything. 

Hypothetically, we could then also change `index=None` to `index="plugin_default"` which might be more meaningful/legible?

FirefoxMetzger:
> What about ellipsis: index=... means all?

Nice, I like that idea. I will think on it for a while to see if I can come up with any downsides, but if not then I'd say we go with that :)

FirefoxMetzger:
I changed it around so that `index=...` means "read all" `index=int` means read that index, and `index=None` means "let the plugin decide".

It seems to work; however, it is a bit weird for metadata, because now `index=int` means metadata for that index (so far so good), but `index=...` means global metadata. What are your thoughts on this @almarklein?

almarklein:
Maybe the plugin default for reading metadata is to read the global data, so that `index=...` == `index=None`?

FirefoxMetzger:
> Maybe the plugin default for reading metadata is to read the global data, so that index=... == index=None?

I think the plugin should decide if the default should be to read local or global metadata, just like it should decide if it should read one or all images by default. That way it can match the default behavior and `iio.imread` followed by `iio.immeta` will do the expected thing and serve matching data.

I was moreso wondering if `iio.immeta(index=...)` seems odd and if we are okay with it looking like that.

almarklein:
> I think the plugin should decide if the default should be to read local or global metadata

Good point, it might depend on the format what makes more sense.

> I was moreso wondering if iio.immeta(index=...) seems odd and if we are okay with it looking like that.

IMO the `...` looks ok when it means *all*. If it means *global* like for meta data, it's a bit odd, but at least its symmetric with `imread()` ...  I suspect the use-case is relatively rare, so I'm willing to accept that it looks a bit off.

FirefoxMetzger:
@almarklein Okay, in that case, I am happy with the PR, too. Any other changes/suggestions before we merge?

almarklein:
Nope, LGTM!

FirefoxMetzger:
awesome. I just synced with master and will merge once the tests are complete. Then i will go and do the same for the pyav plugin.

Issue #763, opened 2022-03-11
Move type hints into separate module


FirefoxMetzger:
I'm wondering if it is worthwhile to move all of our type hints into a separate module. The reason for this is that we depend on `numpy>=1.20` because we use `numpy.typing.ArrayLike`. This doesn't affect core functionality and it may, hence, be useful to relax this constraint.

almarklein:
Having a dependency on a specific minimum version of numpy just for the type hint is indeed not great.

What would it look like to have the typing in a separate module?

FirefoxMetzger:
I'm not 100% sure actually, so I will have to look into that. I know that it is possible to have annotations separate (mainly intended for packages where maintainers don't add them directly, but you still want them), but I never dug into this. I will have a look once I have dealt with more urgent work :)

almarklein:
As someone who is not a particular fan of type hints, them not "polluting" the code, while still having their benefits sounds appealing :D

Issue #652, opened 2021-06-24
Feature request: Support mypy type checking?


rggjan:
Would be great if imageio could be statically type checked with:
https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-type-hints-for-third-party-library

FirefoxMetzger:
Sounds very closely related to #651 . I absolutely agree that type annotations would be a great addition. Being able to statically check them would be even better. 

The main reason this hasn't happened yet is lack of manpower. If you would like to contribute this @rggjan , I'm happy to help you navigate the library and to review the PR :)

rggjan:
Thanks! Yes, I didn't find the other PR, they would probably nicely complement each other...

Issue #651, opened 2021-06-22
Consistent type annotations across the library


FirefoxMetzger:
Some of the functions in our library have the correct annotation already, whereas some (more legacy) functions don't feature annotations yet. At the same time, there is a (pretty useful) trend to outsource annotations into a separate [`.pyi` file](https://www.python.org/dev/peps/pep-0484/#id44) (think C++ headers, but actually just annotations not declarations). It makes sense to follow this trend here because it keeps the code clean while giving us nice typing support.

What is needed here is the following:

- [ ] introduce a .pyi file for each file in the library that contains code (we can ignore the `_plugin.py` files, as they are subject to depreciation/outsourcing in the future)
- [ ] introduce a new `ImageResource` type containing all valid inputs for a Request
- [ ] migrate the existing annotations from our functions into the pyi files
- [ ] add annotations for each function that currently lacks annotations

This is an easy addition and a great way to get an overview of the code-base. If there is a section of the library that you use particularly often, and you would like it to support annotations, partial contributions are also welcome.  If you do, please do leave a comment below, so we can coordinate the work and avoid people racing each other on the same section of the library.

almarklein:
How would the information in the .pyi file be used? Would you include new tools to validate stuff in CI for instance? Or can people using imageio benefit from them?

FirefoxMetzger:
It is primarily for consumers of ImageIO. With proper type annotations, you get a lot of help from a modern IDE; for example, it can more easily deduce the type of a variable and - for example in the case of classes - hint available methods. For example, I can do `instance: PillowPlugin = imopen(..., plugin="pillow")` and then my IDE will provide auto-completion for all the methods that PillowPlugin offers (+ the docstrings of said functions if desired) whenever I type `instance.re...`. Similar to what you are used to from IntelliSense in C++ or Java. Very handy.

The other option is that it allows you to run static/"compile-time" code analysis to verify that data types are handled consistently. This can warn you C-style if your variables change type through a function's execution, or if you pass unexpected types into a function, or your function doesn't return what it claims in some cases. Sometimes this is intentional, but most of the time it is a bug and you probably wish to fix it. For that to work, however, the module needs some indication of what the expected/allowed type is. Hence, type hinting.

PR #784, opened 2022-04-11
MAINT: Better Type Annotations + mypy support

FirefoxMetzger:
Closes: https://github.com/imageio/imageio/issues/763, https://github.com/imageio/imageio/issues/652, https://github.com/imageio/imageio/issues/651

This PR moves type annotations into dedicated `pyi` files and introduces a `ArrayLike` and a `ImageResource` type. 

The `ArrayLike` type is equal to `numpy.typing.ArrayLike` if numpy>=1.20 is installed, otherwise it falls back to `np.ndarray`. I think this should remove our dependency on numpy>1.20.

The `ImageResource` type is a convenience to indicate that the `uri` argument (which we should rename) can accept other types than strings.


almarklein:
Wow. Out of curiosity, did you use a tool to generate these pyi files, or is that all manual work? 😲

FirefoxMetzger:
So far I've done all of this manually. I move out the current type annotations into a pyi file and then I let mypy run over the library to statically validate the codebase and fix any problems that it finds.

I would like to be able to have all the `.pyi` files live in a separate folder/module, similar to header files in a C project. Unfortunately, I don't know if that is currently possible...

FirefoxMetzger:
This PR is likely not perfect, but it accomplishes the main goals:

- remove the numpy>1.20 dependency
- enable mypy validation of files that use imageio
- move type annotations out of the main code and into pyi files for core files (I kept type annotations inside of plugins intact for now until we decide if they should be pulled out as well)

almarklein (approved):
I'm not going to check this line-by-line, but from a high level POV this looks good!

PR #787, opened 2022-04-14
BUG: ensure pts increases even if pyav doesn't flush the frame immediately

FirefoxMetzger:
When writing video with pyav frames may be buffered inside the codec instead of being written immediately. If this happens we currently lose track of how many frames have been written (as we check the stream and see how many are in there). This PR updates our frame counting to use our own counter which is independent of any buffering that pyav or FFmpeg may do internally.


FirefoxMetzger:
@almarklein RTR :)

almarklein (approved):
Nice catch!

Issue #761, opened 2022-03-03
Feature Request: FFMPEG Writer: Provide users with the ability to specify the pixel format


hmaarrfk:
it would be nice to provide users with the ability to specify their own input pixel format as the input.

Users may choose to represent their images in exotic format.

Wrapping the following `self._pix_fmt` assignment in an `if self._pix_fmt is not None:` might be enough if we provide the users with an input parameter too.

https://github.com/imageio/imageio/blob/6c259ba3d21e6f69f46f7ae8ecbe0e1a2e6ec242/imageio/plugins/ffmpeg.py#L588

Let me know if this is something you would want a PR for.

FirefoxMetzger:
> it would be nice to provide users with the ability to specify their own input pixel format as the input. [...] Users may choose to represent their images in exotic format.

Yes, that would be nice. Do you have a specific subset in mind, i.e., how esoteric were you thinking of?

> Wrapping the following self._pix_fmt assignment in an if self._pix_fmt is not None: might be enough if we provide the users with an input parameter too.

My knowledge of the ffmpeg plugin is not as broad as the other plugins, but based on that (and the check I just did) it looks like this is indeed all that would be needed. and I'd also be happy to review a PR for this.

---

At the same time, the pyAV plugin [will be able to do that](https://github.com/FirefoxMetzger/imageio/blob/6f0b8c911cee7fc14c6da1009e0579b11df19830/imageio/plugins/pyav.py#L470-L475). I can't say how long it will take for that to get merged though (it's missing some coverage and documentation at this point, but I have been quite busy lately). Bottom line is: If you are okay with waiting one or two months for this, then you should get this via pyav. If you need it earlier then I will happily review a PR.

hmaarrfk:
The nice thing about the ffmpeg plugin is that it uses the ffmpeg EXE which is standalone compared to a linked application.

Honestly, I've included:
```
writer._pix_fmt = pix_fmt
writer._size = image_shape
writer._depth = 1
writer._initialize()
```

to my code. I know I shouldn't, but for now it works.

Some might choose to have their images in YUV or RAW ;)

FirefoxMetzger:
> The nice thing about the ffmpeg plugin is that it uses the ffmpeg EXE which is standalone compared to a linked application.

Yes calling an executable is nice, because you can easily switch between API-compatible versions. You do pay for that flexibility with performance though (multiprocessing + stdin/stout pipe), but if you don't mind that it is a workable solution indeed.

[...]

PR #788, opened 2022-04-14
FEAT: allow writing bayer images

FirefoxMetzger:
Closes: https://github.com/imageio/imageio/issues/761

This PR adds the ability to provide ndimages formatted as bayer/raw images when writing with pyav, i.e., it allows `in_pixel_format="bayer_..:"` when writing. 

I would have added the same for reading, i.e., the ability to use `format="bayer_..."`, but that will have to wait until pyav adds support for these formats.

### DIFF
--- a/tests/test_pyav.py
+++ b/tests/test_pyav.py
@@ -315,3 +315,36 @@ def test_invalid_resource(tmp_path):
 
     with pytest.raises(IOError):
         iio.imwrite(tmp_path / "foo.abc", img, plugin="pyav")
+
+
+def test_bayer_write():
+    image_shape = (128, 128)
+    image = np.zeros(image_shape, dtype="uint8")
+    buffer = io.BytesIO()
+
+    with iio.imopen(buffer, "w", plugin="pyav", format_hint=".mp4") as file:
+        image[...] = 0
+        for i in range(256):
+            image[::2, ::2] = i
+            file.write(
+                image, is_batch=False, codec="h264", in_pixel_format="bayer_rggb8"
+            )
+
+        image[...] = 0
+        for i in range(256):
+            image[0::2, 1::2] = i
+            image[1::2, 0::2] = i
+            file.write(
+                image, is_batch=False, codec="h264", in_pixel_format="bayer_rggb8"
+            )
+
+        image[...] = 0
+        for i in range(256):
+            image[1::2, 1::2] = i
+            file.write(
+                image, is_batch=False, codec="h264", in_pixel_format="bayer_rggb8"
+            )
+
+    buffer.seek(0)
+    img = iio.imread(buffer, plugin="pyav")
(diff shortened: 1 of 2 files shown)


tests/test_pyav.py
```diff
@@ -315,3 +315,36 @@ def test_invalid_resource(tmp_path):
...
+        image[...] = 0
+        for i in range(256):
+            image[1::2, 1::2] = i
+            file.write(
+                image, is_batch=False, codec="h264", in_pixel_format="bayer_rggb8"
+            )
+
+    buffer.seek(0)
+    img = iio.imread(buffer, plugin="pyav")
+    assert img.shape == (768, 128, 128, 3)
```
almarklein, line 350: I expected this to be (768, 64, 64, 3), since with bayer you specify the rggb pixels individually, right?

FirefoxMetzger: Good catch. The result takes the bayer image, converts it into the default pixel format for mp4 (I believe its YUV420p), and then converts it from that to RGB while loading.

I'm actually not sure how ffmpeg debayers images, i.e., how it converts from bayer to YUV. Once I find the time, I will do some digging. In the meantime, maybe @hmaarrfk has an insight here?

almarklein: Maybe it does keep the raw resolution, otherwise you'd lose some precision for the green channel ...

hmaarrfk: This is the most annoying things about raw image sensors. I'm glad we are making it easier to people to use them.

A monochrome sensor with 1000 x 1000 pixels has 1 MP of monochrome data.

A color sensor of the same shape, almost all of which use bayer patterns---some permutation of `rggb`---, will spec 1MP. However, you only have 0.25 MP red, 0.25 MP blue, and 0.5 MP green.

Software, or some image processing IC, then interpolates these images into  the original 1000 x 1000 x 3 color space.

Is there some redundant information? Yes.
Can we do better? Probably..... but..... there is so much inertia behind JPEG (H.264 and H.265 all of which use blocks) that there is little you can do.
Eventually, when you compress, the hope is that the redundancy is found too, and the image gets smaller again.

hmaarrfk: Wikipedi has a nice 1-5 about how this is done: https://en.wikipedia.org/wiki/Bayer_filter

FirefoxMetzger: Thanks for chiming in @hmaarrfk . So if I get it right then a `(128, 128)` bayer image is actually expected to be converted into a `(128, 128)` RGB image due to interpolation? In that case, I guess what we are testing here is correct :)

hmaarrfk: Correct

almarklein: Ok, ready to merge then!

Issue #789, opened 2022-04-17
Add cv2 backend


FirefoxMetzger:
If you think about it its kind of obvious, which is why I think I never thought about bringing it up earlier. We could (and maybe should) wrap cv2 as a backend. It won't offer many (any?) new formats, but it does perform better (ie. faster) on some formats and it shouldn't be that difficult to wrap. At the same time, cv2 will not depend on ImageIO, because it itself is a C project.

Name: cv2
Focus: image processing (but with a popular IO component)
GH link: https://github.com/opencv/opencv
Website: https://opencv.org/

If you would like to see this backend added, please react to this post or leave a comment below. The more feedback I get on a backend, the more likely I am to prioritize implementing it over solving other issues.

Issue #387, opened 2018-09-26
Poor performance of imwrite on PNG images


sciunto:
Hi,

I realized a significant slow down when I switched from cv2 to imageio of my IOs.

Here is a quick benchmark.
```
import imageio
print('imageio version ', imageio.__version__)
import cv2
print('cv2 version ', cv2.__version__)

from skimage import data
import os
import time

os.makedirs('imageio', exist_ok=True)
os.makedirs('cv2', exist_ok=True)


before = time.perf_counter()
for i in range(40):
    imageio.imwrite('imageio/' + str(i) + '.png', data.camera())
after = time.perf_counter()
print('imageio png: ', after - before)

before = time.perf_counter()
for i in range(40):
    cv2.imwrite('cv2/' + str(i) + '.png', data.camera())
after = time.perf_counter()
print('cv2 png: ', after - before)

before = time.perf_counter()
for i in range(40):
    imageio.imwrite('imageio/' + str(i) + '.jpg', data.camera())
after = time.perf_counter()
print('imageio jpg: ', after - before)

before = time.perf_counter()
for i in range(40):
    cv2.imwrite('cv2/' + str(i) + '.jpg', data.camera())
after = time.perf_counter()
print('cv2 jpg: ', after - before)

```

and the results

```
imageio version  2.4.1
cv2 version  3.4.3
imageio png:  8.024044567995588
cv2 png:  0.36481686899787746
imageio jpg:  0.41310937999514863
cv2 jpg:  0.23017822099791374
```

imageio is 16 times longer than opencv. I know there might be some differences in compression levels, but still...

Are you aware of that? Any hint?

Thanks.

almarklein:
Wow, that's interesting. Imageio uses Pillow for handling png, it would be interesting to see the performance of pure Pillow use as well...

sciunto:
I didn't check with pillow. 

Based on your remark (I didn't read imageio's code so far), I found this issue https://github.com/python-pillow/Pillow/issues/1211

With `compress_level=3` for imageio on png file.
```
imageio version  2.4.1
cv2 version  3.4.3
imageio png:  0.8065581480041146
cv2 png:  0.3645203570049489
imageio jpg:  0.4112834489933448
cv2 jpg:  0.22846891700464766
```

Now, for a zero compression level
```
imageio version  2.4.1
cv2 version  3.4.3
imageio png:  0.4493131430062931
cv2 png:  0.3619267880130792
imageio jpg:  0.41116392200638074
[...]

PR #791, opened 2022-04-18
FEAT: OpenCV plugin

FirefoxMetzger:
Closes: #789 https://github.com/imageio/imageio/issues/387

This PR adds a new plugin that wraps the OpenCV library. It focuses on OpenCVs image formats (no video or GIF reading). It's main use is to allow for more performant reading and/or writing, if this is necessary.

In the future, we might want to consider adding video reading/writing support, which also includes some webcam support as well as support for IP cameras. I believe OpenCV uses ffmpeg under the hood, but uses a custom compilation of it. We would - again - likely not gain much more format support than we already have, but we might get hardware acceleration for encoding/decoding.


FirefoxMetzger:
@almarklein This should be more or less RTR. I need to add one more kwarg to read (and by extension to iter and properties), which allows the user to control the desired pixel format (colorspace). Beyond that everything should be there.

imageio/config/extensions.py
```diff
@@ -95,7 +95,13 @@ def reset(self):
...
         name="Bitmap",
         extension=".bmp",
-        priority=["pillow", "BMP-PIL", "BMP-FI", "ITK", "pyav"],
+        priority=["pillow", "BMP-PIL", "BMP-FI", "ITK", "pyav", "opencv"],
+        external_link="https://en.wikipedia.org/wiki/BMP_file_format",
+    ),
+    FileExtension(
+        name="Device-Independent Bitmap",
+        extension=".dip",
+        priority=["opencv"],
```
almarklein, line 104: Out of curiosity, why are there custom type definitions for some specific formats?

imageio/plugins/opencv.py
```diff
@@ -0,0 +1,219 @@
...
+    OpenCVPlugin.read
+    OpenCVPlugin.iter
+    OpenCVPlugin.write
+    OpenCVPlugin.properties
+    OpenCVPlugin.metadata
+
+Pixel Formats (Colorspaces)
+---------------------------
+
+OpenCV is known to process images in BGR; however, most of the python echosystem
```
almarklein, line 27:
```suggestion
OpenCV is known to process images in BGR; however, most of the python ecosystem
```

imageio/plugins/opencv.py
```diff
@@ -0,0 +1,219 @@
...
+            dtype=img.dtype,
+            is_batch=(index is ...),
+        )
+
+    def metadata(self, index: int = 0, exclude_applied: bool = True) -> Dict[str, Any]:
+        """Format-specific metadata.
+
+        .. warning::
+            OpenCV does not support reading metadata. When called, this function
+            will raise a ``NotImplementedError``.
```
almarklein, line 216: Or should it just return an empty dict then?

setup.py
```diff
@@ -185,12 +185,13 @@ def run(self):
     "simpleitk": [],
     "spe": [],
     "swf": [],
-    "tifffile": ["tifffile"],
+    "tifffile": ["tifffile", "imagecodecs"],
```
almarklein, line 188: This seems an unrelated change. Can you explain why you added it?

almarklein (commented):
Just a few comments. Nice!

imageio/config/extensions.py
FirefoxMetzger: Hm? You mean why I added extensions like `.dip`? OpenCV specifies the formats it supports and accompanying extensions in its docs ([here](https://docs.opencv.org/4.x/d4/da8/group__imgcodecs.html#ga288b8b3da0892bd651fce07b3bbd3a56)). I simply took the information from there and copied it here. I figured that if we say "we support what openCV supports" then users would expect formats that work under plain openCV to be equally recognized in ImageIO.

imageio/plugins/opencv.py
FirefoxMetzger: That is the alternative, yes. I chose to `raise` here because an empty dict could also be interpreted as "there is no metadata".  We could do warning + empty dict as well if you prefer that though.

imageio/config/extensions.py
almarklein: Woops, sorry, I added this comment to the wrong lines :)

imageio/core/imopen.pyi
```diff
@@ -25,6 +26,16 @@ def imopen(
...
 @overload
+def imopen(
+    uri,
+    io_mode: Literal["r", "w"],
+    *,
+    plugin: Literal["opencv"],
+    format_hint: str = None,
+    legacy_mode: Literal[False],
+    **kwargs
+) -> OpenCVPlugin: ...
```
almarklein, line 37: This is what I meant - why are there multiple imopen versions here, for a few specific formats?

FirefoxMetzger: This is typing overload for different plugins. In modern IDEs this will give you indicators like the one in this screenshot. Notice the 1/4 that you can also find in built-ins, numpy functions or similar.

image

(Sidenote: looking at the picture, we may wish to reorder the overloads so it starts with hinting the generic one though; I thought it would start with the last one.)

The other benefit is that validators like mypy or intelli-sense pick up on the overloads for different literals, which means that they can now statically infer that `iio.imopen(..., plugin="pyav")` will return a PyAV plugin. As a result, you will get plugin-specific autocompletion and documentation for the respective plugin, which should yield better discoverability/UX and will allow mypy to validate that plugin-specific kwargs are used correctly.

image

imageio/plugins/opencv.py
FirefoxMetzger: I will make it raise a warning an return an empty dict. I think this will work out smoother for us down the line when we want to allow the v2 API to use v3 plugins (in which case we always have to read metadata).

setup.py
FirefoxMetzger: `imagecodecs` breaks on pypy, so I am removing it again.

imageio/core/imopen.pyi
almarklein: Ok, and the reason why these 3 plugins have their own type-signature and not others because these deviate more from the generic one?

FirefoxMetzger: @almarklein Yes, basically. Each subclass of `PluginV3` that we have in the library will get a dedicated entry here so that users can discover and validate any special methods and kwargs that these plugins offer.

FirefoxMetzger:
@almarklein I rebased the PR on the current master. Any other changes we want to do here, or should we merge when the tests are green?

FirefoxMetzger:
Another rebase on master :)

almarklein:
Let's merge!

Issue #658, opened 2021-07-15
Support for RTSP video source


akshath:
Is there a plan to support RTSP source like IP Cam?

More Info: https://en.wikipedia.org/wiki/Real_Time_Streaming_Protocol

In Open CV one can do it like this 

import cv2
cap = cv2.VideoCapture("rtsp://username:pass@192.168.1.10:554/axis-media/media.amp")

FirefoxMetzger:
@akshath Have you checked imageio-ffmpeg? As far as I know, FFmpeg supports RSTP, so you should be able to read RSTP with imageio using the ffmpeg backend.

akshath:
Thank you .. that worked. Used ffmpeg-python to read frame. But was not able to use imageio-ffmpeg directly (is there an example of how to use it with rtsp?

For anyone looking on how to do it using ffmpeg-python:

<pre>
def read_frame_as_jpg(in_filename, frame_num):
    out, err = (
        ffmpeg
        .input(in_filename)
        .filter_('select', 'gte(n,{})'.format(frame_num))
        .output('pipe:', vframes=1, format='image2', vcodec='mjpeg')
        .run(capture_stdout=True)
    )
    return out

frame = read_frame_as_jpg('rtsp://username:pass@192.168.1.10:554/axis-media/media.amp', 0)
im = imageio.imread(frame)
plt.imshow(im)
plt.show()
</pre>

almarklein:
Have a look at the `input_params` argument to imageio_ffmpeg.read_frames()`. I guess the filter params in the sample below should be passed that way (as ffmpeg cli args).

Issue #642, opened 2021-06-15
Virtualize Request's file object


FirefoxMetzger:
While writing the new API, we noticed that passing a request object around may allow plugins to interact with each other, e.g., by closing the file or by seeking in it and not tidying up after themself. Unfortunately, a quick `seek(0)` before handing a request to a new plugin won't solve this. It breaks a few tests, but I didn't dig into the why.

Request also supports a `firstbytes` object, which allows plugins to obtain a cached version of the file header to, for example, determine file-type. This attribute is only sparsely used by plugins and perhaps a little unintuitive, so we would like to refactor it as well.

As a result, a cleaner way to handle this is to provide virtual file objects to new plugins upon `Request.get_file()`. This virtual object should provide a read-only view into the file while preventing a plugin from corrupting the "raw" file. In addition, it should support caching of the raw file in order to, for example, allow fast lookup of header/metadata information or seeking of non-seekable URIs (e.g. a web URL).

Be advised that this is a hard issue. If you do like a challenge or have questions/suggestions, please leave a comment :)

FirefoxMetzger:
This one is related to #616 .

FirefoxMetzger:
Related: #413

Issue #616, opened 2021-03-31
Replace request.SeekableFileObject with a mmap-ed buffer


FirefoxMetzger:
I've noticed that for some of the input types we perform manual buffering via (`SeekableFileObject`). One problem that I can see in the current implementation is that there the buffer is purely kept in memory, which may be undesirable behavior for *very* large files.

I was wondering if buffering inside a memory-mapped container - based on `mmap` and using a `tmpfile` -  could be a more efficient solution, or if it was explored and discarded previously.

It could also be a nice opportunity to expose our codebase to mmap, which promises some speed increase for I/O due to fewer system calls (potentially) and one less copy of the data..

almarklein:
No, this has not been attempted or explored. Sounds like something that could work!

Issue #413, opened 2018-11-23
Feature Request: Memory Mapping


ricardo-re-azevedo:
Is support for memory mapped objects in the roadmap? As image data size is beginning to outpace memory size it is becoming unfeasible for many to load full image sets into RAM. For example, rather than using `_tifffile.TiffFile(f, **kwargs)` using`_tifffile.tifffile.memmap(f , **kwargs)` to return a memmapped array.

almarklein:
That's interesting. This is not on the roadmap. And I would probably not be eager to add such a feature unless it's possible to actually do it with a significant portion of the plugins.

> As image data size is beginning to outpace memory size it is becoming unfeasible for many to load full image sets into RAM.

I initially read this the other way around (there's not need to write things to the FS, can we keep it in RAM?). I guess it goes both ways ...

PR #792, opened 2022-04-20
BUG: reset request when iterating over plugins

FirefoxMetzger:
While testing the ability to directly decode videos from byte strings I noticed that there is a bug in `imopen`. 

Plugins clean up after themself and, in the process, call `request.finish()`, which backfires when we iterate over all plugins since the next plugin in the list receives a finished request, which had it's `request._bytes` garbage collected. This PR fixes this by creating a fresh request for every plugin being tested and adds a regression/unit test for this.

It also removes a duplicate entry in our known extensions that I found :)

### DIFF
--- a/imageio/config/extensions.py
+++ b/imageio/config/extensions.py
@@ -1499,11 +1499,6 @@ extension_list = [
     ),
     FileExtension(
         name="QuickTime / MOV",
-        extension=".mp4",
-        priority=["pyav"],
-    ),
-    FileExtension(
-        name="QuickTime / MOV",
         extension=".psp",
         priority=["pyav"],
     ),
--- a/imageio/core/imopen.py
+++ b/imageio/core/imopen.py
@@ -132,6 +132,7 @@ def imopen(uri, io_mode, *, plugin=None, format_hint=None, legacy_mode=False, **
                 return plugin(request, **kwargs)
 
         else:
+            request.finish()
             raise ValueError("The `plugin` argument must be a string.")
 
         try:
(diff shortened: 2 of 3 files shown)


FirefoxMetzger:
@almarklein RTR.

almarklein:
Doesn't this mean that the file is opened ones for each newly created request object? If so, maybe we can do something like `plugin.cancel()`? ... would be new API though ...

FirefoxMetzger:
Yes it does. It should only happen rarely though and the two most common cases are either when reading bytes (where "reopening" is fast) or when reading something we can't comprehend (where we will fail anyway, just a bit slower than before).

That said, `Request` is indeed due for an overhaul and this here is more of a workaround than a permanent solution. We ran into the issue of resetting a request before and concluded that the best (but effortful) solution would be to virtualize the file object that we pass to a plugin. This would also benefit or close a bunch of related issues (https://github.com/imageio/imageio/issues/616, https://github.com/imageio/imageio/issues/642, https://github.com/imageio/imageio/issues/413), and would allow us to use pyav to support streams (which would close https://github.com/imageio/imageio/issues/658 among others). 

Since there are so many places where we would benefit this is high on my priority list, but it will require me being able to sit down and sink some serious time into how the new `Request` object should look like before making a proposal. On the bright side though, it is an internal object, so we should be quite free to rework it.

---

Going back to the issue at hand, we could also remove the part of the request object that removes the bytes:

https://github.com/imageio/imageio/blob/7ecea02d7c5633c1aa8afe39af6f4f0de59d8ae3/imageio/core/request.py#L569-L570

and then simply do `request.get_file().seek(0)`. This would solve the problem I am targeting here (that we can't find the correct plugin for `bytes` if we don't get it right on the first try) and avoid reopening. The downside is that it may not always work and that I haven't thought about the consequences of calling `seek` on other ImageResources (e.g., non-seekable ones like HTTP ).

almarklein:
Ok, if you think that re-opening files would be rare, and this is a temporary solution, then better go with the safe solution.

PR #794, opened 2022-04-21
do not pass optional keyword arguments by position

cgohlke:
This is enforced in tifffile > 2022.4.8.


FirefoxMetzger:
Thanks @cgohlke! LGTM assuming the tests are happy, too.

PR #795, opened 2022-04-21
Fix test for tifffile > 2022.4.8

cgohlke:
In tifffile > 2022.4.8, the `is_imagej` property returns `bool`.


FirefoxMetzger:
Nice catch @cgohlke :) I'll run the test suite and will merge once green.

Issue #802, opened 2022-04-29
imageio v3 api fails to read all frames in a MetaMorph stk file


anntzer:
The following zip file contains a single .stk image stack (https://support.moleculardevices.com/s/article/MetaMorph-Software-TIFF-and-STK-file-formats) with 2 frames: [test.zip](https://github.com/imageio/imageio/files/8589321/test.zip) (I zipped it because github doesn't support uploading stks but does support zips...)
imageio v2 volread reads it fine, but the v3 API only sees the first frame:
```
In [17]: imageio.volread("test.stk").shape
Out[17]: (2, 290, 364)

In [18]: imageio.v3.imread("test.stk", index=...).shape
Out[18]: (1, 290, 364)

In [19]: len([*imageio.v3.imiter("test.stk")])
Out[19]: 1
```
(Note that stk files are explicitly listed as supported by the tifffile module.)

FirefoxMetzger:
Interesting find, this looks like a bug. I just tried it locally and I can reproduce this on master. I don't have an immediate insight into why this happens, but I will have a look later this afternoon or over the weekend.

cgohlke:
FWIW, STK files are not regular TIFF. They store "frames" not in separate TIFF pages but contiguously after the image data of the first page, similar to some ImageJ hyperstacks. Use the tifffile series interface to read the image data from all STK frames.

https://github.com/cgohlke/tifffile/blob/525a9a89e74e136eee6f6ffed25825900399f969/tifffile/tifffile.py#L297-L299

FirefoxMetzger:
I've had a look over the weekend and I found the code that causes the difference in behavior:

https://github.com/imageio/imageio/blob/140707ae3c6eace7cf9131662197cc050d34cba3/imageio/plugins/tifffile.py#L437-L448

v2 calls the legacy plugin with image mode `v` (which we are deprecating), whereas v3 calls the plugin with no explicit image mode. I will have a think as to how to resolve this elegantly.

@cgohlke Are you aware of a usage of TIFF that uses multiple series objects and where each series object contains a volume?

Meanwhile, @anntzer would it be okay if we add the test image to our test suite and distribute it together with the tests? This would mean that the image would have to be licensed under something permissible like CC0 or a variant thereof.

cgohlke:
> Are you aware of a usage of TIFF that uses multiple series objects and where each series object contains a volume?

Those can be found in OME-TIFF files. MicroManager and tifffile write such files.

[...]

PR #805, opened 2022-05-02
BUG: correctly read STK volumes in v3

FirefoxMetzger:
Closes: #802 

This PR changes the behavior of tifffile so that we default to reading volumes from STK files. 

It also adds a regression test for this, but it is commented out for now, until we get permission to reuse the file in our test suite. If we don't get it and I find some spare time this week, I will write a piece of code that generates a problematic STK for us to test with.


almarklein (approved):
LGTM. I presume you want to leave this open until we can enable the test?

FirefoxMetzger:
@almarklein Yep, that's exactly right.

FirefoxMetzger:
I have added a `skipif` for the ffmpeg test that is currently failing on pypy until I have time to investigate where the OOM error is coming from. It shouldn't be related to this PR as it just touches tifffile code, but it could be related to how pypy manages memory.

Issue #810, opened 2022-05-10
Apparent bug in `imageio.v3.imwrite`


bwohlberg:
There appears to be a bug in the v3 API related to saving TIFF files. The following simple test script
```python
import numpy as np
import imageio.v3 as iio

v = np.zeros((32, 32), dtype=np.uint16)
iio.imwrite("v.tif", v)
```
fails with error `ValueError: Image must be 3D, or 4D if each voxel is a tuple.` The same example functions correctly when using `import imageio.v2 as iio`. This test was run with the `tifffile` package installed and `imageio` 2.19.1.

FirefoxMetzger:
Thanks for the report; I can reproduce this on master. 

Looks like I missed a case inside a `if/elif/else` block of the legacy plugin wrapper. It should be an easy fix, so I will submit a PR that fixes this soon.

PR #811, opened 2022-05-10
BUG: don't treat im_mode ? as im_mode V

FirefoxMetzger:
Closes: #810 

This PR fixes a bug in the legacy plugin wrapper where the branch for requests with arbitrary image mode (`?`) were treated like requests with volume image mode (`V`).

It also adds a regression test for this.


FirefoxMetzger:
@almarklein RTR

almarklein:
I wonder why that line was commented and not made active ... Well, at least we have an actual use-case to test the desired behaviour against.

FirefoxMetzger:
> I wonder why that line was commented and not made active

Because I didn't account for the case of creating a request with `mode=r?` and instead only considered modes `r[iIvV]` so I thought that `else` would only happen in a specific case which I added as a comment for clarity.

PR #813, opened 2022-05-12
Add stacklevel=2 to some warnings.

Carreau:
This is really important to do as with this value, the warning message
point to the file/line of the caller, not actually where the warning is
issued. This mean that when I see this error in CI, I directly know
where to fix it in downstream projects (in my case napari).

Otherwise It takes me forever to find where the caller is, and I'm
likely to not fix it.


FirefoxMetzger:
Good catch @Carreau and thanks for the PR. I will let CI have a go, and once it's happy I'm happy to merge as well.

FirefoxMetzger:
Thanks to you for contributing :)

PR #815, opened 2022-05-18
Include py.typed in package

SimonSegerblomRex:
[mypy](https://github.com/python/mypy) complains about
```python3
import imageio.v3 as iio
```
with
```bash
error: Skipping analyzing "setuptools": module is installed, but missing library stubs or py.typed marker
```

I don't know much about mypy or packaging, but I verified that this fix solves the issue when installing locally with
```bash
python3 -m pip install .
```

This is covered in the [mypy docs](https://mypy.readthedocs.io/en/stable/installed_packages.html#installed-packages). There it's also mentioned that
>  If you use setuptools, you must pass the option `zip_safe=False` to `setup()`, or mypy will not be able to find the installed package.

– does this apply to imageio?

**Note:**
If you are using
```bash
python3 -m pip install -e .
```
you won't realize this issue.


FirefoxMetzger:
Thanks for the PR @SimonSegerblomRex and good catch on spotting the missing include. There is actually another include that is missing which I will push in a bit.

Regarding the `zip_safe=False` flag, I'm not really sure if we need it. We do package using setuptools, but  https://github.com/python/mypy/issues/8802#issuecomment-628249236 suggests that this flag is mainly to support legacy behavior (egg install rather than wheel install). @almarklein Do you think we should worry about the use case of installing imageIO as an egg (instead of the usual `pip install ...`)?

almarklein:
It looks like in theory the `zip_safe` flag would not be needed, but you never know what users do ... I think it would not hurt to add the flag :)

FirefoxMetzger:
@almarklein Okay, I've added it. Will merge once CI is happy.

Issue #798, opened 2022-04-25
Modernize the Request object


FirefoxMetzger:
This issue discusses ImageIO internals. Unless there are regressions, users won't notice any impact.

In #792 we discussed that it may be time to modernize the `Requests` object that sits between the user's input and a particular plugin. To start with, here is a list of things it needs to be capable of (if anything is missing please commend and I'll add it to the top message):

- (current) let plugins access various ImageResources as buffered FileLike objects
- (current) let plugins access various ImageResources as local files (if the backend doesn't support filelike)
- (current) classify the ImageResource (bytes, file, remote, standard image, ...)
- (current) provide access to the ImageResources path (for local files)
- (current) provide access to the file extension associated with the ImageResource
- (current) provide access to the mode (read/write) and (for legacy support) the image mode (single, multi, vol)
- (new) have aforementioned classification be not private (currently we are referencing internal attributes to get around this)
- (new) a reset mechanism to let a new/next plugin use the request (in order)
- (new) a mechanism for shared ownership (between imopen, a plugin, and any legacy get_readers) that avoids closing the file as long as at least one entity still uses the request
- (new) support for streams and other requests that can't be represented as FileLike
- (new) a non-hacky way to handle dictionaries (we have hacky support for DICOM, but we should either deprecate or fully support it)
- (new) remove `firstbytes` (we should still buffer it, but this ought to happen under the hood without the plugin knowing)

There are two other features that may need more discussion than the the ones mentioned above, because they involve new dependencies:

- (new) use `requests` to handle HTTP. Should we directly depend on it or make it optional, how about writing - PUT, PUSH - auth, etc.
- (new) use `fsspec` to handle various remote locations (S3, gcloud, github - we already do this in the tests and its awesome, ...) Should this be optional or a main dependency, ...

I will draft an interface suggestion for this today after work and will add it as a new comment, which I will reference from here.

FirefoxMetzger:
Here is a first suggestion for a new API that satisfies the above points

```python
from typing import Dict, Optional, BinaryIO
from ..typing import ImageResource

class Request:
    is_filelike:bool
    is_bytes:bool
    is_uri:bool
    is_folder: bool
    # maybe add is_screen, is_device, is_clipboard

    mode:str
    raw_resource:ImageResource
    extension:Optional[str]
    resource_name:Optional[str]  # Normalized name

    def __init__(
        self,
        resource: ImageResource,
        mode: str,
        *,
        extension: str,  # the new format_hint
[...]

PR #816, opened 2022-05-25
MAINT: remove unneeded image_mode references

FirefoxMetzger:
This PR cleans up the codebase by refactoring references to `request.mode.image_mode`. The goal is to remove all references to image_mode and ultimately remove it from the codebase. This is a first step towards refactoring the request object into what we discussed in #798 .


FirefoxMetzger:
This PR is a bit old, but it is RTR as well.

For this one, I will not merge it directly, but instead wait for @almarklein or someone else to look this over before I merge.

imageio/core/legacy_plugin_wrapper.py
```diff
@@ -162,7 +162,7 @@ def legacy_get_writer(self, **kwargs):
         self._request._kwargs = kwargs
         return self._format.get_writer(self._request)
 
-    def write(self, ndimage, **kwargs):
+    def write(self, ndimage, is_batch=None, **kwargs):
```
almarklein, line 165: Needs documentation too?

almarklein (commented):
One question. Previously the user could specify single-image, multi-image or auto. This is still the same, except now auto is resolved in `write()`. How would this work for `read`? Because there we want to resolve auto in the plugins, e.g. to check tiff metadata to determine whether its an image series or volume.

imageio/core/legacy_plugin_wrapper.py
```diff
@@ -179,41 +179,34 @@ def write(self, ndimage, **kwargs):
...
+            is_batch = True
+        elif ndimage.ndim == 2:
+            is_batch = False
+        elif ndimage.ndim == 3 and ndimage.shape[2] < 5:
+            is_batch = False
+        else:
+            is_batch = True
+
+        if not is_batch:
+            ndimage = np.asarray(ndimage)[None, ...]
```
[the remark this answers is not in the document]
FirefoxMetzger: If the provided ndimage is not flagged as a batch of images we prepend a unitary dimension, i.e., we turn it into a "batch" containing a single image. This way we can treat both cases in the same way and combine code paths.

imageio/core/legacy_plugin_wrapper.py
FirefoxMetzger: Yep. I added the docs :)

PR #823, opened 2022-06-30
DOCS: Change mode to "I" in optimization example

lharri73:
Fix a small typo in the documentation.


FirefoxMetzger:
Good catch @lharri73 and thanks for directly fixing it.

PR #824, opened 2022-07-01
MAINT: expose v3 plugins through v2 API

FirefoxMetzger:
The aim of this PR is to refactor the core and expose new plugins via the "old" v2 API. This way, we can deprecate and remove plugins that have been ported to v3 over time and we can also allow the v2 API to stay relevant and have access to new plugins.

### DIFF
--- a/imageio/core/imopen.py
+++ b/imageio/core/imopen.py
@@ -134,14 +134,10 @@ def imopen(
             def loader(request, **kwargs):
                 return config.plugin_class(request, **kwargs)
 
-        elif not legacy_mode:
+        else:
 
             def loader(request, **kwargs):
                 return plugin(request, **kwargs)
-
-        else:
-            request.finish()
-            raise ValueError("The `plugin` argument must be a string.")
 
         try:
             return loader(request, **kwargs)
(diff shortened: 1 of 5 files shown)


FirefoxMetzger:
Thanks for the report @djhoese ! Tracking issues on merged PRs is a bit cumbersome - I turned it into an issue: https://github.com/ssec/sift/issues/372

Let's discuss there :)

cris-deepcell:
I came here from the release notes, trying to figure out what changed.

I have kept using `imageio.v2` for reading PNG files, because it returned 16-bit PNGs as 16-bit arrays. But unfortunately with it stopped doing so, instead returning a 32-bit signed integer array just like `imageio.v3` does.
I am not sure if this change is the culprit, but it's the most likely one. If so, the incompatibility started with v2.28.0, which includes this PR.

Upgrading to [v2.31.0](https://github.com/imageio/imageio/releases/tag/v2.31.0) and Pillow to v10.0.0 allows the v3 plugin to read 16-bit PNGs as a 16-bit array, so that fixes things again.
[Posting this here in the hopes that it helps other lost souls like me.]

Issue #820, opened 2022-06-06
Strict mypy support for imread


dantebben:
When using `imageio.v3.imread`, `mypy` complains that there is no return type, with the following error:

```bash
error: Call to untyped function "imread" in typed context
```

The current work around is to do something like the following:

```python
img:  npt.NDArray[np.float_] = imageio.v3.imread('my_image.png')  # type: ignore[no-untyped-call]
```

Without the `# type: ignore`, `mypy` returns an error.

This is using imageio v2.19.3 and mypy v0.961

To reproduce the issue, with `imageio`, `numpy`, and `mypy` installed:

1. create a file, `test_it.py` with the following:

   ```python
   import imageio
   import numpy as np
   import numpy.typing as npt

   img: npt.NDArray[np.float_] = imageio.v3.imread("my_image.png")
   ```

2. run `mypy` in strict mode:

   ```bash
   $ mypy --strict test_it.py
   ```

3. The test fails, with error described above.

FirefoxMetzger:
Interesting find. The return type for `iio.v3.imread` is defined here:

https://github.com/imageio/imageio/blob/aba1eecb29cbbbfbd688ccba878c67d53b59204a/imageio/v3.pyi#L9-L17

We are not using `npt.NDArray`, because it doesn't exist on older versions of numpy and because it doesn't seem desirable to require numpy 1.20+ just for `numpy.typing`. Unfortunately, I won't have free time this month to investigate the issue, but if we can identify a solution, I can review and merge it.

As a sidenote: `npt.NDArray[np.uint8]` might be a better choice than np.float_ . Most images store uint8, some uint16, and almost none store floating point (except for some HDR and video formats).

FirefoxMetzger:
I just tried to reproduce this. If I install imageio from pypi then I get the result that you describe @dantebben . When I use a local dev installation (`pip install -e .`), however, mypy is happy with the defined return types.

PR #825, opened 2022-07-01
BUG: include PYI files in package

FirefoxMetzger:
Closes: #820 

This PR adds the PYI files to imageio's wheel. This way, mypy should be able to pick up on type hints and this should solve #820


FirefoxMetzger:
@almarklein RTR :)

Issue #817, opened 2022-06-02
RecursionError on Python 3.11


cgohlke:
Just a heads up. I discovered this error while testing of tifffile on Python 3.11:

```
Python 3.11.0b3 (main, Jun  1 2022, 13:29:14) [MSC v.1932 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> from imageio.v2 import imwrite
>>> imwrite('test.tif', [[0]])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Python311\Lib\site-packages\imageio\v2.py", line 238, in imwrite
    with imopen(uri, "wi", **imopen_args) as file:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Python311\Lib\site-packages\imageio\core\imopen.py", line 207, in imopen
    candidate_plugin = config.plugin_class
                       ^^^^^^^^^^^^^^^^^^^
  File "C:\Python311\Lib\site-packages\imageio\config\plugins.py", line 100, in plugin_class
    module = importlib.import_module(self.module_name, self.package_name)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Python311\Lib\importlib\__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1206, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1178, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1138, in _find_and_load_unlocked
  File "C:\Python311\Lib\site-packages\imageio\plugins\__init__.py", line 54, in __getattr__
    return importlib.import_module(f"imageio.plugins.{name}")
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  
<snip>

  File "C:\Python311\Lib\importlib\__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #826, opened 2022-07-01
BUG: resolve infinite recursion in python 3.11 (and add CI for pypy-3.9)

FirefoxMetzger:
Closes: #817 

This PR includes pypy 3.9 in our test matrix. It also adds a refactor of the plugin submodule that should fix the bug reported in #817

Edit: I removed the addition of cpython 3.11-dev to CI from this PR because pillow does not seem to have a prebuilt wheel for this python version yet. Building from source fails without us digging into providing the required dependencies on GH actions, which is IMO beyond the scope of this PR. So - for now - there is no CI for python 3.11, but I am very happy to revisit once upstream dependencies have adopted 3.11.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -81,7 +81,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest", "macos-latest", "windows-latest"]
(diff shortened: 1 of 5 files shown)


FirefoxMetzger:
@almarklein RTR :) Let's chip away at some of our technical dept 💪

PR #830, opened 2022-07-07
BUG: allow multiple reads from imopen

FirefoxMetzger:
While answering a question on SO I found a bug in our pyav plugin that prevents calling `PyAVPlugin.read(...)` more than once.

The reason was that pyav doesn't allow re-assigning or changing the threading model (`thread_type`). Even if we try to assign the current value it will break. The same happens for the thread count. This PR addresses that by checking the value first and only setting it if it is different from the current value.

### DIFF
--- a/imageio/plugins/pyav.py
+++ b/imageio/plugins/pyav.py
@@ -392,8 +392,17 @@ class PyAVPlugin(PluginV3):
 
             return frames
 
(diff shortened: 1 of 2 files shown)


FirefoxMetzger:
@almarklein RTR :)

almarklein (approved):

PR #831, opened 2022-07-10
MAINT: Move FreeImage plugin into own repo

### LINKED ISSUE #809: support freeimage plugin instalaltion using pip install imageio[freeimage]
At the moment the freeimage plugin needs to be installed using either `imageio_download_bin freeimage` or `imageio.plugins.freeimage.download()`. This has to be run after the python environment has been set up with pip.

This is not ideal as one may want to be able to install the plugin when setting up the python environment with pip 
and not later at run time or in a post-installation script. One would also want the executable to be downloaded in the imageio module folder instead of being download in the user local folder. 
In contrast the plugin `gdal` can be installed `pip install imageio[gdal]` and thus is would be great to support `pip install imageio[freeimage]` as well.

FirefoxMetzger:
Closes: https://github.com/imageio/imageio/issues/809

This PR moves the FreeImage plugin into its own repository located here: https://github.com/imageio/imageio-freeimage This is done for several reasons (the repo's readme has some details), but mostly because it makes license implications cleaner. 

It closes #809 by removing any option to install freeimage through ImageIO directly. This is a bit unsatisfactory, because the issue asks for a more convenient way to install via `pip install imageio[freeimage]` without any post-install steps, but appears - unfortunately - necessary to correctly deal with freeimage's license. On the upside, we can now investigate packing freeimage into `imageio-freeimage`'s wheel and allow easy install via `pip install imageio-freeimage` :)


FirefoxMetzger:
@almarklein This is quite a harsh break with little to no warning in advance. I wonder what would be a good way to create a sensible error message / deprecation notice to let people know they have to `pip install imageio-freeimage` now to get any freeimage functionality.

FirefoxMetzger:
@almarklein Asside from a way to inform users on how to migrate this should be RTR.

FirefoxMetzger:
> we've a baby coming soon

Wow, congratulations! 🎉 🎂 

> I might be (even) slower than usual to respond

On this note, do you think it could be worthwhile to start actively looking for 1-2 more people that would be willing to help maintain ImageIO?

> We could keep a list of these formats in imageio, so we can detect when this is happening.

We still have this list for the recognized/supported formats. I kept this so that our docs will continue to list the formats, but point to `imageio_freeimage`. I will think if I can come up with a way to point users to `imageio_freeimage` as an option when something fails and that library could solve it.

FirefoxMetzger:
I am closing this PR because we will have to use a different, less radical, strategy of migrating away from FreeImage.

PR #832, opened 2022-07-10
TEST: use tmp folder for written image

FirefoxMetzger:
This is a small change. One of our tests was creating an image in the current working directory instead of using a temporary folder. This PR fixes that.


FirefoxMetzger:
@almarklein RTR

PR #835, opened 2022-07-13
Fix typos in v3 narrative docs.

anntzer:


FirefoxMetzger:
@anntzer LGTM thanks for taking the time to fix these :)

PR #838, opened 2022-07-19
FEAT: enable HTTP based streams in pyav

FirefoxMetzger:
PyAV/FFmpeg supports [DASH streams](https://en.wikipedia.org/wiki/Dynamic_Adaptive_Streaming_over_HTTP), which we currently don't. This PR addresses that by making pyav use the raw HTTP URL instead of using `request.get_file()`.


FirefoxMetzger:
Since @almarklein is busy at the moment, I will leave this PR open until Friday for anybody that wishes to review it and if I don't receive any feedback I will merge it in :)

PR #839, opened 2022-07-19
BUG: frame-level metadata and duration in pyav

FirefoxMetzger:
This PR fixes a bug in the pyav plugin that returns the container metadata instead of returning frame-level metadata. It also exposes the video duration (in seconds) when reading global metadata for a video.


FirefoxMetzger:
This PR is RTR, but @almarklein is currently on vacation. I will leave it open until Friday and then merge unless somebody wants to review and leave some comments.

PR #840, opened 2022-07-20
WIP: Expose the rotate property of pyav to rotate h264 videos

hmaarrfk:
You can write a simple test writing to a file, then reading the rotate property of the resulting file.

It is visible by the `metadata` property of the file.

The data itself doesn't change, but it is more of a hint for players to orient the video correctly.

These are probably the best tutorials i read:
https://stackoverflow.com/questions/25031557/rotate-mp4-videos-without-re-encoding
https://superuser.com/questions/1393880/edit-mp4-hex-data-to-vertically-or-horizontally-flip-video-in-a-lossless-manner

Ultimately, the support is hit and miss on different players. I wouldn't rely on it. I'll add this to the docstring.

xref: https://github.com/PyAV-Org/PyAV/issues/1012

### DIFF
--- a/imageio/plugins/pyav.py
+++ b/imageio/plugins/pyav.py
@@ -539,6 +539,7 @@ class PyAVPlugin(PluginV3):
         out_pixel_format: str = None,
         filter_sequence: List[Tuple[str, Union[str, dict]]] = None,
         filter_graph: Tuple[dict, List] = None,
+        rotate: Optional[str] = None,
     ) -> Optional[bytes]:
         """Save a ndimage as a video.
 
@@ -572,6 +573,14 @@ class PyAVPlugin(PluginV3):
             contains a (named) set of nodes, and the second dict contains a set
             of edges between nodes of the previous dict. Check the (module-level)
             plugin docs for details and examples.
+        rotate : str
+            If available by the codec, set the metadata to indicate rotation.
+            This string is typically `'0'`, `'90'`, `'180'`, or `'270'`.
+            Restrictions on this parameters are imposed by the file format and
+            should be used with caution.
+            When reading back a stream, one can read the ``metadata``.
+            A rotation of "0" degrees will not appear as a key in the metadata
+            since it is implied.
 
         Returns
         -------
@@ -598,7 +607,12 @@ class PyAVPlugin(PluginV3):
 
         if self._video_stream is None:
             self._init_write_stream(
-                codec, fps, ndimage.shape, in_pixel_format, out_pixel_format
+                codec,
+                fps,
+                ndimage.shape,
+                in_pixel_format,
+                out_pixel_format,
+                rotate=rotate,
             )
         stream = self._video_stream
 
@@ -688,6 +702,7 @@ class PyAVPlugin(PluginV3):
         shape: Tuple[int, ...],
         in_pixel_format: Optional[str],
         out_pixel_format: Optional[str],
+        rotate: Optional[str],
     ) -> None:
         """Initialize encoder and create a new video stream.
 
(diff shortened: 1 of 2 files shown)


FirefoxMetzger:
Interesting, thanks for the PR @hmaarrfk . A few questions:

- Are you only interested in writing, or are you also planning to add functionality to the reader?
- This looks like a "community agreed" standard of setting a metadata value to rotate instead of a codec-specific feature that is separate from the stream's metadata. Is that correct? If so, maybe the better way to expose this could be to expose a `metadata:dict[str, str]` kwarg in the writer which can set `rotation` among other things. What do you think?

Let me know when you feel that the PR is ready to be reviewed :)

hmaarrfk:
The reader can already read the property.

Do you have documentation i can point to for the community agreement? Maybe it is already in the stack overflow posts i shared

hmaarrfk:
It seems that what you are saying is different than https://ffmpeg.org/ffmpeg-bitstream-filters.html#h264_005fmetadata

Do you know to use the bitstream filters (with the command line or not)?

Are you able to see an effect on a video with a reasonable video player?

FirefoxMetzger:
> Do you have documentation i can point to for the community agreement?

Oh no, I think we misunderstood each other. I didn't know that rotation was a metadata field that could be passed to a video and have meaning, so my first response was to assume/ask if this is a similar situation to EXIF.

As far as I understand, EXIF wasn't part of the original standard of JPEG or PNG but was simply appended because the community found it useful. Nowadays most readers consume EXIF correctly and, as such, it is pretty much like a part of the standard. I was wondering if this is similar to the `rotation` field in h264 metadata, but it looks like that assumption is wrong.

> Do you know to use the bitstream filters (with the command line or not)?

I have never used bitstream filters before, no. It does look like an interesting feature though, so I will check if pyav exposes it and if we could do so as well.

> Are you able to see an effect on a video with a reasonable video player?

I didn't get a chance to play around with the rotation flag yet, but I will report back once I get the chance to try it out.


> > Are you only interested in writing, or are you also planning to add functionality to the reader?
>
> See the expanded tests on reading. They mirror my experiments. I believe that videos that are stored in this way are rotated > upon playback by the viewers I've tested (totem, vlc, slack, preview on macos).

Oh, I was thinking of us rotating the image when reading similar to how we can rotate the image if we encounter the EXIF rotation flag in pillow.

> I do find the present API strange. It seems to take a lot of arguments every time you want to send data. I understand the "state free" nature of things, but I would have expected an initialization method.

If you have suggestions on how to improve it, please let me know; I'm always happy to make the library more user-friendly :)

What would be the ideal way of writing a video?

hmaarrfk:
> I didn't get a chance to play around with the rotation flag yet, but I will report back once I get the chance to try it out.

You have have a look at the videos I generated in the "Investigation" section of PyAv https://github.com/PyAV-Org/PyAV/issues/1012

hmaarrfk:
> What would be the ideal way of writing a video?

I'm really not sure. I'm trying to design it too. There are two contracting requirements:

1. Good ways of encoding videos "offline".
2. Good ways of encoding videos "online".

I'm in camp #2.

hmaarrfk:
I think there are also alot of fancy things that you can do with MP4s. Kinda like TIFFs. You can have multiple streams, and they can either interact or not.

FirefoxMetzger:
@hmaarrfk I've created https://github.com/imageio/imageio/issues/841 to continue the discussion on a better API and to avoid derailing this PR :)

> I think there are also alot of fancy things that you can do with MP4s. Kinda like TIFFs. You can have multiple streams, and they can either interact or not.

Yes you could, but in practice most players ignore any video stream after the first. I've also never seen a situation where multiple video streams are desirable. Multiple audio streams can be useful, e.g., for providing a movie in different languages, but we currently don't support audio ... just video :)

FirefoxMetzger:
@hmaarrfk I think #853 covers this use-case as well (an example can be found in the opening comment). Shall we close this one in favor of #853 ?

PR #845, opened 2022-07-25
FEAT: add is_batch kwarg to pillow

FirefoxMetzger:
This PR adds a new kwarg to `PillowPlugin.write` called `is_batch`. Currently, we try to figure out if the passed ndarray is an image or an image batch by looking at the shape or mode. This isn't always correct due to the aliasing of shapes, e.g., a `(5, 5, 3)` image could be a 5x5 RGB image or a batch of 5 5x3 grayscale images.

This PR hence adds `is_batch` as an optional kwarg to allow the user to explicitly pass this information if relevant.

### DIFF
--- a/imageio/plugins/pillow.py
+++ b/imageio/plugins/pillow.py
@@ -248,17 +248,24 @@ class PillowPlugin(PluginV3):
 
         Parameters
         ----------
-        image : ndarray
-            The ndimage to write.
-        mode : {str, None}
+        image : ndarray or list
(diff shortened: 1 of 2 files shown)


FirefoxMetzger:
This should be RTR. I will keep the same procedure as last week until @almarklein comes back from vacation. I will leave the PR unmerged for anyone to review and if there are no open discussion points by Friday, I will merge :)

Issue #829, opened 2022-07-06
how to write a gif making fro multiple images


harper1999:
previously, imageio.mimsave has the param frames receiving a list. Not availbale for v3.imwrite anymore

FirefoxMetzger:
@harper1999 Could you share a small example to demonstrate the problem?

The following works fine on my machine:

```python
>>> import imageio.v3 as iio
>>> frames = iio.imread("imageio:newtonscradle.gif")
>>> iio.imwrite("foo.gif", frames)
>>> iio.imwrite("foo.gif", frames, loop=0) 
>>> frame_list = [x for x in frames]
>>> iio.imwrite("foo2.gif", frame_list, loop=0) 
>>> type(frame_list)
<class 'list'>
```

It produces two GIFs as expected:

![foo](https://user-images.githubusercontent.com/4402489/177612402-246e2bd0-5e41-4f34-816b-19d0692f23c7.gif)

![foo2](https://user-images.githubusercontent.com/4402489/177612416-d6fa5392-d336-4fb5-9a0d-0015c37cff99.gif)

FirefoxMetzger:
@harper1999 Did that answer your question? If so, I'd suggest we close this issue; if not, could you describe your problem in more detail?

FirefoxMetzger:
I will close this question due to inactivity. However, if the problem persists feel free to leave a note and I'm happy to reopen :)

harper1999:
# trying to make all pngs into 1 gif
 frames = [imageio.v3.imread(file) for file in files]  #  multiple pngs in the files
    imageio.mimsave(gif_name, frames, format='GIF-PIL', duration=2)

how does iio.imwrite work in this situation?

FirefoxMetzger:
I would use `imageio.v3.imwrite` for this, because it can handle a list of frames. `imageio.imwrite` is (in ImageIO 2.x) an alias for `imageio.v2.imwrite` which can only handle a single frame. Using the v3 API you would do:

```python
import imageio.v3 as iio

frames = [iio.imread(file) for file in files]
iio.imwrite(
    gif_name,  # name of the resulting file, e.g. fancy.gif
    frames,  # the images that make up the gif
    duration=2, #  show each image for 2 seconds
    loop=0,  # number of repeats, loop=0 means repeat forever
)
```

Note that all images/frames have to have the same shape. If not, you will get an error that the shapes are mismatched which GIF does not support.

harper1999:
[...]

PR #848, opened 2022-07-25
MAINT: add missing option to imwrite type hints

FirefoxMetzger:
While helping debug a script in a recent discussion (https://github.com/imageio/imageio/issues/829#issuecomment-1194134027) we've noticed a bug in ImageIO's typehints: the option to pass `List[ArrayLike]` as frames was missing. This PR adds that :)


FirefoxMetzger:
This one is RTR as well.

It's only a small change, but since it affects something that us user-facing I will leave it open until Friday for anybody who wishes to comment on it :)

PR #850, opened 2022-07-26
DOC: add a migration note about pillow squeezing

FirefoxMetzger:
This PR adds a note to the docs about a behavior change in the pillow plugin related to squeezing the last dimension. The legacy plugin used to squeeze the last dimension; however, the new plugin raises an exception in these cases.

The logic for this is that the new behavior matches pillows behavior (which is more intuitive). It is also more intuitive from a "round-trip" perspective in that the following will raise a - perhaps surprising - exception:

```python
import imageio as iio
import numpy as np

foo = np.random.default_rng().integers(0, 256, (50, 50, 1), dtype=np.uint8)
img_bytes = iio.v2.imwrite("<bytes>", foo, format=".png")
bar = iio.imread(img_bytes)
assert bar.shape == foo.shape
# AssertionError
```


FirefoxMetzger:
This PR is RTR :) I will leave it open until Friday for anyone to comment and if there is no open discussion, I will merge.

Issue #849, opened 2022-07-26
tifffile plugin writes unexpected multi-page files from 3d arrays


slanzmich:
Writing a TIFF image from an array of shape (256, 256, 3) writes a multipage TIFF with 256 pages. (This uses the legacy `tifffile` plugin.)

To write an RGB image of shape (H, W, 3), one needs to call `imageio.v3.imwrite` with an image of shape (1, H, W, 3), which I would expect to confuse some users. Writing an image of shape (3, H, W) creates a TIFF with three pages, which is again different.

One could handle some special cases separately, similar to what is done in in the `pillow` plugin:

```python
if ndimage.ndim == 3 and ndimage.shape[-1] in [1, 3, 4]: ...
```

I tried writing a few images of the highly ambiguous shape (20, 20, 20, N). For `N not in [1, 3, 4]`, opening the file in imageJ creates a stack of 400 images. A trailing dimension of size 1, 3, or 4 produces a stack of 20 images. Having the same behavior for a 3d array (i.e. write a single image for these three cases) would be consistent.

FirefoxMetzger:
> Writing a TIFF image from an array of shape (256, 256, 3) writes a multipage TIFF with 256 pages.

This sounds like a bug, because it should become a single RGB page. I will see if I can write a unit-test for this that reproduces the problem and then I will have a look at why this happens :)

> Writing an image of shape (3, H, W) creates a TIFF with three pages, which is again different.

This is intended. Vanilla TIFF can't store channel-first images on a single page. As such we have to either split the image into multiple pages or swap the axis to make the image channel last. 

We chose the road of splitting into pages and, if memory serves, we did so for two reasons: (1) It is shape-preserving for the round-drip of `iio.imread(iio.imwrite("<bytes>", data), index=...)` and (2) because TIFF usually occurs in the context of bioinformatics and here a shape of `(3, H, W)` often doesn't mean that the first axis represents the RGB colorspace. Instead, these are often different colors of staining pictures and they may occur in numbers other than 3 and may be colors other than RGB, so it's better to be conservative.

> I tried writing a few images of the highly ambiguous shape (20, 20, 20, N). For N not in [1, 3, 4], opening the file in imageJ creates a stack of 400 images.

[...]

PR #851, opened 2022-07-26
BUG: write single TIFF page for single RGB image

FirefoxMetzger:
Closes: #849 , https://github.com/imageio/imageio/issues/852

This PR refactors the way we deduce if an ndimage is a single image or a batch of images when writing with legacy plugins. Previously we used the image mode for this; however, this sometimes yields wrong results. For example, a `(256, 256, 3)` image would be saved as 256 `(256, 3)` pages instead of one `(256, 256, 3)` page. 

This PR also addresses an unclear error message that could be seen when passing non-numeric images to ImageIO. Instead of stating the wrong dtype it would use the name of the object instead. This PR addresses this problem as well.

This PR fixes this and also adds a regression test.

### DIFF
--- a/imageio/core/legacy_plugin_wrapper.py
+++ b/imageio/core/legacy_plugin_wrapper.py
@@ -162,7 +162,7 @@ class LegacyPlugin(PluginV3):
         self._request._kwargs = kwargs
         return self._format.get_writer(self._request)
 
-    def write(self, ndimage, **kwargs):
+    def write(self, ndimage, is_batch=None, **kwargs):
         """
         Write an ndimage to the URI specified in path.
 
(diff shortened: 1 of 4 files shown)


FirefoxMetzger:
Another PR is ready to review :) I will leave this open until Friday before merging so that anyone who wishes may comment. Again, if there are no open points I will merge on Friday so that it becomes part of the next release. (This is the policy I adopted until @almarklein comes back from his break.)

@anntzer If you want, you can try this branch and see if you can now write boolean TIFFs :)

anntzer:
Indeed, that makes it work.  Thanks :)

Issue #841, opened 2022-07-21
Restructure PyAV plugin to allow for easier procedural generation of video


FirefoxMetzger:
In discussions with @hmaarrfk  on a recent PR (https://github.com/imageio/imageio/pull/840#issuecomment-1190695106) it emerged that there seem to be two use-cases for writing video:

1. Writing frames to a video in bulk
2. Writing to a video frame-by-frame as frames become available

At the moment the plugin offers a fairly nice API for the first scenario (`iio.imwrite(file, frames, plugin="pyav", codec=codec`). The second approach is possible as well, but the API isn't particularly nice, as `plugin.write` requires a lot of arguments each time. To make this a bit easier, I would like to start discussions on how to add to the current interface and make it more user-friendly. 

When writing frame-by-frame (use-case 2), one usually opens the video via `iio.imopen`, so my idea would be to extend the PyAV plugin with a set of new functions and some getters/setter. This would leave the high-level API unchanged, but add some nicer UX when using the low-level plugin API:

```python
import imageio.v3 as iio
import numpy as np
import io


def frame_yielder():
    rng = np.random.default_rng()

    while True:
        yield rng.integers(0, 255, size=(256, 256, 3), dtype=np.uint8, endpoint=True)


stream = io.BytesIO()
with iio.imopen(stream, "w", plugin="pyav") as file:
    file.init_video_stream(codec="h264", fps=24, pixel_format="yuv420p")
    file.container_metadata = {"key":"some_val"}
    file.stream_metadata = {"rotation":"90"}
    
    # we could use this opportunity to introduce persistent filters
    file.filter_sequence = [("scale", {"size": "vga", "flags": "lanczos"})]
    file.filter_graph = (
            {
                "step": ("framestep", "5"),
                "scale": ("scale", {"size": "vga", "flags": "lanczos"}),
            },
            [
                ("video_in", "step", 0, 0),
                ("step", "scale", 0, 0),
                ("scale", "video_out", 0, 0),
            ],
        )

    for idx, frame in enumerate(frame_yielder()):
        file.write_frame(frame, metadata=None)

        if idx % 100 == 0:
            break
```

Ofc, this would be fully backward compatible in that `PyAVPlugin.write` and `iio.imwrite(..., plugin="pyav")` would keep doing the same thing.

Thoughts and feedback welcome :)

PR #853, opened 2022-07-27
FEAT: pyav API for procedural video writing 

FirefoxMetzger:
Closes: #841 

This PR adds a set of functions to the pyav plugin that make it easier to generate video procedurally, i.e., to write a video frame-by-frame instead of writing frames in bulk. For simple scenarios, this was already possible before through repeated calls to `PyAVPlugin.write`, but this involved setting a lot of kwargs for each call. It also didn't support using filters across multiple call to `write` and there was no way to set metadata. This PR makes cleans most of that up.

The functions extend the basic imageIO API in a plugin-specific way and are available after opening a ImageResource via `iio.imopen`. (Thanks to our type-hinting setup they will even appear as suggestions inside the context manager 😍 🚀 )

**Examples**

```python
import imageio.v3 as iio

# before this PR
# Note 1: this is roughly the extent of what was possible before
# Note 2: this is still possible (the PR is backwards compatible)
with iio.imopen("test.mp4", "w", plugin="pyav") as file:
    for _ in range(5):
        for frame in iio.imiter("imageio:newtonscradle.gif"):
            file.write(frame, is_batch=False, codec="libx264")

# after this PR
with iio.imopen("test.mp4", "w", plugin="pyav") as file:
    file.init_video_stream("libx264")
    for _ in range(5):
        for frame in iio.imiter("imageio:newtonscradle.gif"):
            file.write_frame(frame)
```

New things that this PR adds

```python
import imageio.v3 as iio

# filter output frames before writing them
with iio.imopen("test.mp4", "w", plugin="pyav") as file:
    file.init_video_stream("libx264")
    file.set_video_filter([("scale", "vga")])

    for _ in range(5):
        for frame in iio.imiter("imageio:newtonscradle.gif"):
            file.write_frame(frame)
```

```python
import imageio.v3 as iio

# set stream-level and container-level metadata
with iio.imopen("test.mp4", "w", plugin="pyav") as file:
    file.init_video_stream("libx264")
    file.container_metadata["comment"] = "This video has a rotation flag."
    file.video_stream_metadata["rotate"] = "90"

    for _ in range(5):
        for frame in iio.imiter("imageio:newtonscradle.gif"):
            file.write_frame(frame)

meta = iio.immeta("test.mp4", plugin="pyav")
assert meta["comment"] == "This video has a rotation flag."
```

### DIFF
--- a/tests/test_pyav.py
+++ b/tests/test_pyav.py
@@ -438,6 +438,21 @@ def test_procedual_writing_with_filter(test_images):
     assert actual.shape == (280, 480, 640, 3)
 
 
+def test_rotation_flag_metadata(tmp_path):
+    with iio.imopen(tmp_path / "test.mp4", "w", plugin="pyav") as file:
+        file.init_video_stream("libx264")
+        file.container_metadata["comment"] = "This video has a rotation flag."
+        file.video_stream_metadata["rotate"] = "90"
+
+        for _ in range(5):
+            for frame in iio.imiter("imageio:newtonscradle.gif"):
+                file.write_frame(frame)
+
+    meta = iio.immeta(tmp_path / "test.mp4", plugin="pyav")
+    assert meta["comment"] == "This video has a rotation flag."
+    assert meta["rotate"] == "90"
+
+
 def test_read_filter(test_images):
     image = iio.imread(
         test_images / "cockatoo.mp4",
--- a/tests/test_pyav.py
+++ b/tests/test_pyav.py
@@ -66,6 +66,7 @@ def test_metadata(test_images: Path):
         assert meta["codec"] == "h264"
         assert meta["encoder"] == "Lavf56.4.101"
         assert meta["duration"] == 14
+        assert meta["fps"] == 20.0
 
         meta = plugin.metadata(index=4)
         assert meta["time"] == 0.2
--- a/tests/test_pyav.py
+++ b/tests/test_pyav.py
@@ -179,7 +179,7 @@ def test_filter_graph(test_images):
     assert frames.shape == (56, 480, 640, 3)
 
 
-def test_write_bytes(test_images, tmp_path):
+def test_write_bytes(test_images):
     img = iio.imread(
         test_images / "cockatoo.mp4",
         plugin="pyav",
@@ -193,7 +193,6 @@ def test_write_bytes(test_images, tmp_path):
         img,
         extension=".mp4",
         plugin="pyav",
-        in_pixel_format="yuv444p",
         codec="libx264",
     )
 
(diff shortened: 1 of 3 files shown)


FirefoxMetzger:
@hmaarrfk When you have time, could you take a look at this PR and tell me what you think?

tests/test_pyav.py
```diff
@@ -390,3 +398,47 @@ def test_uri_reading(test_images):
...
+        file.init_video_stream("h264", fps=30)
+        file.set_video_filter(
+            filter_sequence=[
+                ("scale", {"size": "vga", "flags": "lanczos"}),
+            ]
+        )
+        for frame in iio.imiter(
+            test_images / "cockatoo.mp4", plugin="pyav", format="yuv444p"
+        ):
+            file.write_frame(frame, pixel_format="yuv444p")
```
hmaarrfk, line 428: Are you able to change the pixel format on the fly like this?

FirefoxMetzger: Yes, though this refers to the incoming pixel format. Maybe we should keep the name `in_pixel_format`? 

The underlying stream for writing has to have the same pixel format throughout and is set via `file.init_video_stream(codec, pixel_format=...)`. All incoming frames are converted to that format before they are encoded.

hmaarrfk: got it

FirefoxMetzger: If you want to write a file with multiple pixel formats you could create multiple streams and write to each in the respective format. If you set the PTS appropriately you could even write a viewer that can then read that data back. However, this is _very_ non-standard because I don't know of any viewer that recognizes more than one stream.

Also, this is currently not possible with ImageIO, since it is currently designed for exactly one video stream.

FirefoxMetzger:
@almarklein We designed the V3 plugin API so that they can declare custom methods if it makes sense for them, but since this is the first occurrence of us using this feature I think it would be good to also have your feedback (on a high level).

Do you have any objections to adding functionality like this? It won't affect the high-level API (imread/imwrite/...) and will only be available inside the `iio.imopen` context manager for "advanced usage".

FirefoxMetzger:
@hmaarrfk Did you get a chance to test this proposal? If so, what are your thoughts?

hmaarrfk:
I have not had a chance to test this yet. I've been on  the road for a bit.

I can likely give a go this weekend.

FirefoxMetzger:
@hmaarrfk Am I correct in guessing that you have been busy with other work and still didn't have a chance to take a closer look at this PR?

hmaarrfk:
This is very correct. Honestly, I think this requires "thought" so while I've been able to do other things, I feel like they take a different part of the brain.

I think the weekend I had "free" i got sick ...

FirefoxMetzger:
I've added documentation to this PR.

I would like to not stall this for too long so, since there don't seem to be any objections to the proposal, I will merge once all the tests are happy :)

hmaarrfk:
Thank you for pushing forward on this!

Issue #854, opened 2022-07-27
imageio v3 imwrite writes 3D tiff files that don't roundtrip with imread


anntzer:
A 3D tiff file written with the tifffile module is correctly loaded as a 3D array by both tifffile and imageio3.  If the file is instead written by imageio3, only the first page gets loaded (one would need to use imiter instead to load it).
```
In [7]: import tifffile, imageio.v3 as iio, numpy as np
   ...: tifffile.imwrite("/tmp/a.tiff", np.ones((10, 64, 64), "u4"))
   ...: print(tifffile.imread("/tmp/a.tiff").shape)
   ...: print(iio.imread("/tmp/a.tiff").shape)
   ...: iio.imwrite("/tmp/b.tiff", np.ones((10, 64, 64), "u4"))
   ...: print(tifffile.imread("/tmp/b.tiff").shape)
   ...: print(iio.imread("/tmp/b.tiff").shape)
(10, 64, 64)
(10, 64, 64)
(64, 64)
(64, 64)
```

FWIW, looking at the tiff metadata, I note that "a.tiff" (tifffile) contains `ImageDescription: {"shape": [10, 64, 64]}` on the first plane only, wheread "b.tiff" (imageio3) contains `ImageDescription: {"shape": [64, 64]}` on each individual plane.

The problem is present even with #851.  On imageio2, the problem was not present -- one had to explicitly call volwrite or mimwrite depending on the desired semantics, and the reading correctly roundtripped (volwrite -> volread, mimwrite -> mimread).

FirefoxMetzger:
Thanks for the report @anntzer . I can reproduce this on master.

It is related to the way batches of images are auto-detected in the legacy plugin wrapper. When I do the following it seems to work as it should:

```python
iio.imwrite("b.tiff", np.ones((1, 10, 64, 64), "u4")
print(iio.imread("b.tiff").shape)
# (10, 64, 64)
print(tifffile.imread("b.tiff").shape)
# (10, 64, 64)
```

or on #851 

```python
iio.imwrite("b.tiff", np.ones((10, 64, 64), "u4"), is_batch=False)
print(iio.imread("b.tiff").shape)
# (10, 64, 64)
print(tifffile.imread("b.tiff").shape)
# (10, 64, 64)
```

I will try to have a look if it is an easy change or something more complex later today. If it is more complex it will have to wait until next week. My PhD thesis will go to print early next week so today and tomorrow are the final days for me to do any touch-ups on that document, which takes priority.

[...]

PR #857, opened 2022-08-06
Better legacy batch detection

FirefoxMetzger:
Closes: https://github.com/imageio/imageio/issues/854

This PR performs another refactor of `LegacyPlugin.write`. It adds a sophisticated way of auto-detecting if a ndimage is a batch by identifying and counting non-batch dimensions. In particular, it adds a `volume_support` flag to `config.FileExtension` to allow indicating if a format supports volumetric data. Using this flag, we can more clearly decide if a ndimage that should be written to a given format is a volume or a batch of images.

### DIFF
--- a/imageio/config/extensions.py
+++ b/imageio/config/extensions.py
@@ -35,6 +35,8 @@ class FileExtension:
     external_link : str
         A link to further information about the format. Typically, the format's
         specification.
+    volume_support : str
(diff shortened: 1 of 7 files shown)


FirefoxMetzger:
Same as before. I will leave this PR open for comments until Friday and then merge unless there is points that are still up for discussion :)

PR #871, opened 2022-08-30
Raise error if fps is specified

### LINKED ISSUE #870: Passing `duration` or `fps` has no effect on the output GIF using PillowPlugin
```
import imageio.v3 as iio
import matplotlib.pyplot as plt

gif_path = "test.gif"
frames_path = "{i}.jpg"

n = 20
plt.figure(figsize=(4,4))
for x in range(n):
    plt.scatter(x/n, x/n)
    plt.xlim(0, 1)
    plt.ylim(0, 1)
    plt.savefig(f"{x}.jpg")

frames = np.stack(
    [iio.imread(f"{i}.jpg") for i in range(n)],
    axis=0
)

iio.imwrite(gif_path, frames, fps=1, loop=0)
```
I've tried `duration=1000` too

ahuang11:
Closes #870 

Since imageio v2 supported `fps`, imageio v3 should inform the user that it's no longer supported rather than quietly ignoring.

### DIFF
--- a/imageio/plugins/pillow.py
+++ b/imageio/plugins/pillow.py
@@ -280,6 +280,11 @@ class PillowPlugin(PluginV3):
         image.
 
         """
+        if "fps" in kwargs:
+            raise NotImplementedError(
+                f"The keyword, `fps`, is no longer supported; specify `duration`"
+                f"in milliseconds instead, e.g. `fps=60` == `duration=1/60/1000`"
+            )
 
         extension = self.request.extension or self.request.format_hint
 
--- a/imageio/plugins/pillow.py
+++ b/imageio/plugins/pillow.py
@@ -281,7 +281,7 @@ class PillowPlugin(PluginV3):
 
         """
         if "fps" in kwargs:
-            raise NotImplementedError(
+            raise TypeError(
                 "The keyword `fps` is no longer supported. Use `duration`"
                 "(in ms) instead, e.g. `fps=60` == `duration=1/60/1000`."
             )
--- a/imageio/plugins/pillow.py
+++ b/imageio/plugins/pillow.py
@@ -282,8 +282,8 @@ class PillowPlugin(PluginV3):
         """
         if "fps" in kwargs:
             raise NotImplementedError(
-                "The keyword, `fps`, is no longer supported; specify `duration`"
-                "in milliseconds instead, e.g. `fps=60` == `duration=1/60/1000`"
+                "The keyword `fps` is no longer supported. Use `duration`"
+                "(in ms) instead, e.g. `fps=60` == `duration=1/60/1000`."
             )
 
         extension = self.request.extension or self.request.format_hint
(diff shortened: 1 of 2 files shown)


imageio/plugins/pillow.py
```diff
@@ -280,6 +280,11 @@ def write(
         image.
 
         """
+        if "fps" in kwargs:
+            raise NotImplementedError(
+                "The keyword, `fps`, is no longer supported; specify `duration`"
+                "in milliseconds instead, e.g. `fps=60` == `duration=1/60/1000`"
```
FirefoxMetzger, line 286:
```suggestion
                "The keyword `fps` is no longer supported. Use `duration`"
                "(in ms) instead, e.g. `fps=60` == `duration=1/60/1000`."
```

imageio/plugins/pillow.py
```diff
@@ -280,6 +280,11 @@ def write(
         image.
 
         """
+        if "fps" in kwargs:
+            raise NotImplementedError(
```
FirefoxMetzger, line 284:
```suggestion
            raise TypeError(
```

Python raises a TypeError for unknown kwargs, and I think we should do the same.

tests/test_pillow.py
```diff
@@ -255,6 +255,23 @@ def test_gif_gray(test_images, tmp_path):
...
 
 
+def test_gif_fps_error(test_images, tmp_path):
+    im = iio.v3.imread(
+        test_images / "newtonscradle.gif",
+        plugin="pillow",
+        mode="L",
+    )
+
+    with pytest.raises(NotImplementedError, match="The keyword, `fps`"):
```
FirefoxMetzger, line 265:
```suggestion
    with pytest.raises(TypeError, match="The keyword, `fps`"):
```

FirefoxMetzger (approved):
@ahuang11 Thanks for the PR :) Sorry for taking a while to review; I was attending a conference and didn't get a chance to sit down and look through GH.

I changed the error type to match what python does by default. Beyond that, the PR looks good to go.

tests/test_pillow.py
```diff
@@ -255,6 +255,23 @@ def test_gif_gray(test_images, tmp_path):
...
 
 
+def test_gif_fps_error(test_images, tmp_path):
+    im = iio.v3.imread(
+        test_images / "newtonscradle.gif",
+        plugin="pillow",
+        mode="L",
+    )
+
+    with pytest.raises(TypeError, match="The keyword, `fps`"):
```
FirefoxMetzger, line 265:
```suggestion
    with pytest.raises(TypeError):
```

Sorry, I forgot to make a suggestion here. We typically don't use the `match=` part of `pytest.raises`. It has its uses, but for us it is enough if the test "fails" with the correct exception.

FirefoxMetzger:
Thanks @ahuang11 for the PR :) This should reduce the confusion around when and how to use `fps` or `duration` 🎉 🚀

Issue #856, opened 2022-08-02
pyav testing (wrongly) requires ffmpeg plugin to be installed


anjos:
This seems like a bug introduced on a couple of releases back.  The test function `test_sequential_reading()` from ` tests/test_pyav.py` is failing on conda-forge builds (https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=544225&view=logs&jobId=656edd35-690f-5c53-9ba3-09c10d0bea97&j=656edd35-690f-5c53-9ba3-09c10d0bea97&t=e5c8ab1d-8ff9-5cae-b332-e15ae582ed2d).   The reason, I think, could be as simple as a missing `plugin="pyav",` on calls to `iio.imread()`?

FirefoxMetzger:
Interesting find @anjos . In theory ImageIO should, if no plugin is explicitly given, step over missing plugins and try the next. As such, if imageio-ffmpeg is missing, it should default to pyav.

You are correct that `plugin="pyav"` will likely fix this; however, this might be a symptom of a different underlying cause. I will have a look during the week :)

anjos:
OK, so we have package, I'll patch this like so for the time being on conda-forge.

PR #880, opened 2022-09-19
BUG: raise ImportError when iio-ffmpeg is missing

FirefoxMetzger:
Closes: https://github.com/imageio/imageio/issues/856

This PR slightly refactors the imageio-ffmpeg plugin to raise a `ImportError` when its backend is not installed. This makes our codebase cleaner while also fixing a bug in our test suite that required imageio-ffmpeg to be installed when testing our pyav plugin.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -23,7 +23,6 @@ from pathlib import Path
 import imageio.plugins.bsdf
 import imageio.plugins.dicom
 import imageio.plugins.feisem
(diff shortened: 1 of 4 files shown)


almarklein (approved):
For the record (I had to think twice), this does not mean imageio-ffmpeg is now a required dependency. For some time, the plugin modules are only imported when actually being used, which makes the lazy import in the plugin modules themselves unnecessary.

PR #799, opened 2022-04-25
add audio support for the writer

grovergol:
just forward the audio+path and audio_codec attributes to the write_frames method, so it's possible to include audio when using ffmpeg writer

Issue #790, opened 2022-04-18
Allow `bytes` ImageResource


FirefoxMetzger:
Currently we allow passing `"<bytes>"` as uri/image_resource when writing in order to write to a byte string. I was thinking that we could extend that to allow the (built-in) class `bytes` and make it more intuitive to write bytes.

I.e., I would like to do the following change:

```python
# before:
encoded_img = iio.imwrite("<bytes>", ndimage)

# after:
encoded_img = iio.imwrite(bytes, ndimage)
```

What do you guys think?

almarklein:
mmm, interesting. I think I quite like this :)

PR #882, opened 2022-09-23
Add audio support (finished)

paulinevos:
Based on #799 

Finishes the work in the PR referenced above, as it lacked documentation and a test and the original author had said not to be able to pick it up again.

Opening on a new branch as I have no write access to the original fork. Please lmk if I need to close the PR and request write access to the other fork instead.

### DIFF
--- a/tests/test_ffmpeg.py
+++ b/tests/test_ffmpeg.py
@@ -222,6 +222,43 @@ def test_write_not_contiguous(test_images):
             assert diff.mean() < 2.5
 
 
+def write_audio(test_images, tmp_path, codec=None) -> dict:
+    in_filename = test_images / "realshort.mp4"
+    out_filename = tmp_path / "realshort_audio.mp4"
+
+    in_file = []
+    with iio.read(in_filename, "ffmpeg") as R:
+        for i in range(5):
+            im = R.get_next_data()
+            in_file.append(im)
+
+    # Now write with audio to preserve the audio track
+    with iio.save(
+        out_filename,
+        format="ffmpeg",
+        audio_path=in_filename.as_posix(),
+        audio_codec=codec,
+    ) as W:
+        for im in in_file:
+            W.append_data(im)
+
+    R = iio.read(out_filename, "ffmpeg", loop=True)
+    meta = R.get_meta_data()
+    R.close()
+
+    return meta
+
+
+def test_write_audio_ac3(test_images, tmp_path):
+    meta = write_audio(test_images, tmp_path, "ac3")
+    assert "audio_codec" in meta and meta["audio_codec"] == "ac3"
+
+
+def test_write_audio_default_codec(test_images, tmp_path):
+    meta = write_audio(test_images, tmp_path)
+    assert "audio_codec" in meta
+
+
 def test_reader_more(test_images):
 
     fname1 = test_images / "cockatoo.mp4"
--- a/imageio/plugins/ffmpeg.py
+++ b/imageio/plugins/ffmpeg.py
@@ -106,7 +106,14 @@ macro_block_size: int
     automatic feature set it to None or 1, however be warned many players
     can't decode videos that are odd in size and some codecs will produce
     poor results or fail. See https://en.wikipedia.org/wiki/Macroblock.
-
+audio_path : str | None
+    Audio path of any audio that needs to be written. Defaults to nothing,
+    so no audio will be written. Please note, when writing shorter video
+    than the original, ffmpeg will not truncate the audio track; it
+    will maintain its original length and be longer than the video.
+audio_codec : str | None
+    The audio codec to use. Defaults to nothing, but if an audio_path has
+    been provided ffmpeg will attempt to set a default codec.
 
 Notes
 -----
@@ -542,6 +549,8 @@ class FfmpegFormat(Format):
             ffmpeg_log_level="quiet",
             quality=5,
             macro_block_size=16,
+            audio_path=None,
+            audio_codec=None,
         ):
             self._ffmpeg_api = imageio_ffmpeg
             self._filename = self.request.get_local_filename()
@@ -617,6 +626,8 @@ class FfmpegFormat(Format):
             pixelformat = self.request.kwargs.get("pixelformat", None)
             macro_block_size = self.request.kwargs.get("macro_block_size", 16)
             ffmpeg_log_level = self.request.kwargs.get("ffmpeg_log_level", None)
+            audio_path = self.request.kwargs.get("audio_path", None)
+            audio_codec = self.request.kwargs.get("audio_codec", None)
 
             macro_block_size = macro_block_size or 1  # None -> 1
 
@@ -634,6 +645,8 @@ class FfmpegFormat(Format):
                 ffmpeg_log_level=ffmpeg_log_level,
                 input_params=input_params,
                 output_params=output_params,
+                audio_path=audio_path,
+                audio_codec=audio_codec,
             )
 
             # Seed the generator (this is where the ffmpeg subprocess starts)


FirefoxMetzger:
Thanks @paulinevos I am busy this weekend, but I will have a look at the beginning of the coming week.

paulinevos:
Same @FirefoxMetzger, will take a look at the CI issues on Monday.

tests/test_ffmpeg.py
```diff
@@ -224,6 +226,44 @@ def test_write_not_contiguous(test_images):
...
+
+    # Now write with audio to preserve the audio track
+    with iio.save(
+        out_filename_audio, format="ffmpeg", audio_path=in_filename.as_posix()
+    ) as W:
+        for im in in_file:
+            W.append_data(im)
+
+    # Compare audio streams
+    assert has_audio_streams(in_filename)
```
FirefoxMetzger, line 262: Considering that this is a constant and external to the test, it seems a bit redundant to make assertions about the input file. Can we remove this line?

tests/test_ffmpeg.py
```diff
@@ -224,6 +226,44 @@ def test_write_not_contiguous(test_images):
...
+
+    # Write without audio
+    with iio.save(out_filename, "ffmpeg") as W:
+        for im in in_file:
+            W.append_data(im)
+
+    # Now write with audio to preserve the audio track
+    with iio.save(
+        out_filename_audio, format="ffmpeg", audio_path=in_filename.as_posix()
+    ) as W:
```
FirefoxMetzger, line 257: I must admit that I am not an expert on imageio-ffmpeg internals, so I may be misunderstanding things.

If I call this code as is, will it copy (transmux) the full audio from the input file into the output file, or will it truncate it to be in sync with the video frames, i.e., since we are only writing a fraction of a second of video, will we only copy a fraction of the audio or will we get a full copy of the audio track?

FirefoxMetzger (commented):
Thanks for the PR :) I left a few comments; nothing major except for the `ffprobe` issue that is failing CI.

I also added two commits to your branch to fix the linting and to reorganize where files are stored during testing. The latter is like this because we cache the `test_images` locally and thus treat it as read-only. It is also nice to keep tests from accidentally interfering with each other, because each test gets its own, clean tempdir.

paulinevos:
@FirefoxMetzger thanks for your comments, I've read them and will respond ASAP. Haven't been able to yet as I'm traveling but will soon.

tests/test_ffmpeg.py
paulinevos: Sure, I like to add things like this to ensure that if for some reason whatever it's referencing changes, we'll know that the test is still valid. But I can see your point.

tests/test_ffmpeg.py
```diff
@@ -224,6 +226,44 @@ def test_write_not_contiguous(test_images):
             assert diff.mean() < 2.5
 
 
+def has_audio_streams(file_path):
+    command = ["ffprobe", "-show_streams", "-print_format", "json", file_path]
```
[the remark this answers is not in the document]
paulinevos: Ah yes, of course. I hadn't considered the dependency issue, this was just the most straightforward solution I could find on ye olde internette. 
I'll have to look into the ffmpeg plugin (I've literally never touched this library before but wanted to help with the issue) and figure out how to do that.

tests/test_ffmpeg.py
paulinevos: That's a very good question. Will actually try it out and back to you on that.

tests/test_ffmpeg.py
```diff
@@ -224,6 +226,44 @@ def test_write_not_contiguous(test_images):
...
+    in_file = []
+    with iio.read(in_filename, "ffmpeg") as R:
+        for i in range(10):
+            im = R.get_next_data()
+            in_file.append(im)
+
+    # Write without audio
+    with iio.save(out_filename, "ffmpeg") as W:
+        for im in in_file:
+            W.append_data(im)
```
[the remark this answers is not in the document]
paulinevos: Mostly so we know it doesn't break when you don't provide one, although there are older `write` test cases that don't provide any audio so I guess this is redundant.

yukw777:
This would be an amazing addition to the library!

paulinevos:
@FirefoxMetzger fully intend to still finish it! Been a bit swamped but I was thinking about it today, will have a stab at it tomorrow

paulinevos:
Just to let you know I am working on this here and there, but it's challenging to me cause I'm not super familiar with the Python ecosystem. Bear with me or feel free to oust me from this PR, lol. If anyone has 5 mins to troubleshoot over a call that would be cool too.

vincenzoml:
Looking forward to muxing audio in imageio! Was about to open an issue.

paulinevos:
@vincenzoml Thanks for the nudge ;)

This took me a liiiiittle longer than anticipated due to holidays and taking a while to figure out the Python ecosystem, sorry folks.

tests/test_ffmpeg.py
paulinevos: Ah shoot, I just pushed with the adapted test and realized I hadn't check this. And yes, we'll get the full audio track while the video stops at the last frame. Will try to think of a way to circumvent this.

vincenzoml:
@paulinevos: thank you! Will this patch also enable writing to file-like objects in memory (this can be done e.g. using mp4 and no audio currently)

tests/test_ffmpeg.py
FirefoxMetzger:
```suggestion
```

Then it's a question of how much we trust the test data. It is certainly something we could test, but if we choose to do so, we should keep data tests separate from code tests. As such, I would like to remove it here.

Regarding data tests in general, we currently don't have them. We own the test data and it is very slow changing (I think the last actual addition/deletion was about 2 years ago) so not having tests for it has worked out thus far. That said, they certainly have their place, especially when you are testing with "live data" or external sources so I do like your thinking.

tests/test_ffmpeg.py
FirefoxMetzger: I agree that `imageio-ffmpeg` would be the place where this has to be implemented. The reason for this is that we will likely have to truncate the last audio package to match the exact time at which the last frame is sent.

That said, not having this feature is no dealbreaker as long as we clearly document this limitation. Could you write a sentence or two about this in the `Notes` section of the docs? Something along the lines of: "This will copy the full audio stream, so make sure your video length matches that."

tests/test_ffmpeg.py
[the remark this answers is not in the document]
FirefoxMetzger:
```suggestion
```

Yes, let's only test the scenario of writing audio here. I'm happy without an explicit test that checks if not specifying audio produces no-audio files. However, if you feel like this should be tested, too, let's do it in a second unit test.

FirefoxMetzger:
@paulinevos Glad to see you are picking this up again 🚀 I think we are pretty much LGTM, except for some final cleanup and a note in the docs about the inability to truncate audio.

> Will this patch also enable writing to file-like objects in memory (this can be done e.g. using mp4 and no audio currently)

@vincenzoml Yes. ImageIOs ability to write to file-like objects is independent of the plugin's ability to write audio, so you should be able to write audio to file-like objects using the usual syntax.

tests/test_ffmpeg.py
paulinevos: Will do

FirefoxMetzger (approved):
Awesome work @paulinevos ! This is LGTM once the tests are happy. We use black as our default code formatting tool. Have you used it before?

paulinevos:
@FirefoxMetzger I hadn't but I just ran it. For some reason sphinx won't run for me after installation. Do I need it? The docs check error also doesn't look like it should be relevant to my changes 🤔 

```
WARNING: unsupported theme option 'google_analytics_id' given
```

paulinevos:
I've added a test case that asserts a provided audio codec is present, hopefully that will fix coverage (not sure cause the coverage report I get locally isn't consistent with the one in CI)

vincenzoml:
> @paulinevos Glad to see you are picking this up again 🚀 I think we are pretty much LGTM, except for some final cleanup and a note in the docs about the inability to truncate audio.
> 
> > Will this patch also enable writing to file-like objects in memory (this can be done e.g. using mp4 and no audio currently)
> 
> @vincenzoml Yes. ImageIOs ability to write to file-like objects is independent of the plugin's ability to write audio, so you should be able to write audio to file-like objects using the usual syntax.
> 
> Note, however, that writing to `"<bytes>"` using `imageio-ffmpeg` isn't 100% in memory in that the process involves an intermediate file which is used as a buffer. If you truly need in-memory encoding, i.e., without touching the filesystem, you will have to use our `pyav` plugin. (It doesn't support audio yet, but it is a fairly straightforward to add if you want to give it a shot.)

First of all I'd love to test the functionality that this patch adds; how can I do that? Then indeed, I'd also be very happy to have truly in-memory av-muxing via pyav, and am available to test it if needed, but also writing to an intermediate file may work.

FirefoxMetzger:
@paulinevos Yes, coverage is done and it looks good as is.

I noticed you force-pushed a few times, I assume this is you rebasing on master? If so, I shall have a look into why RTD fails.

paulinevos:
Thanks! No, it's not. I can rebase onto master to see if this resolves it.

I amend commits/edit them in interactive rebase in order to [keep my commits atomic](https://www.pauline-vos.nl/atomic-commits/) so that we'll wind up with one commit that accurately summarizes the introduced change as opposed to lots of separate commits that are only relevant in the scope of the PR. That's why I force push.

FirefoxMetzger:
> I amend commits/edit them in interactive rebase in order to [keep my commits atomic](https://www.pauline-vos.nl/atomic-commits/)

Oh, I will have to read your blog post later, but we (and most other big libraries in this area like numpy, scipy, scikit-image, etc.) explicitly don't do this for a simple reason: We [squash & merge](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/about-pull-request-merges#squash-and-merge-your-commits) instead. This way, we can have the best of both worlds: Incremental commits within the PR make it easy to keep up to date, have others collaborate/contribute on your PR, look at what has changed since you last viewed, etc. Then later, once the PR gets merged, it will become one large commit to master which keeps its history clean.

This is, of course, not the only way of managing your repo, but it is the most common among all the OSS I contributed to and (I think) the flow that lends itself best to open source.

paulinevos:
@FirefoxMetzger Totally valid! Well I'll keep that in mind for next time, cause looks like all the checks are green 🎉 :)

FirefoxMetzger:
Awesome 🚀 Then it is time to merge. 

Thanks for the contribution @paulinevos 💯 It was fun working with you. Let me know if you'd like to tackle another issue; I can suggest something or you can take your pick depending on if you want something more technical or less technical.

paulinevos:
@FirefoxMetzger This was especially fun for me being in a language and library I had pretty much zero experience with lol. Could def attempt another issue if you have one. I'm also a mac user (M1) which I read may be useful to you?

FirefoxMetzger:
@paulinevos I think a small new feature could be nice. There is no explicit issue for this yet, but there is something related to image/video metadata that I always wanted to have.

We have a function called [`improps`](https://imageio.readthedocs.io/en/stable/_autosummary/imageio.v3.improps.html#imageio.v3.improps), which returns [standardized metadata](https://imageio.readthedocs.io/en/stable/_autosummary/imageio.core.v3_plugin_api.ImageProperties.html) (same name and format for all supported containers). Currently it only supports a small set of fields, but it would be nice if we could add the fields `plugin` and `extension`. `plugin` would correspond to the name of the plugin that was used to read the image, and `extension` corresponds to the typical extension of an image/video of that type, e.g. `".png"` for PNG images.

This would make it much easier to interact with and identify images, as it will enable things like:

```python
buffer = ...  # some image ... or at least we think it is

assert iio.improps(buffer).extension == ".jpg"
# buffer is a jpeg image.
```

If that sounds interesting, I'll make an issue for it and we can discuss the details of it there.

---

Alternatively, you could improve the syntax of writing to `bytes` (https://github.com/imageio/imageio/issues/790). Currently this works by using the special `"<bytes>"` keyword as a target, e.g.

```python
import imageio.v3 as iio

# convert PNG to JPG in memory
img = iio.imread("imageio:chelsea.png")
buffer = iio.imwrite("<bytes>", img, extension=".jpg")
```

It would be nice if we could directly use the bytes class itself:

```python
import imageio.v3 as iio

# convert PNG to JPG in memory
img = iio.imread("imageio:chelsea.png")
buffer = iio.imwrite(bytes, img, extension=".jpg")
```

---

There is also an issue that specifically requires a Mac and an M1. That is, however, related to building freeimage, which means that you will need some knowledge about building C libraries and packaging them into a python package alongside normal python code.

I don't know your background, so maybe this is right up your alley. If so, you're very welcome to work on it though I don't recommend it as a second project, because dealing with the packaging and compilation of C extensions can quickly become a frustrating experience. 

I did some initial work on this for windows and linux (lives here: https://github.com/imageio/imageio-freeimage/pull/3 ), but I put it aside for now until I can dedicate a few consecutive days to fix all the missing pieces.

vincenzoml:
Sorry for naiveness: I still don't know how to use this. First of all which version of Imageio do I need and which plugins? Then, how to write the file with audio muxing? I have a function generating audio frames (so that I can, e.g., generate a list of pairs of values for each video frame, corresponding to the audio fragment of that frame) and a function generating video frames. One naive way to write the video to bytesIO is

frames = []
for t in range(num_frames):
    frames.append(f(t))

video_data=io.BytesIO()
imageio.mimsave(video_data,frames,"mp4",fps=10)

Or I could use "<bytes>". But how to add audio to it?

Issue #887, opened 2022-10-03
When iterate over supported extenstions of imageio got ModuleNotFoundError


Czaki:
In tests, when iterating over supported extensions of imagio it ends with `ModuleNotFoundError` exception.

As ffmpeg support is optional, if there is a lack of this dependency, it should end with not reporting support for such extension, not exception about the missed format. 

https://github.com/napari/superqt/actions/runs/3172165766/jobs/5166348387
```
 /opt/hostedtoolcache/Python/3.10.7/x64/lib/python3.10/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
napari/_qt/__init__.py:54: in <module>
    from .qt_main_window import Window
napari/_qt/qt_main_window.py:56: in <module>
    from .qt_viewer import QtViewer
napari/_qt/qt_viewer.py:15: in <module>
    from napari_builtins.io import imsave_extensions
napari_builtins/io/__init__.py:1: in <module>
    from ._read import magic_imread, napari_get_reader
napari_builtins/io/_read.py:25: in <module>
    IMAGEIO_EXTENSIONS = {x for f in imageio.formats for x in f.extensions}
napari_builtins/io/_read.py:25: in <setcomp>
    IMAGEIO_EXTENSIONS = {x for f in imageio.formats for x in f.extensions}
/opt/hostedtoolcache/Python/3.10.7/x64/lib/python3.10/site-packages/imageio/core/format.py:640: in <genexpr>
    return iter(x.format for x in self._formats)
/opt/hostedtoolcache/Python/3.10.7/x64/lib/python3.10/site-packages/imageio/config/plugins.py:85: in format
    module = importlib.import_module(self.module_name, self.package_name)
/opt/hostedtoolcache/Python/3.10.7/x64/lib/python3.10/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
/opt/hostedtoolcache/Python/3.10.7/x64/lib/python3.10/site-packages/imageio/plugins/ffmpeg.py:136: in <module>
    import imageio_ffmpeg
E   ModuleNotFoundError: No module named 'imageio_ffmpeg
```

FirefoxMetzger:
Interesting find. Could you point me to the actual code that produces this exception?

Czaki:
[...]

PR #888, opened 2022-10-03
BUG: Make legacy FormatManager skip uninstalled plugins

FirefoxMetzger:
Closes: #887 

While merging #880 we introduced a regression in `FormatManager`. This PR resolves the regression and adds a test for this for it to avoid regressing again :)

### DIFF
--- a/imageio/core/format.py
+++ b/imageio/core/format.py
@@ -31,6 +31,7 @@ a format object using ``imageio.formats.add_format()``.
 
 import sys
 import warnings
+import contextlib
 
 import numpy as np
 from pathlib import Path
--- a/imageio/core/format.py
+++ b/imageio/core/format.py
@@ -635,12 +635,9 @@ class FormatManager(object):
         available_formats = list()
 
         for config in known_plugins.values():
-            if config.is_legacy:
-                try:
-                    config.format
-                except ImportError:
-                    pass  # format not installed
-                else:
+            with contextlib.suppress(ImportError):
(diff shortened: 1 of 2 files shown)


imageio/core/format.py
```diff
@@ -631,7 +631,18 @@ class FormatManager(object):
 
     @property
     def _formats(self):
-        return [x for x in known_plugins.values() if x.is_legacy]
+        available_formats = list()
```
Czaki, line 634:
```suggestion
        available_formats = []
```

This is faster and recommended

imageio/core/format.py
```diff
@@ -631,7 +631,18 @@ class FormatManager(object):
...
+
+        for config in known_plugins.values():
+            if config.is_legacy:
+                try:
+                    config.format
+                except ImportError:
+                    pass  # format not installed
+                else:
+                    available_formats.append(config)
```
Czaki, line 643:
```suggestion
            with contextlib.suppress(ImportError):
                # if an exception is raised, then format not installed
                if config.is_legacy and config.format is not None:                    
                    available_formats.append(config)
```

this requires adding `import contextlib` at the beginning of the file, but for me, it is closer.

Czaki (approved):
few minor comments

imageio/core/format.py
FirefoxMetzger: @Czaki Nice. Learned about a handy new function today, thanks for the pointer :)

Issue #893, opened 2022-10-21
Saving TIFF files with v3 API and using the compression keyword results in unexpected argument


lagru:
### Description:

I am struggling to make compression work when saving a TIFF with imageio's v3 API (see the reproducing code below). From the [plugin documenation](https://imageio.readthedocs.io/en/stable/_autosummary/imageio.plugins.tifffile.html#metadata-for-writing) I get the impression that some form of compression is supported (previous version of tifffile used `compress`, now it's called `compression`). I've also tried specifying `plugin='TIFF'` and `plugin='TIFF-PIL'` explicitly with a similar outcome. Same behavior if I uninstall tifffile. Is imageio perhaps not detecting the externally installed tifffile package?

I would be grateful for guidance on how to save compressed TIFF-files with the v3 API.

### Code to reproduce:

```python
import tifffile  # version 2022.10.10
import numpy as np
import imageio.v3 as iio  # version 2.22.2
import imageio.v2 as iio2

image = np.ones((3, 4))

tifffile.imwrite("/tmp/foo.tiff", data=image, compression="zlib")  # works fine
with iio2.get_writer("/tmp/baz.tiff", mode="i") as writer:
    writer.append_data(image, {"compression": "zlib"})  # works fine

iio.imwrite(uri="/tmp/bar.tiff", image=np.ones((3, 3)), compression="zlib")  # raises
# TypeError: TiffFormat.Writer._open() got an unexpected keyword argument 'compression'
with iio.imopen("/tmp/bar.tiff", "w") as file:
    file.write(image, compression="zlib")  # raises 
# TypeError: TiffFormat.Writer._open() got an unexpected keyword argument 'compression'
```

<details>
<summary>imageio.plugins.formats.show()</summary>

`imageio.plugins.formats.show()` prints:
```
TIFF - TIFF format [.tif .tiff .stk .lsm]
BMP-PIL - Windows Bitmap via Pillow [.bmp]
BUFR-PIL - BUFR via Pillow [.bufr]
CUR-PIL - Windows Cursor via Pillow [.cur]
DCX-PIL - Intel DCX via Pillow [.dcx]
DDS-PIL - DirectDraw Surface via Pillow [.dds]
DIB-PIL - Windows Bitmap via Pillow []
EPS-PIL - Encapsulated Postscript via Pillow [.ps .eps]
FITS-PIL - FITS via Pillow [.fit .fits]
[...]

PR #894, opened 2022-10-25
BUG: Support `compression` and `compressionargs` in tifffile plugin

FirefoxMetzger:
Closes: https://github.com/imageio/imageio/issues/893

This PR adds a `metadata` kwarg to the LegacyPlugin, which allows passing metadata to a format's `append_data` function. It also adds support for the (not so) new kwargs `compression` and `compressionargs` which tifffile has introduced to manage compression. Among other things, these changes make it possible to use TIFF compression in the v3 API.

### DIFF
--- a/imageio/plugins/tifffile.py
+++ b/imageio/plugins/tifffile.py
@@ -123,6 +123,17 @@ description : str
 compress : int
     Values from 0 to 9 controlling the level of zlib (deflate) compression.
     If 0, data are written uncompressed (default).
+compression : str, (int, int)
+    Compression scheme used while writing the image. If omitted (default) the
+    image is not uncompressed. Compression cannot be used to write contiguous
+    series. Compressors may require certain data shapes, types or value ranges.
+    For example, JPEG compression requires grayscale or RGB(A), uint8 or 12-bit
+    uint16. JPEG compression is experimental. JPEG markers and TIFF tags may not
+    match. Only a limited set of compression schemes are implemented. 'ZLIB' is
+    short for ADOBE_DEFLATE. The value is written to the Compression tag.
+compressionargs:
+    Extra arguments passed to compression codec, e.g., compression level. Refer
+    to the Imagecodecs implementation for supported arguments.
 predictor : bool
     If True, horizontal differencing is applied before compression.
     Note that using an int literal 1 actually means no prediction scheme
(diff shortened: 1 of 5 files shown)


imageio/plugins/tifffile.py
```diff
@@ -529,6 +531,17 @@ def _sanitize_meta(meta):
                     # 1(=NONE) translation to False expected by TiffWriter.save
                     if key == "predictor" and not isinstance(value, bool):
                         ret[key] = value > 1
+                    elif key == "compress" and value != 0:
+                        warnings.warn(
+                            "The use of `compress` is deprecated. Use `compression` and `compressionargs` instead.",
+                            DeprecationWarning,
+                        )
+
+                        if _tifffile.__version__ < "2022":
```
xkszltl, line 540: Does that mean there's no any version using `compress`?

imageio/plugins/tifffile.py
```diff
@@ -191,6 +191,8 @@
     "resolution",
     "description",
     "compress",
+    "compression",
```
xkszltl, line 194: Is that all? Maybe we can add the entire set of new args they currently have?
- https://github.com/cgohlke/tifffile/blob/a73812e98f58b068a590de862ed547820fd332ad/tifffile/tifffile.py#L1514-L1551

imageio/plugins/tifffile.py
FirefoxMetzger: Yes. `compress` was used at some point to facilitate compression and while I didn't check the version of tifffile where this change happened, it has been deprecated for at least one year.

imageio/plugins/tifffile.py
FirefoxMetzger: We could; although we would also have to also document those new kwargs. I would much rather spend that time on writing a v3 plugin for tifffile. That would also solve the problem of reoccurring maintenance costs for keeping the plugin in sync.

Speaking off documentation, I just noticed that I forgot to add documentation for `compression` and `compressionargs` ... will do :)

xkszltl: IMHO undocumented feature is still better than not working, so if I'm doing it I would probably just copy-paste their arg list here. But anyway, my issue is resolved so will let someone else to worry about the rest.

Issue #726, opened 2022-01-16
Mac unable to read camear


jinfagang:

FirefoxMetzger:
Mac has a tighter security system than the other OS. Did you double-check that you gave permission to the shell executing the python script/interpreter to access your webcam?

FirefoxMetzger:
@jinfagang Did you manage to solve the issue? If not, could you provide some more detail so that we can try to reproduce the behavior locally?

(As with similar issues in the past, if I don't hear from you in the next couple of days, I will close this issue as solved. However, you can simply comment if it persists and I can reopen.)

almarklein:
I can reproduce this. Could you try `iio.get_reader('<video0>', fps=30)`?

almarklein:
This relates to #699. It was me who proposed to set the default framerate to 15, because earlier reports from mac users indicated that that was necessary.

PR #899, opened 2022-11-01
BUG: Avoid webcam hang on MacOS

### LINKED ISSUE #898: Webcam frames are all the same
I just followed the example code to capture frames from my webcam, but the frames I get are all the same (the initial image). Here's a minimal reproducible example:

Running on MacBook Air (M2, 2022) macOS Monterey 12.6, python3.9, imageio 2.22.3

```python
import imageio.v3 as iio
import numpy as np

prev_frame = None
for idx, frame in enumerate(iio.imiter("<video0>")):
    if prev_frame is None:
        prev_frame = frame
        continue
    print(np.array_equal(frame, prev_frame))
    prev_frame = frame
```

You'll notice that it only outputs True despite you moving in front of the camera.

almarklein:
Fixes #898

### DIFF
--- a/imageio/plugins/ffmpeg.py
+++ b/imageio/plugins/ffmpeg.py
@@ -321,20 +321,23 @@ class FfmpegFormat(Format):
             # metadata when we boot ffmpeg ... maybe we could refactor this so we can
             # get the metadata beforehand, but for now we'll just give it 2 tries on MacOS,
             # one with fps 30 and one with fps 15.
-            need_fps = (
-                self.request._video
-                and platform.system().lower() == "darwin"
-                and "-framerate" not in str(self._arg_input_params)
-            )
-            if need_fps:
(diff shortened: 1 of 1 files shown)


FirefoxMetzger:
@almarklein Thanks for the PR :) 

Do you think there is an easy way to add a regression test for this? If not, we can merge it as is; otherwise, I would like to have it because we keep getting issues about the webcam on MacOS breaking (https://github.com/imageio/imageio/issues/726, https://github.com/imageio/imageio/issues/391, https://github.com/imageio/imageio/pull/525).

almarklein:
Could add a test, but CI does not have a camera, so it would only run locally. Not sure if that's worth the effort.

FirefoxMetzger:
@almarklein Yep, that matches my thinking. I will merge this as is and see if we can add proper tests for webcam support at a later date.

PR #903, opened 2022-11-02
DOC: add code of conduct

FirefoxMetzger:
This PR adds an explicit code of conduct to ImageIO. 

While we didn't have any incidents during the time I have been here, it is required to have an explicit code to apply for affiliation with NumFOCUS. I would like to apply and eventually attract and grow a community around ImageIO; hence this PR.

@almarklein @jni Could I trouble both of you to have a look at this draft and tell me what you think? I'm especially hoping for insights from @jni since scikit-image is fiscally sponsored by NumFOCUS and have thus I assume you have gone through this process at some point :)

Of course, feedback from others is also highly appreciated 🚀

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,3 +118,7 @@ We welcome contributions of any kind. Here are some suggestions on how you are a
 - report/fix any bugs you encounter while using ImageIO
 
 To assist you in getting started with contributing code, take a look at the [development section](https://imageio.readthedocs.io/en/stable/development/index.html) of the docs. You will find instructions on setting up the dev environment as well as examples on how to contribute code.
+
+<h2>Code of Conduct</h2>
+
+ImageIO is a community-driven open-source project developed by a diverse group of contributors. We want to create an open, inclusive, and positive community. Please read the Code of Conduct for guidance on how to interact with others in a way that makes this community thrive.
--- a/docs/code_of_conduct.rst
+++ b/docs/code_of_conduct.rst
@@ -1,0 +1,116 @@
+Code of Conduct
+===============
+
+TL:DR
+------
+
+- Be professional, kind, and don't insult others.
+- Strive to create an environment of learning and healthy discussion. 
+- Don't bully people.
+
+Introduction
+------------
+
+This Code of Conduct applies to all spaces managed by the ImageIO project,
+including all public and private mailing lists, issue trackers, wikis, blogs,
+Twitter, and any other communication channel used by our community. 
+
+This Code of Conduct should be honored by everyone who participates in the
+ImageIO community formally or informally, or claims any affiliation with the
+project, in any project-related activities and especially when representing the
+project, in any role.
+
+This code is not exhaustive or complete. It serves to distil our common
+understanding of a collaborative, shared environment and goals. Please try to
+follow this code in spirit as much as in letter, to create a friendly and
+productive environment that enriches the surrounding community.
+
+Specific Guidelines
+-------------------
+
+We strive to:
+
+1.	Be open. We invite anyone to participate in our community. We prefer to use
+  	public methods of communication for project-related messages, unless
+  	discussing something sensitive. This applies to messages for help or
+  	project-related support, too; not only is a public support request much more
+  	likely to result in an answer to a question, but it also ensures that any
+  	inadvertent mistakes in answering are more easily detected and corrected.
+2.	Be empathetic, welcoming, friendly, and patient. We work together to resolve
+  	conflict and assume good intentions. A community where people feel
+  	uncomfortable or threatened is not a productive one.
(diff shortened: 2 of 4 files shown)


almarklein (approved):

stefanv:
The PSF was offering to help arbitrate CoC issues. While CoC issues *can* be a conflict of interest, they are also at the *core* of interest to a healthy project community. I therefore recommend that projects try to resolve matters amicably, internally, and escalate to an external committee only if necessary. External groups necessarily are not intimately familiar with project history / culture (pro & con, I suppose!), but unless they are very careful can easily escalate matters unnecessarily.

Importantly, when you receive a CoC complaint, you need to be sensitive in how you communicate with both the reporter and the implicated party, since you do not yet know what the outcome will be. You would think this obvious, but practical experience shows it is not :shrug:

jni:
Ok so the way that the PSF CoC stuff works according to Noa Tamir (sklearn+others) is, you write to their email:

conduct-wg@python.org

and say you want to partner with them. To me @stefanv this bit:

> You would think this obvious, but practical experience shows it is not 🤷

means that you need people who have experience or dedicated training (or both) to handle complaints. Reading the CoC here, I would say that you need at least a fallback on the PSF in case of conflicts of interest, at a minimum.

jni:
This summary from the PyPA adopting the PSF CoC from Brett Cannon is a nice summary and I think a good way to go:

https://discuss.python.org/t/move-to-the-psf-code-of-conduct/3876/19

> 1. Have a CODE_OF_CONDUCT.md file (repo or org).
> 2. MD file links to [https://www.python.org/psf/conduct/ ](https://www.python.org/psf/conduct/).
> 3. MD file lists who the admins/moderators are to contact in case of an incident.
> 4. MD file mentions that in cases which involve the above admins/moderators, feel free to email [conduct-wg@python.org](mailto:conduct-wg@python.org).
> 5. Admins/moderators are expected to report up to the Conduct WG any incidents for recording purposes.
> 6. We promise to help admins/moderators with any questions/needs they may have.
> 7. The PyPA also gets to directly use the PSF CoC and any ancillary things (i.e. reporting guidelines and enforcement guide).

stefanv:
That sounds good.

FirefoxMetzger:
Wow, I can't believe that it's been over a month already since I had time to sit down and push this forward. Sorry for the wait!

> My personal attitude towards CoCs is that I know nothing so I'd rather copy verbatim from people who know something. 😜

I agree with you on the knowing nothing part @jni. I haven't had any CoC issues since I joined ImageIO (or ever?), so there is definitely a learning experience to be had.

> the source of the CoC text (I notice you reference NumPy and NumFOCUS, but only at the very end of the PR)

The [NumPy CoC](https://numpy.org/code-of-conduct/#endnotes) mentioned SciPy as their source of inspiration in the endnotes and SciPy [does the same](https://docs.scipy.org/doc/scipy/dev/conduct/code_of_conduct.html#endnotes), so I just copied that behavior.

> the changes from the source material

It will be difficult to provide a concise diff because one major change is that I reorganized the sectioning of the reporting guide. The one I found on NumPy was comprehensive, but a bit chaotic to navigate so I streamlined it so that I can easily find the agreed-upon procedure if it comes down to handling a case.

Apart from that, it is mostly me kicking out statements that don't make sense for ImageIO, or that are tautological. For example, within the purely virtual space of ImageIO the statement to refrain from "inappropriate physical contact" doesn't make much sense. Similarly, the statement that we don't allow "other unethical or unprofessional conduct" in a document that specifies what is deemed ethical and unethical behavior is tautological.

Another point is that some phrases in the NumFOCUS CoC elegantly shoot itself in the foot. For example, it prohibits "Sharing private content, such as emails sent privately or non-publicly, or direct message history, without the sender’s consent". However, if we are strict with this, then it means we can't say something like "I discussed XYZ with person A over a beer last week, and she thinks the following idea might work: *lays out idea*" without asking person A for consent first. Worse, a person that wants to report "private harassment" (a breach of NumFOCUS's CoC) can't share any evidence of this without breaking the very same CoC or getting consent from the reported person.

I get the intention behind the above statements, but I would like to be conservative and stick to clear examples of bad behavior in the text.

> The PSF was offering to help arbitrate CoC issues.

I agree that it would make sense to have a fallback in place for cases where I have a conflict of interest. This could be the PSF (happy to reach out to them; even better if somebody could make an introduction), but I wonder if NumFOCUS might be the better candidate for this considering that I am adding the CoC to apply for affiliation.

Do you guys think the NumFocus conduct group might agree to a similar arrangement as PyPA and PSF have, too?

stefanv:
NumFOCUS does not currently have a group of experts that provide that function.

Issue #904, opened 2022-11-03
v3.imwrite to grayscale PNG of image with shape HWC does not work


hookxs:
Numpy image of shape HWC with C==1 (grayscale) is not correctly saved to .png when calling imageio.imwrite (defaulting to pillow backend). I would swear this used to work in the (not so distant) past because I have results from some older code where these images are saved correctly. Now imageio just silently fails without throwing any error (as a result, I was using the code for some time after imageio update without noticing that the logged images are rubbish). Grayscale HW images work fine as expected. I don't know if this is intentional feature or a bug but I think that silent conversion from HW1 to HW would be much better than silent failure. Or even throwing an error would be better. I have imageio 2.22.0 and pillow 9.2.0.

FirefoxMetzger:
That sounds really annoying. I can indeed reproduce this on master:

```python
import imageio.v3 as iio
import numpy as np

buffer = iio.imwrite("<bytes>", np.ones((128, 128, 1), dtype=np.uint8), extension=".png")
print(iio.improps(buffer).shape)
# prints (1, 128, 1)
```

This is definitely a bug and you should have gotten an exception. The image is recognized as a batch of images and, writing a PNG should fail and raise an exception since PNG doesn't support batches. We were relying on pillow to raise the exception here since their PNG writer doesn't support the kwarg to indicate batch writing; however, it looks like pillow just silently eats unknown kwargs without raising anything...

I will see what changes need to be made to handle this correctly and make a PR. 

Note, however, that the PR will simply add an exception. You still can't write a `(128, 128, 1)` image to PNG, even as a single image. This would lose the last dimension and, in other formats, we made the decision to raise an exception. This is more explicit compared to `np.squeeze`ing under the hood, which might surprise the user. I will do the same here.

hookxs:
Thanks. I would think that silent conversion to HW would work but exception is fine of course if it's consistent with the rest of the library.

FirefoxMetzger:
The reason for raising an exception is that you should be able to expect your data to be preserved; especially for lossless formats like PNG. This means that something like the following should work:

```python
import numpy as np
import imageio.v3 as iio

rng = np.random.default_rng()

source_img = rng.integers(0, 255, (128, 128, 3), dtype=np.uint8)
buffer = iio.imwrite("<bytes>", source_img, extension=".png")
read_img = iio.imread(buffer)

assert np.allclose(read_img, source_img)
```

[...]

PR #906, opened 2022-11-04
BUG: fix silent failure when writing gray images with explicit color channel using pillow

FirefoxMetzger:
Closes: https://github.com/imageio/imageio/issues/904

Writing gray images with explicit color channel (`(A, B, 1)`) with pillow fails silently and produces a `(1, B, 1)` image. This PR catches those cases and raises an exception instead.

This solution does alias with auto-detecting batches of `(B, 1)` images, i.e., it will also raise an exception for - say - a GIF of shape `(batch, B, 1)`; however, this can be avoided by explicitly setting `is_batch=True` and avoiding the auto-batch-detection. Considering that the scenario of writing such images is very rare compared to the scenario of an explicit grayscale color channel, I am happy to raise by default.

### DIFF
--- a/imageio/plugins/pillow.py
+++ b/imageio/plugins/pillow.py
@@ -308,9 +308,10 @@ class PillowPlugin(PluginV3):
             )
         elif ndimage.ndim == 2:
             is_batch = False
+        elif ndimage.ndim == 3 and ndimage.shape[-1] == 1:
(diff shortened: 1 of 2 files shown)


FirefoxMetzger:
This PR also refactors the pillow tests to use the, now standard, `import imageio.v3 as iio` syntax.

Issue #818, opened 2022-06-03
imageio v3 api reads all pages of a multipage TIFF at once


thvoigtmann:
I have a two-page TIFF, one 8bit grayscale image of shape (15,10) per page, and I am expecting `imageio.v3.imiter` to iterate over the two pages. Yet, it reads them both at once:
```python
>>> from skimage import data_dir
>>> file = data_dir + '/multipage.tif'
>>> import imageio
>>> len([*imageio.v3.imiter(file)])
1
>>> for img in imageio.v3.imiter(file):
...     img.shape
... 
(2, 15, 10)

```

The same code prior to [#805](https://github.com/imageio/imageio/pull/805) fulfills my expectation and iterates over the two pages:
```python
>>> from skimage import data_dir
>>> file = data_dir + '/multipage.tif'
>>> import imageio
>>> len([*imageio.v3.imiter(file)])
2
>>> for img in imageio.v3.imiter(file):
...     img.shape
... 
(15, 10)
(15, 10)
```

The TIFF file in question is part of `skimage`: https://github.com/scikit-image/scikit-image/blob/c95a8fd46a49a38de1f6032c84d9d61160b78283/skimage/data/multipage.tif
It appears to have been created with ImageMagick.

thvoigtmann:
@FirefoxMetzger It doesn't appear to be specific to that TIFF file, nor to whether it's grayscale or RGB. I attach a little TIFF that I made with the GIMP by exporting two layers, each of shape (15, 10, 3). Same story as above, with the newest `imageio.v3.imiter` I get a single array of shape (2, 15, 10, 3).
File in a ZIP archive (supported by github): [multipage.zip](https://github.com/imageio/imageio/files/8830673/multipage.zip)

FirefoxMetzger:
Sorry for taking such a long time to get around to the issue @thvoigtmann . This month doesn't leave me much time to work on side projects.

@almarklein perhaps you have some time to investigate this? Otherwise, it may have to wait until I am a little more free.

almarklein:
Now that I think of it, I think this can be explained by the addition of the "?"in #805. This means that without specifying whether it's an image or volume, it assumes that one is interested in the full array:

https://github.com/imageio/imageio/blob/aba1eecb29cbbbfbd688ccba878c67d53b59204a/imageio/plugins/tifffile.py#L428-L449

[...]

Issue #437, opened 2019-03-05
Support passing more/all args to tiffile


almarklein:
Spun out of #307.

kmilos:
> I'm reluctant to support and document _all_ parameters, but we might make it possible to pass a dict that represent kwargs for tifffile and refer to the tifffile docs.

Exactly, that's what I meant as well

berl:
Just to chime in that it would be nice to have access to the tifffile kwargs through a dictionary. I'm currently bothered by the multiprocessing used in tifffile, so I want to use
`tifffile.imread(blah.tif, maxworkers=1)` but
 `imageio.imread(image.tif, maxworkers=1)` gives 
`TypeError: unexpected keyword argument: maxworkers`
I realize that `tifffile` is a bit of a moving target (I'm guessing that's why its ~1yr behind cgohlke's source code), but the keyword access would help `imageio` through some of the non-breaking changes

cgohlke:
> I'm currently bothered by the multiprocessing used in tifffile

May I ask if there is any specific bug with the use of multi-threading in tifffile? Or do you simply want to turn of internal multi-threading because you are using your own higher level multiprocessing?

One issue that I am aware of is that `ThreadPoolExecutor` "swallows" Python exceptions.

Tifffile can use multi-threading on 4 nested levels, which can each significantly speed up reading large amounts of data from TIFF files in certain cases. I'm trying to find a more consistent API with sensible defaults...

berl:
@cgohlke no actual bugs, I just want to turn it off.  My current use case (lots of single-plane, single-channel TIFs) benefits from some parallelization, but I'd like to manage it myself.
thanks for all the work on tifffile, btw!

PR #914, opened 2022-11-29
FEAT: tifffile v3

FirefoxMetzger:
Closes: https://github.com/imageio/imageio/issues/818, https://github.com/imageio/imageio/issues/437 (must be closed manually)

This PR updates the tifffile plugin to use the v3 plugin API. It still needs some documentation, but I want to park it here so that interested parties can give feedback :)

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -62,7 +62,7 @@ SphinxDocString._str_member_list = lambda self, name: []
 
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ["_templates"]
-autodoc_mock_imports = ["av", "cv2", "imageio_ffmpeg"]
+autodoc_mock_imports = ["av", "cv2", "imageio_ffmpeg", "tifffile"]
 
 # The suffix of source filenames.
 source_suffix = ".rst"
--- a/docs/reference/index.rst
+++ b/docs/reference/index.rst
@@ -99,3 +99,4 @@ support.
     imageio.plugins.spe
     imageio.plugins.swf
     imageio.plugins.tifffile
+    imageio.plugins.tifffile_v3
--- a/imageio/config/extensions.py
+++ b/imageio/config/extensions.py
@@ -822,7 +822,7 @@ extension_list = [
     ),
     FileExtension(
         extension=".stk",
-        priority=["tifffile","TIFF"],
+        priority=["tifffile", "TIFF"],
     ),
     FileExtension(
         name="ShockWave Flash",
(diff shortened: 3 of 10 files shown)


FirefoxMetzger:
Note to self: This PR needs updating to use `index=...` instead of `index=None` when bulk reading.

lschr:
Thanks for this!

When looking through the code, I found a possible inconsistency: in `metadata`, it seems like `index` refers to the page, while in all other methods, `index` refers to the series and `page` refers to the page.

FirefoxMetzger:
@ishr Yes, that's true. I have to dig a bit further into tifffile to see if and how series-level metadata is handled and then update `TifffilePlugin.metadata`.

FirefoxMetzger:
Okay, now everything should be implemented and documented. What is still missing is to get full code coverage and then a final review.

lschr:
Awesome!

It seems like the documentation of `properties()` is not quite right though. What is described as behavior for `index=None` is actually true for `index=Ellipsis`.

FirefoxMetzger:
Oh yes, good catch. The documentation also needs to update to `index=...` instead of `index=None`

FirefoxMetzger:
@mkcor I know you have been using tifffile a lot, so this might be interesting for you. I've written a v3 plugin for tifffile to expose all its "new" features in ImageIO. Feel free to give it a try and tell me what you think :)

The PR is ready and I will merge towards the end of the week unless anything major pops up.

mlt:
I'm getting a regression with metadata description field containing XML (Open Microscopy Environment). I can navigate metadata dict alright with v2 but not with v3 & immeta :(

```python
ValueError                                Traceback (most recent call last)
Cell In[13], line 12
     10 image_rgb = iio.imread(filename)
     11 # meta = image_rgb.meta
---> 12 meta = iio.immeta(filename)
     13 fig = plt.figure()
     14 ax = fig.gca()

File ~\miniconda3\envs\scikit-env\lib\site-packages\imageio\v3.py:250, in immeta(uri, index, plugin, extension, exclude_applied, **kwargs)
    247     call_kwargs["index"] = index
    249 with imopen(uri, "r", **plugin_kwargs) as img_file:
--> 250     metadata = img_file.metadata(**call_kwargs)
    252 return metadata

File ~\miniconda3\envs\scikit-env\lib\site-packages\imageio\plugins\tifffile_v3.py:307, in TifffilePlugin.metadata(self, index, page, exclude_applied)
    305                 metadata.update(flavor_metadata[0])
    306             else:
--> 307                 metadata.update(flavor_metadata)
    308 else:
    309     # tifffile may return a TiffFrame instead of a page
    310     target = target.keyframe

ValueError: dictionary update sequence element #0 has length 1; 2 is required
```

FirefoxMetzger:
@mlt Interesting find. Could you create a dedicated issue for that?

PR #916, opened 2022-12-08
GitHub Actions add Python 3.11 to the testing

cclauss:


.github/workflows/ci.yml
```diff
@@ -21,10 +21,10 @@ jobs:
     name: Build Docs
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
-    - uses: actions/setup-python@v2
+    - uses: actions/checkout@v3
+    - uses: actions/setup-python@v4
       with:
-        python-version: '3.10'
+        python-version: '3.11'
```
FirefoxMetzger, line 27:
```suggestion
        python-version: '3.10'
```

I think I prefer testing the docs on 3.10 for the time being. There is definitely value in having 3.11 in the unit tests, but I don't think the docs will get any direct benefit. RTD actually builds them on 3.9, so if anything we may wish to match that and then upgrade here once RTD bumps that default.

.github/workflows/ci.yml
```diff
@@ -35,8 +35,10 @@ jobs:
     name: "Linting"
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-python@v2
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: 3.x
```
FirefoxMetzger, line 41: Could you elaborate why we need to specify 3.x here?

FirefoxMetzger (commented):
Thanks for the improvement @cclauss 🚀 

I'm definitely happy to try CI on 3.11 again now that it is officially released. I experimented with this during the summer, but a lot of backends didn't have precompiled binaries at that stage, which made building on CI a nightmare. Let's see how it does this time around :)

FirefoxMetzger:
Ah, looks like pyav is not supported on 3.11 yet. I will have a look at this tomorrow to see if we can fix this.

cclauss:
PyAV 10 does support Py3.11 but...
https://github.com/imageio/imageio/blob/b3b040a79b96718cef89fcfa9149cda151763309/setup.py#L192

.github/workflows/ci.yml
cclauss: v4 of this action refuses to guess which `python-version` to use...
* https://github.com/actions/setup-python/releases/tag/v4.0.0

FirefoxMetzger:
> PyAV 10 does support Py3.11 but...

That's what I wanted to test later today. The problem is that it also has a bug that breaks one of our unit tests (see: https://github.com/PyAV-Org/PyAV/issues/1045), which is why it is currently disabled. If this is the only thing keeping us from testing ImageIO on 3.11 then I think we could also xfail that unit test instead and raise a warning for metadata read with pyav 10.0.0 instead.

I'll check in the afternoon, when I have more time to investigate.

cclauss:
Rock and roll!!

.github/workflows/ci.yml
```diff
@@ -102,6 +105,11 @@ jobs:
         if: matrix.TEST_FULL == 1
         run: |
             pip install -q -U simpleitk astropy
+      - name: Downgrade PyAV
+        # PyAV 10.0.0 is needed on py3.11 but has issues, so we only use it on
+        # py3.11 for now
+        if: matrix.PYAV_10_0_0 != 1
+        run: pip install -U av<=10.0.0
```
cclauss: pip has 5 different --uxxx options so it is more self-documenting to be explicit.
```suggestion
        run: pip install --upgrade av<=10.0.0
```

cclauss: It is also possible to:
pip install --upgrade av<10.0.0; ; python_version < '2.11'
pip install --upgrade av>=10.0.0; ; python_version >= '2.11'

FirefoxMetzger:
I am seeing the lytro tests break on the windows CI with py3.11 due to an access violation. On my local windows machine, however, I can't replicate this behavior.

I have to put this down for now, because I need to deal with another task for today, but I plan to come around to it later this day to see if I can figure out why this breaks.

cclauss:
It looks like it passed on the last test run and failed on the second last run.

FirefoxMetzger:
Okay, so it was a sporadic failure on one CI machine. Let's see if it pops up in other PRs moving forward or if it is one of GH Action's runners that was breaking.

LGTM!

Issue #923, opened 2022-12-26
ValueError: Could not find a backend to open `video.ogv` with iomode `r?`.


ArtyomVancyan:
I am trying to read a `.ogv` file with `imageio.get_reader`, and it raises this error.
### STDOUT
```text
Traceback (most recent call last):
  File "/home/artyom/Python/thumbnails/ffmpeg/__init__.py", line 8, in __init__
    self.reader = imageio.get_reader(filename)
  File "/home/artyom/anaconda3/lib/python3.9/site-packages/imageio/v2.py", line 159, in get_reader
    image_file = imopen(uri, "r" + mode, **imopen_args)
  File "/home/artyom/anaconda3/lib/python3.9/site-packages/imageio/core/imopen.py", line 298, in imopen
    raise err_type(err_msg)
ValueError: Could not find a backend to open `valerian-1080p.ogv`` with iomode `r?`.
```

### PACKAGE VERSIONS
imageio==2.23.0
imageio-ffmpeg==0.4.7

FirefoxMetzger:
OGG Vorbis should be supported by pyAV. imageio-ffmpeg should support this as well, but we might need to manually add the extension for this to work (due to how the legacy API works).

Do you have a snippet of code that reproduces the behavior?

ArtyomVancyan:
> Do you have a snippet of code that reproduces the behavior?

No, I just debugged the `imopen`.

> imageio-ffmpeg should support this

It seems, it does not :)

FirefoxMetzger:
I am asking because when I tried it locally, I could read and write OGV without issues:

```python
>>> import imageio.v3 as iio
>>> test_frames = iio.imread("imageio:newtonscradle.gif")
>>> iio.imwrite("test.ogv", test_frames, codec="libtheora")
>>> frames = iio.imread("test.ogv")
>>> frames.shape
(36, 150, 200, 3)
```

That said, I am using pyAV which is only available in the v3 API (`legacy_mode=False`), which is why I would need more context to better understand your situation and see what could/should be done about it. For this, a snippet that allows me to reproduce your behavior is always a good start.

ArtyomVancyan:
Sure I have tried this before, and that works fine, but in my case, I want to read all formats of videos with a single function, unfortunately when reading another format of video such as `.mp4`, my computer memory leaks for some reason.

FirefoxMetzger:
Interesting. Does the memory leak disappear when you set the legacy mode to `False`? 

[...]

PR #924, opened 2022-12-27
GH-923: Add an opportunity to pass `legacy_mode`, keeping the default behavior

ArtyomVancyan:
I have investigated and found that the `.ogv` format requires the `legacy_mode` to be `False`, and `get_reader` uses a static `True`.

In this change, I have kept the default behavior and given an opportunity to change the `legacy_mode` from outside, so that we can use the `get_reader` function for reading `.ogv` files as well.


FirefoxMetzger:
Thanks for the PR @ArtyomVancyan . I have a few general thoughts before reviewing in more detail:

`legacy_mode` is an internal flag that facilitates backward compatibility while we reorganize internals. It will disappear sooner or later, so I wouldn't want to make it part of the public API unless there is a very good reason to do so. In particular, `get_reader` (which is part of the legacy API) should behave according to legacy rules, which is why it is statically set to `True`. 

Regarding OGV specifically, it should already be supported in the v3 API, so I would ask you to give this a shot first and see where/if this fails. New code should prefer using `imageio.v3` in general. `get_reader` and other functions in `imageio.v2` exist to allow libraries to migrate at their own pace, but should eventually be changed to using v3

Could you add some detail to the investigation so that I can see why this is necessary?

ArtyomVancyan:
> Could you add some detail to the investigation so that I can see why this is necessary?

By saying an **investigation** I meant debugging the code and finding the cause of the issue. The debugger has thrown the error on `imopen`, and then I noticed that `plugin` and `extension` are not overridden, and only `legacy_mode` is remaining, which was not using the default value defined at `imopen`, but uses a static `True`. So I changed it, and it works now.

Additionally, thanks for the clarification about `legacy_mode`.

PR #930, opened 2023-01-09
Avoid np.stack in pyav plugin.

### LINKED ISSUE #927: PyAV: Avoiding the use of np.stack
https://github.com/imageio/imageio/blob/4658049bc37b6e89796233306df393d8a51b68bd/imageio/plugins/pyav.py#L597

Over the years, I've found that np.stack is pretty detrimental to performance. The reason is simple: one is copying memory around. That really hurts especially when you are doing video processing.

One might want to optimize these calls away.

You can likely "check for the same dimensions" with:

```
if any(f.shape != ndimage[0].shape for f in ndimage):
    raise ValueError("All frames should have the same shape")
```

In either case, this is a pretty minor for me since I don't use that codepath (yet) but I figured i would flag it as a potential contribution for new contributors.

cgkoutzigiannis:
Closes #927.
Removed the call to np.stack and added a new test case, where I reshaped one frame, to validate new functionality.

### DIFF
--- a/tests/test_pyav.py
+++ b/tests/test_pyav.py
@@ -265,8 +265,8 @@ def test_raises_exception_when_shapes_mismatch(test_images, tmp_path):
     )
 
     # touch the codepath that creates a video from a list of frames
-    frame_list = list(frames_expected)
-    frame_list[0].reshape(200, 150)
+    frame_list = list(frames_expected)    
+    frame_list[0] = frame_list[0].reshape(200, 150)
 
     with pytest.raises(ValueError, match="All frames should have the same shape"):
         iio.imwrite(
--- a/tests/test_pyav.py
+++ b/tests/test_pyav.py
@@ -259,6 +259,24 @@ def test_gif_write(test_images, tmp_path):
     # with iio.v3.imopen("test2.gif", "w", plugin="pyav", container_format="gif", legacy_mode=False) as file:
     #     file.write(frames, codec="gif", out_pixel_format="gray", in_pixel_format="gray")
 
+def test_raises_exception_when_shapes_mismatch(test_images, tmp_path):
+    frames_expected = iio.imread(
+        test_images / "newtonscradle.gif", plugin="pyav", format="gray"
+    )
+
+    # touch the codepath that creates a video from a list of frames
+    frame_list = list(frames_expected)
+    frame_list[0].reshape(200, 150)
+
+    with pytest.raises(ValueError, match="All frames should have the same shape"):
+        iio.imwrite(
+            tmp_path / "test.gif",
+            frame_list,
+            plugin="pyav",
+            codec="gif",
+            out_pixel_format="gray",
+            in_pixel_format="gray",
+        )
 
 def test_gif_gen(test_images, tmp_path):
     frames = iio.imread(
--- a/tests/test_pyav.py
+++ b/tests/test_pyav.py
@@ -260,16 +260,14 @@ def test_gif_write(test_images, tmp_path):
     # with iio.v3.imopen("test2.gif", "w", plugin="pyav", container_format="gif", legacy_mode=False) as file:
     #     file.write(frames, codec="gif", out_pixel_format="gray", in_pixel_format="gray")
 
+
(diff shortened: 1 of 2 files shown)


FirefoxMetzger:
Thanks for the contribution @cgkoutzigiannis ! I'll let the tests run and review the PR tomorrow morning.

FirefoxMetzger:
While reviewing your PR I got curious what pyav would do if it encounters a frame with the wrong shape. Turns out that pyav will internally convert the frame into the shape and pixel format that the stream has specified. (@hmaarrfk FYI)

<details>
<summary>Example writing non-uniform frames.</summary>

```python
import av
import numpy as np
import imageio.v3 as iio

file_name = "test.mp4"
fps = 24

img = 255*np.ones((200, 150, 3), dtype=np.uint8)
frame = av.VideoFrame.from_ndarray(img)

img = np.zeros((255, 255), dtype=np.uint8)
frame2 = av.VideoFrame.from_ndarray(img, "gray")

video_frames = iio.imread("imageio:cockatoo.mp4", plugin="pyav")
print(f"video frames shape: {video_frames.shape}")

with av.open(file_name, "w") as file:
    stream = file.add_stream("libx264", fps)
    stream.pix_fmt = "gray"

    for idx in range(fps):
        for packet in stream.encode(frame):
            file.mux(packet)

    for idx in range(fps):
        for packet in stream.encode(frame2):
            file.mux(packet)

    for frame in video_frames:
        frame = av.VideoFrame.from_ndarray(frame)
        for packet in stream.encode(frame):
            file.mux(packet)

    for packet in stream.encode():
        file.mux(packet)

frames = iio.imread(file_name, plugin="pyav")
print(frames.shape)

```

</details>

This means that we could theoretically remove this check entirely, because of the internal conversion. At the same time, I think we should raise an exception because it is surprising behavior that your data gets reshaped without you requesting it, i.e., I would like the following to always work:

```python
some_frames = [...]
iio.imwrite("video.mp4", some_frames)
read_frames = iio.imread("video.mp4")

for frameA, frameB in zip(some_frames, read_frames):
    assert frameA.shape == frameB .shape

    # equal up to encoding-related errors
    assert almost_equal(frameA, frameB )

```

How do you think, @cgkoutzigiannis @hmaarrfk ?

tests/test_pyav.py
```diff
@@ -259,6 +259,24 @@ def test_gif_write(test_images, tmp_path):
...
     #     file.write(frames, codec="gif", out_pixel_format="gray", in_pixel_format="gray")
 
+def test_raises_exception_when_shapes_mismatch(test_images, tmp_path):
+    frames_expected = iio.imread(
+        test_images / "newtonscradle.gif", plugin="pyav", format="gray"
+    )
+
+    # touch the codepath that creates a video from a list of frames
+    frame_list = list(frames_expected)    
+    frame_list[0] = frame_list[0].reshape(200, 150)
```
FirefoxMetzger, line 269:
```suggestion
def test_raises_exception_when_shapes_mismatch(tmp_path):
    frame_list = [
        np.ones((200, 150), dtype=np.uint8),
        np.ones((256, 256), dtype=np.uint8),
    ]
```

We don't need to load an entire video for this test. Since we expect a failure, it is enough to have two frames

FirefoxMetzger:
> I think we should still raise the exception since this is a implementation detail of pyav

Okay, in that case, we should perform this check in `PyAVPlugin.write_frame` instead of `PyAVPlugin.write`. This way, we can also handle incremental writing and bulk writing in one go. This would mean extending the following if-block with an `elif` that raises an exception if the frame's shape doesn't match the stream's shape:

https://github.com/imageio/imageio/blob/c591a1c4ea20269382877110ce486c5bf84bd99d/imageio/plugins/pyav.py#L865-L867

@cgkoutzigiannis Could you move the exception to that function?

FirefoxMetzger:
@cgkoutzigiannis It's been a while since this PR has seen its last update. Are you still interested in adding this, or may someone else take it up?

FirefoxMetzger:
I had some time so I finished this one up. Thanks @cgkoutzigiannis for the contribution 🚀

Issue #935, opened 2023-01-20
PyAVPlugin.metadata raise TypeError when video does not contains duration


kchaiyo:
I tried to use imageio.v3.immeta to read video's metadata. The video binary is part of live stream, so there is no EBML field named /Segment/Info/Duration in the binary.

To reproduce, pass a video without duration into immeta as follow

`metadata = imageio.v3.immeta(data, index=..., plugin='pyav')`

According to the stack trace and some debugging, the error is raised from line 709 in plugin/pyav.py

`duration = float(self._video_stream.duration * self._video_stream.time_base)`

A fix could be made by not populate duration field into output metadata dict when self._video_stream.duration is not present.

imageio version is 2.24.0
av (pyav) version is 9.2.0

FirefoxMetzger:
Interesting. For streams, there is indeed no duration (or infinity?). So far, we only have limited support for streams, but I am very happy to extend it.

@kchaiyo I like your suggestion. Do you know if/how we could generate a test input for this? If so, we can create a suitable regression test for it and incorporate the update :)

kchaiyo:
Originally I found the error from data received from AWS's Kinesis Video Stream. The stream was produced by this gstreamer command.
`gst-launch-1.0 -v v4l2src device=/dev/video0 ! videoconvert ! videorate ! video/x-raw,format=I420,width=640,height=480,framerate=15/1 ! x264enc  bframes=0 key-int-max=45 bitrate=500 tune=zerolatency ! video/x-h264,stream-format=avc,alignment=au ! kvssink stream-name=<stream-name> storage-size=128`
The significant 

I was trying to reduce the steps necessary to reproduce this issue by trying to save the stream before passing to Kinesis Video Stream with this command.
`gst-launch-1.0 -v v4l2src device=/dev/video0 ! videoconvert ! video/x-raw,format=I420,width=640,height=480,framerate=15/1 ! x264enc  bframes=0 key-int-max=45 bitrate=500 tune=zerolatency ! video/x-h264,stream-format=avc,alignment=au ! matroskamux ! filesink location=capture.mkv`
However, for this case duration field was exist with value of 0, so the issue cannot be reproduced.

kchaiyo:
I was trying to reduce the steps necessary to reproduce this issue. This command would record a video stream from camera and save into an mkv file locally. The video stream is no needed to be long, as all we need is the metadata in header, so we can interrupt the command just after few seconds of recording.

[...]

PR #937, opened 2023-01-27
BUG: Handle unknown video duration in pyav metadata

kchaiyo:
Implement a check if video container has value in duration field before populate it's value in metadata

only apply to plugin pyav

### DIFF
--- a/imageio/plugins/pyav.py
+++ b/imageio/plugins/pyav.py
@@ -742,7 +742,9 @@ class PyAVPlugin(PluginV3):
                 }
             )
             if self._video_stream.duration is not None:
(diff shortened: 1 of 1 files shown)


kchaiyo:
Issue mentioned in #935.
No test implement for this fix yet.

FirefoxMetzger:
Thanks for kicking this off @kchaiyo ! I will have a look once I get to do my next round of implementation on the pyAV.

FirefoxMetzger:
I don't have an easy way to create a test video for this to set up a regression test. That said, I am okay merging without one here because it will be covered once we get around to implementing support for streams (in particular webcams).

Issue #926, opened 2023-01-03
PyAV Plugin: Allow ffmpeg to reformat the color if it wants to?


hmaarrfk:
I was comparing the performance of different video encoding paths and ultimately, I found that:

https://github.com/imageio/imageio/blob/4658049bc37b6e89796233306df393d8a51b68bd/imageio/plugins/pyav.py#L864

is a source of potential slowdowns. The slowdown was most apparently when using it with nvidia's encoders, but it was still "measurable" with libopenh264

I think it is ultimately due to some multi threading or GPU optimizations that can be done. Or something...

Would it be possible to avoid that call? Was there a reason to add it?

<details> <summary> potentially useful benchmark </summary>

```python
import numpy as np
from imageio import imopen
import av
from pathlib import Path


images = np.random.randint(0, 256, (512, 256, 256, 3), dtype='uint8')
encoder = 'libopenh264'

# %%
def doit_imageio():
    total_frames = images.shape[0]

    with imopen("output_pyav.mp4", "w", plugin="pyav") as file:
        file.init_video_stream(encoder, fps=30, pixel_format='yuv420p')
        file.set_video_filter(None, None)
    
        for i in range(total_frames):
            file.write_frame(
                images[i], pixel_format="rgb24"
            )

# %%
def doit_pyav():
    total_frames = images.shape[0]
    height, width = images.shape[1:3]
    
    container = av.open("pyav.mp4", mode="w")
    stream = container.add_stream(encoder, rate=30)
    stream.width = width
    stream.height = height
    stream.pix_fmt = "yuv420p"
    stream.codec_context.gop_size = 60
    stream.codec_context.thread_count = 4
    # sc_threshold to 0? to avoid scene cut detection
    # keyint_min?
    
    
    for frame_i in range(total_frames):
        image = images[frame_i]
        frame = av.VideoFrame.from_ndarray(image, format="rgb24")
    
        for packet in stream.encode(frame):
            container.mux(packet)
    
    # Flush stream, cannot use the stream after a flush command
    for packet in stream.encode():
        print("packet")
        container.mux(packet)
    
    # Close the file
    container.close()

# %%
doit_pyav()
# doit_imageio()
```
</details>

FirefoxMetzger:
Interesting. I will have a closer look once I have access to my dev machine again (we are traveling back home tomorrow).

[...]

Issue #933, opened 2023-01-16
pyav read should remember the current index and avoid a seek if it can


hmaarrfk:
https://github.com/imageio/imageio/blob/c591a1c4ea20269382877110ce486c5bf84bd99d/imageio/plugins/pyav.py#L457

Seeking videos is expensive since one can only seek to a keyframe.

This should "avoid" seeking if it is can detect that the codec context is already at the correct location.

FirefoxMetzger:
Hm, what is the correct location? Are you thinking about the scenario of reading consecutive frames (i.e. `index=old_index+1`)?

hmaarrfk:
This is "quite a common" scenario.

At the very least, it should remember:

> If i continue to use the same generator I used previously, the next index is "previous index + 1". this means I do not need to seek! Otherwise, seek.

This little optimization is really helpful.

Otherwise, using a video like "random access" provides terrible performance.

FirefoxMetzger:
I'm asking because the "read frame n+1" use case fits better with the `iter` paradigm in my opinion. Currently you can do this via:

```python
import imageio.v3 as iio

for idx, frame in enumerate(iio.imiter("imageio:cockatoo.mp4", "r")):
    if idx < start_idx:
        continue

    ...  # do something useful with each frame

# alternatively, you can use iter outside a loop
# (but within an imopen context)
with iio.imopen("imageio:cockatoo.mp4", "r") as file:
    frame_gen = file.iter()
    for _ in range(start_index):
        next(frame_gen)

    # use the generator to get the next image when needed
    ...
    img = next(frame_gen)
    ...
```

That said, decoding the first `start_idx` many frames can be painful as well. I always wanted a `start_idx` and/or `start_time` kwarg for these methods in pyav. This might be as simple as adding a call to `self._seek` (+- some edge cases with filters).

There is also the "read a frame at `index` where `index > previous_idx`" use case. This case is more interesting, in that we could be smart and check if the new `index` lies between `previous_idx` and the next keyframe, in which case it is indeed faster to keep decoding than to jump back and decode from there.

hmaarrfk:
I understand the `iter`. "Iterators" work sometimes. But other times you just want to index it yourself.

hmaarrfk:
I think the `start_index` issue is one of them that doesn't work well with the iterator context.

for data analysis, I think of the data as a volume, so sometimes it is really convenient to use the indexing syntax.

FirefoxMetzger:
[...]

Issue #932, opened 2023-01-11
Preallocate result array in PyAVPlugin.read while bulk reading


hmaarrfk:
as a followup to https://github.com/imageio/imageio/issues/927

The PyAV plugin read function uses stack. This is somewhat less than desirable.

What will happen is:

1. PyAV will create an internal frame
2. Reading a single frame will create a copy of the data (this is likely inevitable)
3. Once all the data has been collected, it will then be-recopied.

It would be good to:

1. Preallocate the data
2. Iterate through the frames. As we iterate, we copy the data from PyAV's frame to the buffer we created.

https://github.com/imageio/imageio/blob/c591a1c4ea20269382877110ce486c5bf84bd99d/imageio/plugins/pyav.py#L426

Mostly, I'm going through trying to integrate with this and writing some notes on findings.

FirefoxMetzger:
Well spotted. We can add a fast-path for this if the number of frames inside the video is known in advance. I.e., we can (now) use `PyAVPlugin.properties().shape` to get the shape and allocate an appropriately sized array in advance and avoid the copy introduced by `np.stack`. This would be a nice addition and I'm happy to accept a PR for it 👍.

We will likely also have to provide a `out` argument to `PyAVPlugin.iter` to allow passing the view from `read` to `iter`. This would be nice to have as part of the public API anyway, because it allows optimizations like:

```python
with iio.imopen("imageio:cockatoo.mp4", "r", plugin="pyav") as file:
    props = file.properties()
    buffer = np.empty(props.shape, dtype=props.dtype)
    for _ in file.iter(out=buffer):
        pass  # do stuff with buffer
```
which avoids allocating a new array each frame.

At the same time, we also need to keep the path that uses `np.stack`, because this doesn't work in the general case. Ffmpeg may return a frame count of -1 for some videos whenever it can't figure out the number of frames in the video. In these cases, we need to fall back on the current implementation.


As a future step, it would be really awesome if we could hand that buffer upstream (to pyav) and avoid another copy. Another enhancement would then be to add `out`  to `PyAVPlugin.read`, which would allow users to preallocate the array.

Would you be interested in contributing something like this?

Issue #929, opened 2023-01-09
Why are you overwriting the timebase for pyav


hmaarrfk:
https://github.com/imageio/imageio/blob/c591a1c4ea20269382877110ce486c5bf84bd99d/imageio/plugins/pyav.py#L785

It seems strange that this is happening. In my experience you do not need to do this, and can provide fractional timebases to pyav for more precise playback speeds (3.5 fps)

Is there a technical reason for this line or can we remove it?

Best,

Mark

hmaarrfk:
> That said, I don't know what the implications of this are nor do I know the limits of this.

Yeah, there are some liminations that I learned of during my testing. One of them is that certain codecs, mpeg, only support 16 bit denominator (65k max).


> by various media players

That i haven't tested so intricately.

I think that av might require some commonality between codec context time base, and the stream's

https://github.com/PyAV-Org/PyAV/blob/main/av/container/output.pyx#L103

I'm not sure how that reconciles with it.

It might be that you can remove this. People that want compatibility with "media players", may cast to an int.

This is the code that i use my (new) av wrapper:
```
        stream = container.add_stream(
            codec,
            rate=Fraction(frame_rate).limit_denominator(65_535)
        )
```

FirefoxMetzger:
That works. We should be mindful though that the limit imposed on the denominator means that the framerate will "only" be accurate up to the first 4 significant digits.

> This is the code that i use my (new) av wrapper:

I think this would be useful for ImageIO as well. Could you contribute it together with a note in `init_video_stream` that mentions the limitation of a 16bit denominator (and the resulting 4 digits of precision)?

PR #939, opened 2023-02-02
FEAT: optimize pyav

FirefoxMetzger:
Closes: https://github.com/imageio/imageio/issues/926, https://github.com/imageio/imageio/issues/929, https://github.com/imageio/imageio/issues/932, https://github.com/imageio/imageio/issues/933

This PR is a collection of optimizations and tweaks to improve the performance of our pyav plugin. It adds the following improvements:

- It removes an explicit (and redundant) colorspace conversion that ffmpeg can handle more efficiently internally (#926)
- It allows floats to be used for `fps`, which allows non-integer framerates (#929).
- It checks the container's `variable_fps` flag when reading metadata to automatically set `constant_framerate` if not manually specified
- It preallocates the output array when bulk reading (`index=...`) if the output shape is known in advance. (#932)
- It optimizes sequential calls to `read` so that they reuse the current decoder state if possible. (#933)
- It exposes the frame type as metadata
- It allows setting `max_keyframe_interval` and `force_keyframes` (open/closed GOP) when writing to control intra frame behavior.

### DIFF
--- a/imageio/plugins/pyav.py
+++ b/imageio/plugins/pyav.py
@@ -761,7 +761,7 @@ class PyAVPlugin(PluginV3):
         self,
         codec: str,
         *,
-        fps: int = 24,
+        fps: float = 24,
         pixel_format: str = None,
     ) -> None:
         """Initialize a new video stream.
@@ -773,7 +773,7 @@ class PyAVPlugin(PluginV3):
         ----------
         codec : str
             The codec to use, e.g. ``"x264"`` or ``"vp9"``.
-        fps : str
+        fps : float
             The desired framerate of the video stream (frames per second).
         pixel_format : str
             The pixel format to use while encoding frames. If None (default) use
@@ -782,7 +782,7 @@ class PyAVPlugin(PluginV3):
         """
 
         stream = self._container.add_stream(codec, fps)
-        stream.time_base = Fraction(1, int(fps))
+        stream.time_base = Fraction(1/fps).limit_denominator(int(2**16-1))
         if pixel_format is not None:
             stream.pix_fmt = pixel_format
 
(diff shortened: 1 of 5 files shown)


FirefoxMetzger:
@hmaarrfk This PR implements a lot of your suggestions. Are you interested in reviewing it?

imageio/plugins/pyav.py
```diff
@@ -1047,30 +1080,63 @@ def _unpack_frame(self, frame: av.VideoFrame, *, format: str = None) -> np.ndarr
...
+        self._next_keyframe_idx = idx + self._next_idx
+
     def _seek(self, index, *, constant_framerate: bool = True) -> Generator:
         """Seeks to the frame at the given index."""
 
-        # this may be made faster for formats that have some kind
-        # of keyframe table in their header data.
-        if not constant_framerate:
+        if index == self._next_idx:
+            return  # fast path :)
```
hmaarrfk, line 1102: nice way to do it.

hmaarrfk:
I've also found that OpenCV is quite aggressive on the number of threads it uses. Which is why it can seem like opencv is fast.

When i use the same number of threads between OpenCV and pyav, i can close the gap in performance quite significantly.

PyAV still requires a "copy" of the data out of the plantes RGB format, we can make those PRs to PyAV. But I want their help resolving https://github.com/PyAV-Org/PyAV/issues/1045 which is quite important in future workflows i have in mind.

FirefoxMetzger:
almost there :) I have to fix a CI issue (GH actions now uses MacOS 12 which doesn't have pre-built wheels for OpenCV, so we need to exclude that plugin on this MacOS version) and then we can merge.

almarklein:
🚀

hmaarrfk:
Great job! Thanks!

PR #941, opened 2023-02-05
BUG: format hint test doesn't need to establish a connection

FirefoxMetzger:
I recently learned that one of our tests tries to establish a connection to `sample.com` to make sure that the `format_hint` kwarg of our unit-test suite works. This is a bad idea for many reasons, so this PR makes it use a test image instead.


FirefoxMetzger:
I'll also use this as a chance to update to the latest black version :)

Issue #940, opened 2023-02-05
Add audio support for pyav


Azleur:
This is a proposal to add new methods to the `pyav` plugin API in order to expose pyav's audio read/write capabilities. Reading could look something like this:
```python
with iio.imopen(...) as file:
    audio = file.read_audio(...)
    for audio_frame in file.read_audio():
        ... 
```
And writing could look something like this:
```python
with iio.imopen(...) as file:
    file.init_audio_stream(...)
    for chunk in audio_frames:
        file.write_audio_frame(chunk)
```
I can contribute the code changes for this addition.

FirefoxMetzger:
Looks good from my side. Feel free to get started with a PR and let me know if you need any help or if I should review anything.

PR #962, opened 2023-03-11
Add pyav audio support

Azleur:
See issue #940

### DIFF
--- a/imageio/plugins/pyav.py
+++ b/imageio/plugins/pyav.py
@@ -295,6 +295,7 @@ class PyAVPlugin(PluginV3):
         self._container = None
         self._video_stream = None
         self._video_filter = None
+        self._audio_stream = None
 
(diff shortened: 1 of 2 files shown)


FirefoxMetzger (changes requested):
Nice work @Azleur 🚀 

I think this is pretty much complete and only needs some minor tweaking to make the API consistent with the rest of ImageIO.

There is one feature I was thinking about while reviewing that I think makes a lot of sense, so I wonder what you think about it: In my mind, the default use-case for reading audio is to read it alongside some video frames/stream and to then pass both the frame and the corresponding audio into some ML pipeline or deep learning model. Currently, there is no easy way for a user to figure out which segment of audio corresponds to a video frame stored at a given index. Do you think it would be useful to add code that allows reading/iterating over the audio in batches that correspond to the start and duration at which individual frames are shown?

PR #963, opened 2023-03-18
Update SPE plugin for v3 API

lschr:


lschr:
@FirefoxMetzger Could you have a look at this? I think the No-Internet Install check fails due to unrelated problems (it has worked before).

FirefoxMetzger:
Yes, the test fails, because it exceeds GH's rate limit for pulling images. I have seen it fail sporadically due to a network timeout (rare), but not yet due to rate limiting issues. I'll re-run it and it should dissapear.

lschr:
Thanks! Do you have suggestions concerning the failing codecov checks? Most of the complaints concern reading of metadata not present in the test file. All I can think of would be to patch the header of the test file using a hex editor to test the additional code paths.

FirefoxMetzger:
I will have to look into the code itself (the current plan is to do it this afternoon) before I can answer your question on coverage @Ischr .

If I can find a way to create a test file I will propose it, otherwise we can also continue without having the plugin covered 100%, since the old plugin wasn't fully covered either. I prefer full coverage though, so I want to make that option work if possible.

FirefoxMetzger:
Sorry @lschr for the delay on reviewing this PR. I haven't forgotten, but want to have https://github.com/imageio/imageio/pull/824 merged before starting work on this. The work here replaces the SPE plugin with a v3 version, which is amazing 🤩 but will make SPE unavailable in v2 until we merge the aforementioned PR. Ideally, I would like this to be backwards compatible, hence the delay.

imageio/plugins/spe.py
```diff
@@ -575,172 +433,426 @@ def _open(self, char_encoding="latin1", check_filesize=True, sdt_meta=True):
...
+        ):
+            m[k] = bool(m[k])
+
+        # Extract SDT-control metadata if desired
+        if sdt_control:
+            SDTControlSpec.extract_metadata(m, char_encoding)
+
+        return m
+
+    def _metadata_post_v3(self) -> Dict[str, Any]:
```
FirefoxMetzger, line 755: Same here. You just moved it right?

imageio/plugins/spe.py
```diff
@@ -575,172 +433,426 @@ def _open(self, char_encoding="latin1", check_filesize=True, sdt_meta=True):
...
+        dict with key `"__xml"`, whose value is the XML metadata
+        """
+
+        info = self._parse_header(Spec.basic, "latin1")
+        self._file.seek(info["xml_footer_offset"])
+        xml = self._file.read()
+        return {"__xml": xml}
+
+    def properties(self, index: int = ...) -> ImageProperties:
+        """Standardized ndimage metadata
```
FirefoxMetzger, line 769:
```suggestion
        """Standardized ndimage metadata.
```

imageio/plugins/spe.py
```diff
@@ -575,172 +433,426 @@ def _open(self, char_encoding="latin1", check_filesize=True, sdt_meta=True):
...
+        return (self.read(index=i) for i in range(self._len))
+
+    def metadata(
+        self,
+        index: int = ...,
+        exclude_applied: bool = True,
+        char_encoding: str = "latin1",
+        sdt_control: bool = True,
+    ) -> Dict[str, Any]:
+        """Get metadata
```
FirefoxMetzger, line 493:
```suggestion
        """SPE specific metadata.
```

imageio/plugins/spe.py
```diff
@@ -575,172 +433,426 @@ def _open(self, char_encoding="latin1", check_filesize=True, sdt_meta=True):
...
+    def properties(self, index: int = ...) -> ImageProperties:
+        """Standardized ndimage metadata
+
+        Parameters
+        ----------
+        index
+            If the index is an integer, select the index-th frame and return
+            its properties. If index is an ellipsis (...), return the
+            properties of all frames in the file stacked along a new batch
+            dimension.
```
FirefoxMetzger, line 777:
```suggestion
        index : int
            If the index is an integer, select the index-th frame and return
            its properties. If index is an Ellipsis (...), return the
            properties of all frames in the file stacked along a new batch
            dimension.
```

imageio/plugins/spe.py
```diff
@@ -575,172 +433,426 @@ def _open(self, char_encoding="latin1", check_filesize=True, sdt_meta=True):
...
+
+        Returns
+        -------
+        properties
```
FirefoxMetzger, line 781:
```suggestion
        properties : ImageProperties
```

imageio/plugins/spe.py
```diff
@@ -575,172 +433,426 @@ def _open(self, char_encoding="latin1", check_filesize=True, sdt_meta=True):
...
+        if index is Ellipsis:
+            return ImageProperties(
+                shape=(self._len, *self._shape),
+                dtype=self._dtype,
+                n_images=self._len,
+                is_batch=True,
+            )
+        return ImageProperties(shape=self._shape, dtype=self._dtype, is_batch=False)
+
+    def _parse_header(
```
FirefoxMetzger, line 794: This function, too, just appears to have moved. Could you confirm?

tests/test_spe.py
```diff
@@ -1,51 +1,53 @@
...
 
 
-def test_spe_reading(test_images):
-    fname = test_images / "test_000_.SPE"
+def test_imopen(test_images, tmp_path):
+    with iio.imopen(test_images / "test_000_.SPE", "r") as img:
+        assert isinstance(img, spe.SpePlugin)
+
+    with iio.imopen(tmp_path / "test.spe", "w") as img:
+        assert not isinstance(img, spe.SpePlugin)
```
FirefoxMetzger, line 22:
```suggestion
    with pytest.raises(OSError):
        iio.imopen(tmp_path / "test.spe", "w", plugin="SPE")
```

Instead of testing that another plugin is found (which could fail) we should instead test that SPE declines to read SPE files. 

This is also in preparation for a feature on my wishlist, which is that plugin tests should run isolated from other plugins, i.e., if we run the tests for SPE then all other plugins should be disabled to avoid accidentally testing a different plugin. (The tested plugin might fail and ImageIO has logic in place to (sometimes) fall back on a different one that is known to be able to read this file type if possible.)

FirefoxMetzger (changes requested):
Fantastic work @lschr I really like what you did with the SPE plugin in this PR. Most of my comments are related to docstrings (added quite a few periods `.` in various places as well as some type indicators to the parameters).

There is only one bigger comment in `imread` where we could combine the case of reading `index=...` and `index=int`. Otherwise the PR is ready to roll :)

Also, sorry for taking so long to get around to reviewing this. Life has been somewhat of a roller coaster these past few weeks, but I'm happy to have found the time today to take a look at this. It's always fun to review clean and well-written code.

imageio/plugins/spe.py
```diff
@@ -533,37 +381,47 @@ def extract_metadata(meta: Mapping, char_encoding: str = "latin1"):
...
-        def _open(self, char_encoding="latin1", check_filesize=True, sdt_meta=True):
-            self._file = self.request.get_file()
-            self._char_encoding = char_encoding
+        self._file = self.request.get_file()
 
-            info = self._parse_header(Spec.basic)
+        try:
+            # Spec.basic contains no string, no need to worry about character
+            # encoding.
+            info = self._parse_header(Spec.basic, "latin1")
```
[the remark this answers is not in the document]
lschr: `Spec.basic` contains no string, hence the `char_encoding` argument is not used. We could also pass `None` (or anything else here. Could the comment just above the `_parse_header` call be phrased better?

imageio/plugins/spe.py
```diff
@@ -533,37 +381,47 @@ def extract_metadata(meta: Mapping, char_encoding: str = "latin1"):
...
-                data_end = (
-                    info["xml_footer_offset"]
-                    if info["file_header_ver"] >= 3
-                    else os.path.getsize(self.request.get_local_filename())
-                )
+                if info["file_header_ver"] >= 3:
+                    data_end = info["xml_footer_offset"]
+                else:
+                    self._file.seek(0, os.SEEK_END)
+                    data_end = self._file.tell()
```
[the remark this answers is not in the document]
lschr: If this has an impact on performance, I'd suggest `seek()`ing the start of image data.

imageio/plugins/spe.py
```diff
@@ -575,172 +433,426 @@ def _open(self, char_encoding="latin1", check_filesize=True, sdt_meta=True):
...
+        -------
+        A Numpy array of pixel values.
+        """
+
+        if index in (None, Ellipsis):
+            self._file.seek(Spec.data_start)
+            data = np.fromfile(
+                self._file,
+                dtype=self._dtype,
+                count=self._shape[0] * self._shape[1] * self._len,
```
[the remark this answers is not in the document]
lschr: As far as I know, there is no mention of multiple channels in the docs.

imageio/plugins/spe.py
```diff
@@ -575,172 +433,426 @@ def _open(self, char_encoding="latin1", check_filesize=True, sdt_meta=True):
...
+
+    def iter(self) -> Iterator[np.ndarray]:
+        """Iterate over the frames in the file
+
+        Yields
+        ------
+        A Numpy array of pixel values.
+        """
+
+        return (self.read(index=i) for i in range(self._len))
```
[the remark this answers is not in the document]
lschr: I honestly have no idea, but I never considered seeking anything to worry about compared to the overall slowness of Python.

imageio/plugins/spe.py
lschr: There were some changes. First, the old plugin read the metadata upon loading and saved it as `self._meta`. I made the code a little more compact. There is no more `"frame_shape"` metadata field since it is redundant.

imageio/plugins/spe.py
lschr: This is new. The old plugin had no `properties()`.

imageio/plugins/spe.py
```diff
@@ -533,37 +381,47 @@ def extract_metadata(meta: Mapping, char_encoding: str = "latin1"):
         meta.pop("exposure_sec")
 
 
-class SpeFormat(Format):
-    """See :mod:`imageio.plugins.spe`"""
+class SpePlugin(PluginV3):
+    def __init__(self, request: Request, check_filesize: bool = True) -> None:
```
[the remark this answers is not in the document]
lschr: This is not so easy as I initially thought.

What you suggest would mean to set the default arguments of `metadata()` to `None` in order to be able to detect whether the user changed them and then define the actual defaults via `if arg is None` in the method, hiding them from the user/documentation. I don't think that this is worth just to introduce anyhow deprecated parameters.

What can be done easily, though, is 

1. value stored in `plugin.kwarg` (user input to `plugin.__init__`)
1. user input to `plugin.metadata(...)`
1. default of `plugin.metadata(...)`

This would preserve backward compatibility and explicit default arguments. Nobody should be passing arguments to `__init__()` and `metadata()` in the same code anyhow – either they still use v2 or already v3.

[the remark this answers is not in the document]
lschr: I implemented the latter in 0ff1db8e13329ad868fd75eecf309cd5e98d22a2.

imageio/plugins/spe.py
```diff
@@ -575,172 +433,426 @@ def _open(self, char_encoding="latin1", check_filesize=True, sdt_meta=True):
...
+            If `True`, try to decode special metadata written by the
+            SDT-control software.
+
+        Returns
+        -------
+        dict mapping metadata names to values. For SPE v3 files, this is the
+        metadata stored in XML format. For SPE v2 files, metadata fields
+        include
+
+        ROIs : list of dict
```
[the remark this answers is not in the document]
lschr: Nevermind, now it worked.

imageio/plugins/spe.py
lschr: Sorry, you probably meant `_parse_header()`. The only change is that `char_encoding` is now a parameter. Previously, `self._char_encoding`  was used.

imageio/plugins/spe.py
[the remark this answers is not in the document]
FirefoxMetzger: > If this has an impact on performance, I'd suggest seek()ing the start of image data.

Yes, that would be the best place to seek to, especially if it is different from the end of the header.

The reason it will improve performance is that most OSes will opportunistically start buffering/reading a file to improve subsequential calls to read. Setting the pointer to the start of the image data before giving control back to the user allows the OS to do this while the user performs other tasks. It will likely only be a small boost, but since it is free why not :)

imageio/plugins/spe.py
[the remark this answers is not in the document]
FirefoxMetzger: That is also true.

imageio/plugins/spe.py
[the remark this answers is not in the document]
FirefoxMetzger: Theoretically, users of v3 can do something the following (which hopefully never happens, because it is indeed not super useful):

```python
with iio.imopen("fancy.spe", plugin="SPE", char_encoding="utf-8") as file:
    meta = file.metadata(char_encoding="iso8859-1")
```

Nevertheless, with this syntax, I would have expected ImageIO to use iso 8859-1 for the data. However, given that this is a real oddball, I think it will be okay to have the constructor's `char_encoding` take preference if that makes implementation easier.

I will take a look when I do the next round of code review - currently schedules for the end of this week.

imageio/plugins/spe.py
```diff
@@ -575,172 +433,426 @@ def _open(self, char_encoding="latin1", check_filesize=True, sdt_meta=True):
...
+        modulation_script : str
+            (only for files created by SDT-control)
+            Laser modulation script. Replaces the "spare_4" key.
+        """
+
+        if self._file_header_ver < 3:
+            return self._metadata_pre_v3(char_encoding, sdt_control)
+        return self._metadata_post_v3()
+
+    def _metadata_pre_v3(self, char_encoding: str, sdt_control: bool) -> Dict[str, Any]:
```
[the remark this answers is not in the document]
FirefoxMetzger: Yes, the list of metadata fields is useful :) This comment was meant for the docstring of `_metadata_pre_v3` (which is a private method).

imageio/plugins/spe.py
FirefoxMetzger: Yes, I was referring to `char_encoding`. Somehow GH decided to refer to the past 4 lines instead of just referencing the definition that I selected ._. sorry for the confusion.

imageio/plugins/spe.py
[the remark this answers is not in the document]
lschr: I see that this is possible, but I would expect

- v2 API users / legacy code not to pass any arguments to `metadata()` since this did not exist in v2
- v3 API users to properly call the new API, omitting any deprecated parameters existing only for v2 compatibility

I'd say, anyone using the code above is doing risky stuff intentionally and has to deal with the fallout themselves ;-)

imageio/plugins/spe.py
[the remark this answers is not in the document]
lschr: This is now implemented.

imageio/plugins/spe.py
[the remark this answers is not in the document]
lschr: Ah. I this is my first time using the code review function. Now I understand that the comment belongs to the bottom line ;-)
So, what I wrote about `_metadata_post_v3` actually belongs here.

imageio/plugins/spe.py
lschr: My last comment is actually about `_metadata_pre_v3`, but except for the part about `"frame_shape"` it also applies to `_metadata_post_v3`

imageio/plugins/spe.py
```diff
@@ -575,172 +457,437 @@ def _open(self, char_encoding="latin1", check_filesize=True, sdt_meta=True):
...
+        if index is Ellipsis:
+            read_offset = Spec.data_start
+            count = self._shape[0] * self._shape[1] * self._len
+            out_shape = (self._len, *self._shape)
+        elif index < 0:
+            raise IndexError(f"Index `{index}` is smaller than 0.")
+        elif index >= self._len:
+            raise IndexError(
+                f"Index `{index}` exceeds the number of frames stored in this file (`{self._len}`)."
+            )
```
FirefoxMetzger, line 488: Nice Error handling :)

imageio/plugins/spe.py
```diff
@@ -533,37 +382,70 @@ def extract_metadata(meta: Mapping, char_encoding: str = "latin1"):
...
-        return request.extension in self.extensions
+        Parameters
+        ----------
+        request : Request
+            A request object representing the resource to be operated on.
+        check_filesize : bool
+            If True, compute the number of frames from the filesize, compare it
+            to the frame count in the file header, and raise a warning if the
+            counts don't match. (Certain software may create files with
+            incorrect counts in the header.)
```
FirefoxMetzger, line 403:
```suggestion
        char_encoding : str
            Deprecated. Exists for backwards compatibility; use ``char_encoding`` of
            ``metadata`` instead.
        sdt_control
            Deprecated. Exists for backwards compatibility; use ``sdt_control`` of
            ``metadata`` instead.
```

Same here, let's add a note to not use these parameters.

imageio/plugins/spe.py
```diff
@@ -3,206 +3,54 @@
...
-    (only for files created by SDT-control)
-    User-entered comment. This replaces the "comments" field.
-datetime : datetime.datetime
-    (only for files created by SDT-control)
-    Combines the "date" and "time_local" keys. The latter two plus
-    "time_utc" are removed.
-modulation_script : str
-    (only for files created by SDT-control)
-    Laser modulation script. Replaces the "spare_4" key.
+
```
FirefoxMetzger, line 16:
```suggestion
char_encoding : str
    Deprecated. Exists for backwards compatibility; use ``char_encoding`` of
    ``metadata`` instead.
sdt_control
    Deprecated. Exists for backwards compatibility; use ``sdt_control`` of
    ``metadata`` instead.

```

Since we added new arguments to the constructor, we should document them. Even if it is only to say that they exist for compatibility and should not be used.

FirefoxMetzger (changes requested):
Nice work @lschr ! One final change in the docstrings and then we can merge 🚀

FirefoxMetzger (approved):

FirefoxMetzger:
LGTM! Awesome work @Ischr ; thanks for upgrading our SPE plugin 🚀

PR #964, opened 2023-03-19
DEV: remove pypy 3.7 testing and improve CI/CD scripts

FirefoxMetzger:
This PR cleans up our CI and CD scripts. It removes pypy 3.7 from the test matrix (no longer downloadable) and combines the config that both the CI and the CD use when running the test suite and building the docs. This way, we can be more confident that what we test on the PR is the same as what is tested when releasing.

### DIFF
--- a/.github/workflows/build_docs.yml
+++ b/.github/workflows/build_docs.yml
@@ -1,0 +1,19 @@
+name: Documentation
+
+on:
+  workflow_call:
+
+jobs:
+  docs-GH:
+    name: Build Docs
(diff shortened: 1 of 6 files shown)


FirefoxMetzger:
I just checked and, as it turns out, pypy 3.7 is no longer downloadable. Instead, you can currently choose between 3.8 or 3.9. I think we should follow suite and drop explicit testing for 3.7, which seems to have been deprecated in pypy v7.3.10.

I will update this PR to do this and use the chance to clean up our CI scripts a bit.

Issue #915, opened 2022-11-30
Number of images via v3 API


lschr:
Is there a way to get the number of images in a file via the v3 API?

FirefoxMetzger:
The first thing that comes to mind is something like

```python
import imageio.v3 as iio

n_images = iio.improps(ImageResource, index=...).shape[0] 
```

This works if you have multiple images of the same shape, e.g.:

```python
iio.improps("imageio:chelsea.png", index=...).shape[0]  # 1
iio.improps("imageio:newtonscradle.gif", index=...).shape[0]  # 36
iio.improps("imageio:cockatoo.mp4", index=..., plugin="pyav").shape[0]   # 280
```

However, if you have multiple images with different shapes (especially inside a TIFF) then your options are more limited. Here, you can either inspect the format-specific metadata (via `immeta`), or you can count via

```python
n_images = len([x for x in iio.imiter(ImageResource)])
```

Neither approach is ideal though, because immeta doesn't report the number of contained images for all formats and counting can be expensive as it decodes pixel data.

I think it could be useful to add a `n_images` field to `ImageProperties` to create a uniform way of checking this. What do you think?

---

Along those lines, it would probably also be useful to have a `format` and `backend` field that gives you a programmatic way of inspecting the file format (e.g. if you are reading bytes) and the backend (e.g., to check which one was automatically selected) ...

lschr:
> Here, you can either inspect the format-specific metadata (via `immeta`), or you can count via
> 
> ```python
> n_images = len([x for x in iio.imiter(ImageResource)])
> ```

This would mean reading potentially gigabytes of data just to get the number of frames, so I think it is not really an option.

> I think it could be useful to add a `n_images` field to `ImageProperties` to create a uniform way of checking this. What do you think?

I think this is a great idea. For TIFF, it would be possible to return either the number of series (e.g. if `index=..., page=None`) or the number of pages (e.g. `index=..., page=...` or `index=None, page=...`).

lschr:
I started implementing this, but then realized that it would not work so well. Returning the number of images only makes sense when querying file-global properties (i.e. `index=None`), but not when querying metadata for a single image.

What would you think about adding a new method, e.g. `frame_count()`, to the `PluginV3` class, which is to be implemented by subclasses?

FirefoxMetzger:
> I started implementing this, but then realized that it would not work so well. Returning the number of images only makes sense when querying file-global properties (i.e. index=None), but not when querying metadata for a single image.

[...]

PR #966, opened 2023-03-23
FEAT: Add `n_images` to `ImageProperties`

lschr:
This exposes the number of frames/images in a file via `improps()`.

See also #915.

### DIFF
--- a/imageio/core/legacy_plugin_wrapper.py
+++ b/imageio/core/legacy_plugin_wrapper.py
@@ -305,7 +305,6 @@ class LegacyPlugin(PluginV3):
         return ImageProperties(
             shape=image.shape,
             dtype=image.dtype,
-            n_images=None,
             is_batch=False,
         )
 
--- a/imageio/core/legacy_plugin_wrapper.py
+++ b/imageio/core/legacy_plugin_wrapper.py
@@ -296,7 +296,8 @@ class LegacyPlugin(PluginV3):
         return ImageProperties(
             shape=image.shape,
             dtype=image.dtype,
-            is_batch=True if index is None else False,
+            n_images=image.shape[0] if index is Ellipsis else None,
+            is_batch=index is Ellipsis,
         )
 
     def get_meta(self, *, index=None):
--- a/imageio/core/legacy_plugin_wrapper.py
+++ b/imageio/core/legacy_plugin_wrapper.py
@@ -291,13 +291,22 @@ class LegacyPlugin(PluginV3):
             index = _legacy_default_index(self._format)
 
         # for backwards compatibility ... actually reads pixel data :(
-        image = self.read(index=index)
+        if index is Ellipsis:
+            image = self.read(index=0)
+            n_images = self.legacy_get_reader().get_length()
+            return ImageProperties(
+                shape=(n_images, *image.shape),
+                dtype=image.dtype,
+                n_images=n_images,
+                is_batch=True,
+            )
 
+        image = self.read(index=index)
         return ImageProperties(
             shape=image.shape,
             dtype=image.dtype,
-            n_images=image.shape[0] if index is Ellipsis else None,
-            is_batch=index is Ellipsis,
+            n_images=None,
+            is_batch=False,
         )
 
     def get_meta(self, *, index=None):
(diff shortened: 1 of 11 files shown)


imageio/core/v3_plugin_api.py
```diff
@@ -35,6 +37,7 @@ class ImageProperties:
 
     shape: Tuple[int, ...]
     dtype: np.dtype
+    n_images: Optional[int]
```
FirefoxMetzger, line 40:
```suggestion
    n_images: Optional[int] = None
```

Saves us from setting `None` when reading single images.

imageio/core/legacy_plugin_wrapper.py
```diff
@@ -291,12 +291,22 @@ def properties(self, index=None):
...
+                n_images=n_images,
+                is_batch=True,
+            )
 
+        image = self.read(index=index)
         return ImageProperties(
             shape=image.shape,
             dtype=image.dtype,
-            is_batch=True if index is None else False,
+            n_images=None,
```
FirefoxMetzger, line 308:
```suggestion
```

We can make `None` the default in the definition.

imageio/core/v3_plugin_api.py
```diff
@@ -20,6 +20,8 @@ class ImageProperties:
         The shape of the loaded ndimage.
     dtype : np.dtype
         The dtype of the loaded ndimage.
+    n_images : int or None
```
FirefoxMetzger, line 23:
```suggestion
    n_images : int
```

We don't document `None` if it is the default value.

imageio/plugins/tifffile_v3.py
```diff
@@ -368,13 +371,24 @@ def properties(self, *, index: int = None, page: int = None) -> ImageProperties:
...
+                dtype=target_page.dtype,
+                n_images=n_pages,
                 is_batch=True,
                 spacing=_get_resolution(target_page)["resolution"],
             )
         else:
             props = ImageProperties(
                 shape=target_page.shape,
                 dtype=target_page.dtype,
+                n_images=None,
```
FirefoxMetzger, line 391:
```suggestion
```

imageio/plugins/opencv.py
```diff
@@ -270,13 +270,28 @@ def properties(
...
+                n_images=n_images,
+                is_batch=True,
+            )
 
+        img = self.read(index=index, flags=flags, colorspace=colorspace)
         return ImageProperties(
-            shape=img.shape,
-            dtype=img.dtype,
-            is_batch=(index is ...),
+            shape=img.shape, dtype=img.dtype, n_images=None, is_batch=False
```
FirefoxMetzger, line 294:
```suggestion
            shape=img.shape, dtype=img.dtype, is_batch=False
```

FirefoxMetzger (approved):
Nice work, the code is really clean and I only have a few minor comments. Thanks for the PR 🚀 

Regarding the CI failure: Is this branch in sync with the latest master? It failed because of an old issue that I've fixed a few days ago. If it is on the latest version and still fails I will have to reopen that issue.

lschr:
> Regarding the CI failure: Is this branch in sync with the latest master?

I think this was a legitimate complaint. I had forgotten to update OpenCV tests, and there was no test for calling `properties()` with integer index.

Also, I vaguely remember some inconsistencies in the documentation of `properties()`, `metadata()` and `read()` regarding `index=None` vs `index=...`. Is this material for another PR or should we fix it here? (Or maybe my memory is wrong and there are no problems.)

FirefoxMetzger:
> Also, I vaguely remember some inconsistencies in the documentation of properties(), metadata() and read() regarding index=None vs index=.... Is this material for another PR or should we fix it here? (Or maybe my memory is wrong and there are no problems.)

That could be the case. In the initial iterations of v3 we used `None` to indicate "read all ndimages or file-level metadata". However, we switched this to `...` soon after, because we needed a sentinel to indicate "no default / plugin-level default" and it made sense to have that be `None`, which is typically used for this purpose.

Let's fix any inconsistencies in a separate PR. This one is ready to go and can be part of Monday's release :)

imageio/core/v3_plugin_api.py
```diff
@@ -20,6 +20,8 @@ class ImageProperties:
         The shape of the loaded ndimage.
     dtype : np.dtype
         The dtype of the loaded ndimage.
+    n_images : int
+        Number of images in the file if ``index=...``, `None` for single images
```
FirefoxMetzger, line 24:
```suggestion
        Number of images in the file if ``index=...``, `None` for single images.
```

Issue #974, opened 2023-04-25
imageio only writes one frame for a gif


akaszynski:
The recent release of `imageio==2.28.0` no longer writes multiple frames to a gif. Here's a MWE:

```py
import imageio
import numpy as np

# Create sample frames (numpy arrays)
frame1 = np.random.randint(0, 256, (100, 100, 3), dtype=np.uint8)
frame2 = np.random.randint(0, 256, (100, 100, 3), dtype=np.uint8)
frame3 = np.random.randint(0, 256, (100, 100, 3), dtype=np.uint8)

# Set the frames in a list
frames = [frame1, frame2, frame3]

# Write the frames to an animated GIF
with imageio.get_writer('/tmp/tmp.gif', mode='I', duration=0.5) as writer:
    for frame in frames:
        writer.append_data(frame)
```
Generates:
![tmp](https://user-images.githubusercontent.com/11981631/234162570-2b520c0b-47e8-4e4f-b7de-2d939c174bfe.gif)

And it should be:
![tmp](https://user-images.githubusercontent.com/11981631/234162496-049d3607-458f-45e6-90a8-73aac9a13095.gif)

FirefoxMetzger:
Well spotted @akaszynski . It is an interesting issue, because - as it turns out - in the 2+ years that we now have the new v3 plugin nobody seems to have tried to use it to incrementally write a GIF (at least from what we know). Neither the old nor the new plugin have a unit test for this scenario either, and as a result, we never noticed this regression until now.

I fixed it in #978 and will add a regression test. Should roll out next week or the week after :)

wrench1997:
I use ffmpeg to solve this problem.

PR #979, opened 2023-04-27
BUG: Forward get_reader/get_writer kwargs to read/write

FirefoxMetzger:
This PR fixes a bug where kwargs were silently ignored when using `get_reader` or `get_writer` with a v3 plugin. This PR fixes this (adds the required plumbing) and adds a regression test.

It will wait for https://github.com/imageio/imageio/pull/978 to be merged, because the test depends on it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 <!--next-version-placeholder-->
 
+## v2.28.0 (2023-04-24)
+### Feature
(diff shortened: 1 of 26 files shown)


FirefoxMetzger:
Referencing #974 too, because I discovered this will working on the regression reported there.

miriam-groeneveld:
This causes a deprecation warning in the following code, will v2 be deprecated soon?
```
from imageio import get_reader         
with open("some_file", "rb") as f:   
  with get_reader(f, mode="i", pilmode="L") as r:
    return r.get_data(0)```

PR #983, opened 2023-04-29
Fix pillow image EXIF orientation transforms

assert0:
Demo of fix:

```
import os
import imageio

# wget https://www.galloway.me.uk/media/other/EXIF_Orientation_Samples.zip
# unzip EXIF_Orientation_Samples.zip

d = "EXIF_Orientation_Samples"
for f in os.listdir(d):
    path = f"{d}/{f}"
    data = imageio.imread(path, rotate=True)
    imageio.imwrite(f"{d}/{f}.png", data)
```

### DIFF
--- a/imageio/plugins/pillow.py
+++ b/imageio/plugins/pillow.py
@@ -51,9 +51,9 @@ def _exif_orientation_transform(orientation: int, mode: str) -> Callable:
         3: lambda x: np.rot90(x, k=2),
         4: lambda x: np.flip(x, axis=axis - 1),
         5: lambda x: np.flip(np.rot90(x, k=3), axis=axis),
-        6: lambda x: np.rot90(x, k=1),
+        6: lambda x: np.rot90(x, k=3),
         7: lambda x: np.flip(np.rot90(x, k=1), axis=axis),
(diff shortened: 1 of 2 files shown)


FirefoxMetzger:
Thanks for the report @assert0 and for proposing the fix. It looks like we already have a test that asserts that ExIF rotation happens the other way around so the question is which version is "correct".

I am inclined to believe that your fix is correct and that our tests are wrong, but I will have to dig up the specs to double-check this. The modes mean 90° and 270° rotation, but a quick google is inconclusive if that means 90° *clock-wise* or *counter clock-wise*, so it's down to the specs to determine the correct behavior.

assert0:
Updated pull request to include updated pytests to verify that imageio's transform produces the same orientation result as Pillow's exif_transpose function.

FirefoxMetzger (approved):
Finally got a chance to dig up the EXIF 2.2 specs ([link](http://web.archive.org/web/20131018091152/http://exif.org/Exif2-2.PDF)) to check in which directions rotations should be applied.

You are correct @assert0 that rotation should be CW not CCW. The PR looks good from my side, so I will go ahead and hit merge.

Thanks for the contribution 🚀

PR #984, opened 2023-04-29
BUG: Avoid exception when n_frames is not set by Pillow

### LINKED ISSUE #958: AttributeError: n_frames
```
>>> from imageio.v3 import improps
>>> improps(r"test_libuemf_ref.emf")
  File "<stdin>", line 1, in <module>
  File "C:\Python311\Lib\site-packages\imageio\v3.py", line 198, in improps
    properties = img_file.properties(**call_kwargs)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Python311\Lib\site-packages\imageio\plugins\pillow.py", line 455, in properties
    n_frames: int = self._image.n_frames
                    ^^^^^^^^^^^^^^^^^^^^
  File "C:\Python311\Lib\site-packages\PIL\Image.py", line 529, in __getattr__
    raise AttributeError(name)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: n_frames
```

 imageio 2.26.0, Pillow  9.4.0

(The test file can be downloaded e.g. here: https://github.com/mathog/libUEMF/blob/master/test_libuemf_ref.emf)

radarhere:
Resolves #958

Not all Pillow image formats have `n_frames` defined. See https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.n_frames
> The number of frames in this image.
>
> This attribute is only defined by image plugins that support animated images. Plugins may leave this attribute undefined if they don’t support loading animated images, even if the given format supports animated images.
>
> Given that this attribute is not present for all images use `getattr(image, "n_frames", 1)` to check the number of frames that Pillow is aware of in an image regardless of its format.

### DIFF
--- a/imageio/plugins/pillow.py
+++ b/imageio/plugins/pillow.py
@@ -538,8 +538,9 @@ class PillowPlugin(PluginV3):
         height: int = self._image.height
         shape: Tuple[int, ...] = (height, width)
 
-        n_frames: int = self._image.n_frames
+        n_frames: Optional[int] = None
         if index is ...:
+            n_frames = getattr(self._image, "n_frames", 1)
             shape = (n_frames, *shape)
 
         dummy = np.asarray(Image.new(mode, (1, 1)))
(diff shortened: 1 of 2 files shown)


FirefoxMetzger (approved):
Thanks for the fix @radarhere 

LGTM once CI is done doing its thing.

PR #986, opened 2023-04-30
DEV: refactor invoke in no-internet CI

FirefoxMetzger:
Invoke's latest release broke our CI that simulates running ImageIO without internet. It is the only place where we still rely on invoke so I decided to refactor. Now we call pytest directly (still one line) and have one element less in CI that can break.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,7 +63,7 @@ jobs:
             rm -rf ./imageio ./build ./egg-info
       - name: Unit tests
         run: |
-            pytest -v -m not needs_internet --cov imageio --cov-config .coveragerc --cov-report
+            pytest -v -m not needs_internet
         env:
           IMAGEIO_NO_INTERNET: 1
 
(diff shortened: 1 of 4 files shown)


FirefoxMetzger:
@almarklein You are the only person I know that has used our invoke scripts, so I would appreciate your opinion on this one.

The latest release of invoke (from Saturday 2023-04-28) does not work with our scripts anymore and has broken our CI. As a result, I refactored the one CI task that still needed them from `invoke test --installed` to `pytest -v -m "not needs_internet"` which removes invoke as a test dependency.

Since we no longer depend on invoke after this PR I am thinking of deprecating these tasks altogether, because I doubt that it is actively in use and it means one thing less to maintain. How do you think about the invoke scripts/tasks?

FirefoxMetzger:
I will merge this to not block tomorrow's automatic release with a failing CI; however, I am still interested in thoughts on deprecating `invoke` (which we can easily undo by removing the warning if there is merit to keeping it around).

almarklein:
Yeah, lets remove it! 👍

PR #994, opened 2023-05-24
Fix mode issue

### LINKED ISSUE #991: BUG Using get_reader with pilmode leads to DeprecationWarnings
We were using the following code to load an image:

```
from imageio import get_reader         
with open("some_file", "rb") as f:   
  with get_reader(f, mode="i", as_gray=True) as r:
    return r.get_data(0)
```

From version 2.28.1 (or maybe earlier) this raises a TypeError: "The keyword `as_gray` is no longer supported. Use `mode='L'` instead."
As we are using the get_reader function, the only way to set this mode is using `pilmode`:

```
from imageio import get_reader         
with open("some_file", "rb") as f:   
  with get_reader(f, mode="i", pilmode="L") as r:
    return r.get_data(0)
```

This raises a Depracation Warning: `pilmode` is deprecated. Use `mode` instead.

Besides the DeprecationWarning, it also seems like the suggested alternative mode is not correct, or at least, does not result in the same data. 
In version 2.22.4:
`as_gray` returns data with type `float`. 
Mode `L` returns data with type `uint`. 
Mode `F` returns data with type `float` (From the `as_gray` [docs](https://imageio.readthedocs.io/en/stable/_autosummary/imageio.plugins.pillow_legacy.html) it seems it should be mode `F`)

miriam-groeneveld:
Closes #991 
Adds pilmode to the kwargs, therby avoiding the deprecation warning

### DIFF
--- a/imageio/v2.py
+++ b/imageio/v2.py
@@ -181,7 +181,7 @@ class LegacyWriter:
 
         if (
             type(self.instance).__name__ == "PillowPlugin"
-            and kwargs.get("pilmode") is not None
+            and "pilmode" in kwargs
         ):
             kwargs["mode"] = kwargs["pilmode"]
             del kwargs["pilmode"]
--- a/imageio/v2.py
+++ b/imageio/v2.py
@@ -179,10 +179,7 @@ class LegacyWriter:
         self.last_index = 0
         self.closed = False
 
-        if (
-            type(self.instance).__name__ == "PillowPlugin"
-            and "pilmode" in kwargs
-        ):
+        if type(self.instance).__name__ == "PillowPlugin" and "pilmode" in kwargs:
             kwargs["mode"] = kwargs["pilmode"]
             del kwargs["pilmode"]
 
--- a/imageio/v2.py
+++ b/imageio/v2.py
@@ -112,6 +112,13 @@ class LegacyReader:
         self.last_index = 0
         self.closed = False
 
+        if (
+            type(self.instance).__name__ == "PillowPlugin"
+            and kwargs.get("pilmode") is not None
+        ):
+            kwargs["mode"] = kwargs["pilmode"]
+            del kwargs["pilmode"]
+
         self.read_args = kwargs
 
     def close(self):
@@ -171,6 +178,13 @@ class LegacyWriter:
         self.instance = plugin_instance
         self.last_index = 0
         self.closed = False
+
+        if (
+            type(self.instance).__name__ == "PillowPlugin"
+            and kwargs.get("pilmode") is not None
+        ):
+            kwargs["mode"] = kwargs["pilmode"]
+            del kwargs["pilmode"]
 
         self.write_args = kwargs
 
@@ -250,7 +264,7 @@ def is_volume(ndimage):
 # Base functions that return a reader/writer
 
 
-def get_reader(uri, format=None, mode="?", pilmode=None, **kwargs):
+def get_reader(uri, format=None, mode="?", **kwargs):
     """get_reader(uri, format=None, mode='?', **kwargs)
 
     Returns a :class:`.Reader` object which can be used to read data
@@ -268,8 +282,6 @@ def get_reader(uri, format=None, mode="?", pilmode=None, **kwargs):
         Used to give the reader a hint on what the user expects (default "?"):
         "i" for an image, "I" for multiple images, "v" for a volume,
         "V" for multiple volumes, "?" for don't care.
-    pilmode : str
-        Used to provide mode parameter that need to be passed to the reader
     kwargs : ...
         Further keyword arguments are passed to the reader. See :func:`.help`
         to see what arguments are available for a particular format.
@@ -279,9 +291,6 @@ def get_reader(uri, format=None, mode="?", pilmode=None, **kwargs):
     imopen_args["legacy_mode"] = True
 
     image_file = imopen(uri, "r" + mode, **imopen_args)
-
-    if pilmode is not None:
-        kwargs["mode"] = pilmode
 
     if isinstance(image_file, LegacyPlugin):
         return image_file.legacy_get_reader(**kwargs)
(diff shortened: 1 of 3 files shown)


imageio/v2.py
```diff
@@ -278,6 +280,9 @@ def get_reader(uri, format=None, mode="?", **kwargs):
 
     image_file = imopen(uri, "r" + mode, **imopen_args)
 
+    if pilmode is not None:
+        kwargs["mode"] = pilmode
+
```
FirefoxMetzger, line 285: Nice thinking 💯 This gets us 90% there but will, unfortunately, prevent using `pilmode` with the old pillow plugins (e.g., `PNG-PIL`) that we are in the process of deprecating. These plugins expect `pilmode` not `mode` and if we add this piece of logic here things will break.

Could you move it to the `LegacyReader` class constructor located here:

https://github.com/imageio/imageio/blob/c93536ca403863db3270990911ff438b7cf1e047/imageio/v2.py#L110-L113

`pilmode` will be passed as a `kwarg` (assuming we remove the explicit mention if it in `get_reader`), so we can check for it's existence via `kwargs["pilmode"]`.

We should also only change this kwarg if users use the `PillowPlugin`. We will need a check/guard for this as well, otherwise we might create surprising behavior for other plugins. We want to avoid importing the plugin (it requires pillow to be installed which is not always true) so the check will have to be something like `type(self.instance).__name__ == "PillowPlugin"`.

Does this make sense, or should I add more detail?

imageio/v2.py
```diff
@@ -250,7 +250,7 @@ def is_volume(ndimage):
 # Base functions that return a reader/writer
 
 
-def get_reader(uri, format=None, mode="?", **kwargs):
+def get_reader(uri, format=None, mode="?", pilmode=None, **kwargs):
```
FirefoxMetzger, line 253:
```suggestion
def get_reader(uri, format=None, mode="?", **kwargs):
```

Moving the renaming to `LegacyReader` will avoid us bleeding a plugin-specific kwarg into the high-level API. I'd very much prefer that, because it will leave the API as-is :)

imageio/v2.py
```diff
@@ -268,6 +268,8 @@ def get_reader(uri, format=None, mode="?", **kwargs):
         Used to give the reader a hint on what the user expects (default "?"):
         "i" for an image, "I" for multiple images, "v" for a volume,
         "V" for multiple volumes, "?" for don't care.
+    pilmode : str
+        Used to provide mode parameter that need to be passed to the reader
```
FirefoxMetzger, line 272:
```suggestion
```

Correspondingly with my previous comment, we will not need to document `pilmode` here if we move it to `LegacyReader`.

FirefoxMetzger (changes requested):
Thanks for the PR @miriam-groeneveld 🚀 this already looks very good :)

Your solution ended up very close to what I thought of after looking at the code. The only difference is that you modify the main API (which affects other plugins), which is something I'd like to avoid if possible. I'd prefer to handle the renaming of `pilmode` to `mode` on a lower level (`LegacyReader`) which will be local to the use of `pilmode` with `PillowPlugin` without affecting other plugins. That's pretty much the main change I'd ask you to do before I merge this.

Apart from this change, I think it would be nice if we could pick up two other improvements that come out of https://github.com/imageio/imageio/issues/991 :

1. Could you update the `as_gray` deprecation notice (here: https://github.com/imageio/imageio/blob/c93536ca403863db3270990911ff438b7cf1e047/imageio/plugins/pillow.py#L203) to say that the replacement for `as_gray` should be `mode="F"` for a backward compatible result and `mode="L"` for an integer-valued result?
2. Could you do the same changes we are discussing for `LegacyReader`/`get_reader` to `LegacyWriter`? (It suffers from the same `pilmode` problem as the reader.)

If you don't want to add those changes that's perfectly fine, too. Just let me know and I'll create a separate PR for them ;)

imageio/v2.py
miriam-groeneveld: No, this makes sense, I am jsut not sure about this part:
```    
if isinstance(image_file, LegacyPlugin):
    return image_file.legacy_get_reader(**kwargs)
else:
    return LegacyReader(image_file, **kwargs)
```
Moving the handling of pilmode to the LegacyReader will only handle it for the `else` branch, should it not be handled by the legacy_get_reader too?

FirefoxMetzger: Yes that's correct and intentional; we don't need to do anything for the `LegacyPlugin` branch.

`LegacyPlugin` is a wrapper for plugins written for our old plugin system. This includes our old pillow plugin, which is the original reason for having `pilmode`. It will do the right thing without intervention from our side.

Trivia: The reason we have `pilmode` in the first place is that the old plugin system uses `mode` as a kwarg to decide if it should read or write. `mode` is also used by pillow to indicate the colorspace, and also by other backends to indicate various "modes". Obviously this is not ideal. Our new plugin uses `io_mode` to decide reading or writing, which frees up the name `mode` and allows plugins to use the name as a kwarg. Hence, our new pillow plugin can use `mode` to indicate the colorspace to use, which is what you would expect from pillow (since this is how it uses `mode`).

imageio/v2.py
```diff
@@ -172,6 +179,13 @@ def __init__(self, plugin_instance: PluginV3, **kwargs):
         self.last_index = 0
         self.closed = False
 
+        if (
+            type(self.instance).__name__ == "PillowPlugin"
+            and kwargs.get("pilmode") is not None
```
FirefoxMetzger, line 184:
```suggestion
            and "pilmode" in kwargs
```

We still need to wait one more month for python 3.7 to reach EoL before we can remove 3.7 support and start using the walrus operator here. 

Until then, it should be enough to just check if the argument has been used without inspecting the value (I think).

FirefoxMetzger (approved):
We are pretty much there 🚀 I found a typo and I think one line that can be slightly refactored, but otherwise this PR looks ready.

imageio/v2.py
miriam-groeneveld: That makes sense, thanks for clarifying that!

FirefoxMetzger (approved):
the linter wasn't happy, so I reformatted the respective line.

LGTM once CI comes back green.

PR #1012, opened 2023-06-30
TEST: Flag missing tests that require internet connection

olebole:
I am currently working on the update of the [imageio package](https://tracker.debian.org/pkg/python-imageio) for Debian. Debian requires that no internet resources are used during the build (which includes a test of the package).
Here I found a few test cases that require internet access but were not marked as such. This PR fixes them.

### DIFF
--- a/tests/test_freeimage.py
+++ b/tests/test_freeimage.py
@@ -175,6 +175,7 @@ def test_get_ref_im():
             assert rim.shape[:2] == (41, 31)
 
 
+@pytest.mark.needs_internet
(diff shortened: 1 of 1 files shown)


FirefoxMetzger (approved):
Thanks for the PR and well spotted @olebole 🚀 

(Also sorry for the late review. I've been on vacation in China and didn't have access to GH.)

FirefoxMetzger:
Note that CI will fail because an upstream dependency (pillow) released a new major version, which breaks one of our tests. This is unrelated to this PR and there already is a PR to address this. Once that PR's CI is green I will merge and update this PR.

PR #1023, opened 2023-07-27
fix: Python files

criadoperez:
Corrected some texts


FirefoxMetzger (approved):
Thanks for the PR @criadoperez  ❤️ 

Improvements like this are always welcome. I'll hit merge once CI is green. (You didn't do any code changes, so I suspect there to be no problems.)

PR #1034, opened 2023-08-28
Fix key error in tifffile improps

aeisenbarth:
`_get_resolution` can return None, and `ImageProperties.spacing` is optional. If "resolution" is missing, we should not raise a KeyError, but just set it to None.


FirefoxMetzger:
Thanks for the contribution @aeisenbarth !

There is another PR in flight that addresses this problem. It also refactors `_get_resolution` and removes it all together: https://github.com/imageio/imageio/pull/1031

FirefoxMetzger:
I will close this since it has been superseded by #1031 which includes this change and is now merged.

PR #1035, opened 2023-08-29
Maintenance of readthedocs.yaml

lschr:


FirefoxMetzger:
Thanks again for improving ImageIO @lschr 🚀

PR #1039, opened 2023-09-25
Fixes #992 - Turns the error raising into a conversion

### LINKED ISSUE #992: Deprecation in pillow causing problems with imageio gif writing
Hi there, thanks for all the great work on imageio! 🙂 

I had a bug report at napari/napari-animation#174 which boils down to a deprecation in pillow causing a problem with the imageio gif writing, reporting here 

```python
import numpy as np
import imageio

writer = imageio.get_writer(
    'test.gif',
    fps=20,
)
for i in range(5):
    writer.append_data(np.random.random((32, 32)).astype(np.uint8))
```

```txt
  File "/Users/alisterburt/micromamba/envs/napari-animation/lib/python3.9/site-packages/imageio/v2.py", line 215, in append_data
    return self.instance.write(im, **self.write_args)
  File "/Users/alisterburt/micromamba/envs/napari-animation/lib/python3.9/site-packages/imageio/plugins/pillow.py", line 354, in write
    raise TypeError(
TypeError: The keyword `fps` is no longer supported. Use `duration`(in ms) instead, e.g. `fps=50` == `duration=20` (1000 * 1/50).
```

14NGiestas:
Closes: #992 

So this error raising broke a PyVista function, also any other package that depends on imageio for writing GIFs/MP4s is broken.
The conversion to the new parameter is known, So just demoting it to an warning and doing what is suggested in the message itself automatically solves the issue.

It may be easier than PR every package that depends on this one to update a single parameter...

### DIFF
--- a/imageio/plugins/pillow.py
+++ b/imageio/plugins/pillow.py
@@ -387,10 +387,11 @@ class PillowPlugin(PluginV3):
 
         """
         if "fps" in kwargs:
-            raise TypeError(
+            warnings.warn(
                 "The keyword `fps` is no longer supported. Use `duration`"
                 "(in ms) instead, e.g. `fps=50` == `duration=20` (1000 * 1/50)."
             )
+            kwargs['duration'] = 1000 * 1/kwargs.get('fps')
 
         if isinstance(ndimage, list):
             ndimage = np.stack(ndimage, axis=0)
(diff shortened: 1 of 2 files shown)


FirefoxMetzger:
Fair point. A warning does make more sense here. 

[Note that this warning will become an exception again ImageIO v3.0.0 since deprecations will be removed when we update the version major (release date TBA).]

FirefoxMetzger:
Thanks for the PR @14NGiestas ! Should roll out tomorrow :)

Issue #1021, opened 2023-07-21
heif format / .heic image file support


smidm:
Are there any plans to support .heic image files? For pillow there is https://pypi.org/project/heif-image-plugin/ based on https://github.com/carsales/pyheif.

FirefoxMetzger:
There is no explicit roadmap for HEIC, but if you (or someone else) is interested in writing a plugin, I'm happy to review :)

bigcat88:
> Are there any plans to support .heic image files? For pillow there is https://pypi.org/project/heif-image-plugin/ based on https://github.com/carsales/pyheif.

Also can take a look at [pillow-heif](https://github.com/bigcat88/pillow_heif), if you wish to.

PR #1040, opened 2023-09-28
Support HEIF image file

Pandede:
See #1021.
Read HEIF image with the elegant package [pillow_heif](https://github.com/bigcat88/pillow_heif) by @bigcat88 who comments in the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 <!--next-version-placeholder-->
 
+## v2.31.4 (2023-09-25)
+
+### Fix
+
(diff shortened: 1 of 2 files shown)


Pandede:
Sorry for the redundant commits as I forgot to update my forked repository first.

bigcat88:
@Pandede you should not put it as an extra Pillow dependency.

It should be in ```extras_require`` or in ```plugins``` as a standalone record.

Pandede:
> @Pandede you should not put it as an extra Pillow dependency.
> 
> It should be in `extras_require` or in `plugins` as a standalone record.

You mean the user should install the package via `pip install imageio[heif]`?
Could this make users think that plugins other than Pillow also support HEIF?

bigcat88:
> 
> You mean the user should install the package via `pip install imageio[heif]`? Could this make users think that plugins other than Pillow also support HEIF?

Then just a usual way to just specify this as a separate entry in the plugins.

Pandede:
I remake the pull request as the redundant commits.

Issue #1044, opened 2023-10-16
Support for Pillow v10.1 (re: setting `image.mode`)


tcompa:
Hi there, and thanks for your support.

In https://github.com/imageio/imageio/blob/master/imageio/plugins/pillow.py#L312, the `mode` attribute of a `PIL.Image` object is set directly, as in
```python
image.mode = desired_mode
```

I think this is not working any more as of the recent Pillow v10.1:
> If you attempt to set the mode of an image directly, e.g. im.mode = "RGBA", you will now receive an AttributeError. This is not about removing existing functionality, but instead about raising an explicit error to prevent later consequences. The convert method is the correct way to change an image’s mode.
> (https://pillow.readthedocs.io/en/stable/releasenotes/10.1.0.html#setting-image-mode)

---

The minimal failing example would be something like
```python
from imageio.plugins.pillow import Image

img = Image.Image()
img.mode = "RGBA"
```
This example goes through with Pillow v10.0, but with Pillow v10.1 if fails as in:
```python traceback
Traceback (most recent call last):
  File "/tmp/x.py", line 4, in <module>
    img.mode = "RGBA"
AttributeError: can't set attribute 'mode'
```

EDIT: the test above is with imageio 2.31.5

---

(original issue in our repo, for the record: https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/570)

FirefoxMetzger:
👍 good find. This does break our pillow plugin in a few places, because we have reasons to _want_ those later consequences. 

In a nutshell, the `convert` function from pillow doesn't allow raw modes as input; however, using them allows us to do certain performance optimizations. For example:

https://github.com/imageio/imageio/blob/0bbf1962c7c36ebee28639aee8441e01d4168215/imageio/plugins/pillow.py#L289-L312

I need to see if/how we can keep this functionality given the recent changes.

FirefoxMetzger:
I tried to "sidestep" the `Image.mode` problem by setting `Image._mode` instead until there is a better solution available. 

Unfortunately, I found some additional problems related to GIF writing (among other things, one of our test images is now detected as a "compression bomb" by Pillow) and noticed that the JPEG writer is flakey on my local Mac (produces non-deterministic results that cause sporadic test failures).

Long story short, I will pin Pillow to v10.0.0 until I can take a better look at all the changes introduced by pillow v10.1 (PR lives here https://github.com/imageio/imageio/pull/1046).

mgorny:
[...]

PR #1045, opened 2023-10-18
BUG: avoid AttributeError when setting mode in Pillow v10.1

FirefoxMetzger:
Closes: #1044 

Pillow v10.1 removed the setter for `Image.mode` as this can have unintended consequences. This is done by making `Image.mode` a `@property` and keeping the mode in `Image._mode`. 

While okay in general, it breaks our setup as there is currently no good way of setting 16-bit return types directly (which we want when reading 16-bit grayscale PNGs). As a result, this change breaks our current setup. 

This PR introduces a "quick fix" by directly setting `Image._mode` instead of the previous `Image.mode`. I'm not fond of this solution, but until something better becomes available this is the only way I can think of to maintain our current behavior.

### DIFF
--- a/imageio/plugins/pillow.py
+++ b/imageio/plugins/pillow.py
@@ -324,7 +324,7 @@ class PillowPlugin(PluginV3):
                     "version of pillow which will make this warning dissapear.",
                     UserWarning,
                 )
-            elif minor == 0 and patch == 0:  # pillow==10.0.0
+            elif minor < 1:  # pillow<10.1.0
                 # Pillow can directly decode into 16-bit grayscale in this
                 # version
                 image.mode = desired_mode
--- a/imageio/plugins/pillow.py
+++ b/imageio/plugins/pillow.py
@@ -315,6 +315,7 @@ class PillowPlugin(PluginV3):
                 desired_mode = "I;16B"
 
             try:
+                # feature test for direct unpacking into 16-bit buffer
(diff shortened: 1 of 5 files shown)


FirefoxMetzger:
Looks like there is not just a problem with `Image.mode` becoming read-only, but also some additions to pillow's GIF reader that cause test failures.

I'll have to set some time aside to figure out where/why this breaks.

lagru:
@FirefoxMetzger thanks for the quick reaction and working on this!

lagru:
No worries. We've applied the same temporary fix to scikit-image and are in no particular hurry. Please don't feel rushed. :)

FirefoxMetzger:
Looks like I finally managed to solve all the CI issues. Now it's only test coverage left. 

I also want to check why we can now use "rawmodes" like "I;16" in `PIL.Image.convert`, which wasn't possible before but is now. I need to double-check if this internally promotes to 32-bit or reads straight into 16-bit.

FirefoxMetzger:
For future reference: Recent pillow versions no longer allow direct decoding of 16-bit grayscale PNGs. While we instruct Pillow to do so using the (now available) mode `I;16` in `Image.convert`, this will internally decode the image as 32-bit and then convert to 16-bit in a copy.

I will create a ticket over at pillow and see if we can bring the old functionality back since it is quite useful and I would eventually like to use it in other places as well.

Issue #691, opened 2021-11-20
Add rawpy backend


FirefoxMetzger:
I found an I/O library that I wasn't aware of today; perhaps it is worth adding a plugin for it. It is called rawpy and provides access to various RAW formats via [libraw](https://www.libraw.org/).

GH Link: https://github.com/letmaik/rawpy
Docs: https://letmaik.github.io/rawpy/api/
Tutorials: https://github.com/letmaik/rawpy-notebooks/blob/master/README.md

If you would like to see this backend added, please react to this post or leave a comment below. The more feedback I get on a backend, the more likely I am to prioritize implementing it over solving other issues :)

Scarlet-2:
hello i am just starting on programming and would like to make my first contribution can i help with this issue?

FirefoxMetzger:
Hej @Scarlet-2 , super cool that you decided to start learning how to code. Sure, you can absolutely take a shot at implementing this. 

If you don't mind me asking: How familiar are you already with the concepts in Open Source and (for this PR specifically) image formats? Depending on this I can suggest a way to get you started.

Scarlet-2:
well it is my first time trying to help or you could say participating in open source and i am not very familiar with image formats

FirefoxMetzger:
Right. Do you know your way around Git and your IDE of choice? 

In that case, we have some instructions on getting started with plugin writing here: https://imageio.readthedocs.io/en/stable/development/index.html#implementing-a-new-plugin 
 I'd suggest that you start setting up a dev environment for ImageIO (up until the point "Installation") and let me know if you run into trouble along the way.

Scarlet-2:
Ok if i can start getting to it i will tell you because right now i am solving some problems in my college

tahseenadit:
Hi, I would like to pick it up and work on it.

FirefoxMetzger:
Closed by: https://github.com/imageio/imageio/pull/1063

PR #1063, opened 2024-02-20
rawpy plugin

tahseenadit:
This PR adds rawpy plugin. [https://github.com/imageio/imageio/issues/691](Issue)

- Created a python script in the plugins folder for rawpy plugin wrapper.
- Created a python script in the test folder and added one test for the rawpy plugin.
- Updated setup.py, plugins.py and extensions.py to include rawpy plugin.

Note: This is work in progress. This PR has been opened to keep track of the work and is still in draft.


imageio/config/extensions.py
```diff
@@ -596,7 +596,7 @@ def reset(self):
     ),
     FileExtension(
         extension=".nef",
-        priority=["RAW-FI"],
+        priority=["rawpy", "RAW-FI"],
```
FirefoxMetzger, line 599:
```suggestion
        priority=["RAW-FI", "rawpy"],
```

Since this is a new plugin, I would like to start with it having a low priority while we test for bugs/issues. Once it has been around for a while it makes sense to have it become the default choice for supported raw formats.

imageio/config/extensions.py
```diff
@@ -762,7 +762,7 @@ def reset(self):
     ),
     FileExtension(
         extension=".raw",
-        priority=["RAW-FI", "LYTRO-ILLUM-RAW", "LYTRO-F01-RAW"],
+        priority=["rawpy", "RAW-FI", "LYTRO-ILLUM-RAW", "LYTRO-F01-RAW"],
```
FirefoxMetzger, line 765:
```suggestion
        priority=["RAW-FI", "LYTRO-ILLUM-RAW", "LYTRO-F01-RAW", "rawpy"],
```

Same as my previous comment. Start in a "low-impact" setting and slowly increase.

imageio/plugins/rawpy.py
```diff
@@ -0,0 +1,70 @@
+# -*- coding: utf-8 -*-
+# imageio is distributed under the terms of the (new) BSD License.
```
FirefoxMetzger, line 2: We typically don't add a license header. Any reason you've added it here?

imageio/plugins/rawpy.py
```diff
@@ -0,0 +1,70 @@
...
+import rawpy
+import numpy as np
+
+from ..core.request import URI_BYTES, InitializationError, IOMode, Request
+from ..core.v3_plugin_api import PluginV3
+
+
+class RawPyPlugin(PluginV3):
+    """A class representing the rawpy plugin.
+    This class is a subclass of PluginV3.
```
FirefoxMetzger, line 19:
```suggestion
```

Redudant with `RawPyPlugin(PluginV3)`. Let's keep the docs lean :)

imageio/plugins/rawpy.py
```diff
@@ -0,0 +1,70 @@
...
+
+    .. autosummary::
+    :toctree: _plugins/rawpy
+
+    RawPyPlugin.read
+    """
+    def __init__(self, request: Request) -> None:
+        """Instantiates a new rawpy plugin object
+
+        parameters
```
FirefoxMetzger, line 32:
```suggestion
        Parameters
```

imageio/plugins/rawpy.py
```diff
@@ -0,0 +1,70 @@
...
+        ----------
+        request: {Request}
```
FirefoxMetzger, line 34:
```suggestion
        request : Request
```

Our docs are a bit inconsistent on this end (due to legacy). I'm trying to slowly move towards a common format.

imageio/plugins/rawpy.py
```diff
@@ -0,0 +1,70 @@
...
+            A request object representing the resource to be operated on.
+        """
+
+        super().__init__(request)
+
+        self._image: rawpy = None
+        self._image_file = request.get_file()
```
FirefoxMetzger, line 41: TODO: You are likely aware, but you will have to add some logic here that tests if rawpy can read the resource that is passed in. 

You'll have to verify the header, check the extension, or something similar. Likely rawpy has a built-in way of doing this and all you need to do is pass the file handle to rawpy, handle a potential exception, and raise `IOError` if the image is incompatible or can not be read with rawpy.

This is done because plugin selection will test a plugin's compatibility by instantiating it with the given request. If this works (no exception is raised) it will assume that the plugin can handle the request/image and will stop searching for alternatives.

imageio/plugins/rawpy.py
```diff
@@ -0,0 +1,70 @@
...
+
+        Returns
+        -------
+        image: ndimage
+            The image data
+        """
+
+        nd_image = None
+
+        if self._request.mode.io_mode == IOMode.read:
```
FirefoxMetzger, line 54: This test is redundant. By the time `read` is called we should have already opened the file in mode `"r"` and it is safe to assume/assert that the request's mode was "read".

imageio/plugins/rawpy.py
```diff
@@ -0,0 +1,70 @@
...
+                    self._image = raw_image.postprocess(**kwargs)
+                nd_image = self._image
+            except (AttributeError, TypeError) as ex:
+                print(ex)
+            except (rawpy.NotSupportedError, rawpy.LibRawError, rawpy.LibRawFatalError, rawpy.LibRawNonFatalError) as ex:
+                print(ex)
+            except Exception as ex:
+                print(ex)
+        else:
+            print("Read is not supported!")
```
FirefoxMetzger, line 66: Could you explain the logic of this branch? Since `read` is returning something other than `NotImplementedError` I would have assumed that Rawpy supports reading files.

setup.py
```diff
@@ -100,6 +100,7 @@
     "tifffile": ["tifffile"],
     "pyav": ["av"],
     "pillow-heif": ["pillow-heif"],
+    "rawpy": ["rawpy"]
```
FirefoxMetzger, line 103: Do you know if there are additional installation steps beyond installing the python package?

tests/test_rawpy.py
```diff
@@ -0,0 +1,20 @@
...
+import rawpy
+import imageio.v3 as iio
+
+
+def test_nef_local(test_images):
+    """Test for reading .nef file from .test_images dir.
+    """
+    # Check image path
+    im_path = str(test_images / "Nikon.nef")
+    assert os.path.exists(im_path)
```
FirefoxMetzger, line 14:
```suggestion
```

No need to assert this. We are not testing the correctness of our testing fixtures here. (We assume they work as they should)

All assertions should be done to assert correctness of the RawPy plugin. This way pytest will report "failed" if the test fails and "Error" if something in the supporting logic fails.

tests/test_rawpy.py
```diff
@@ -0,0 +1,20 @@
...
+
+    # Test if plugin's content mathces rawpy content
+    im_plugin = iio.imread(im_path)
+    im_rawpy = rawpy.imread(im_path)
+    assert im_plugin.all() == im_rawpy.raw_image.all()
```
FirefoxMetzger, line 19: This will not do what you expect. `ndarray.all()` will return `True` if all elements in the array evaluate to `True` and `False` otherwise. In other words, the current logic checks if both images have at least one pixel with value `0`.

Instead, use the following check:

```python
assert np.allclose(im_plugin, im_rawpy)
```

Sidenote: In my other tests, I tend to use the names
```
actual = iio.imread(im_path)
expected = rawpy.imread(im_path)
```

You are, obviously, free to use any variable name you like, but I think this choice makes it quite clear what the purpose of these arrays is, i.e., it is immediately obvious that `expected` contains the desired result whereas `actual` is what the plugin has produced.

FirefoxMetzger (commented):
Left a quick review. Great progress so far 🚀

So far I only have small comments on how plugins are used/digested by ImageIO and how you will have to adjust the logic to comply with these particularities.

imageio/plugins/rawpy.py
tahseenadit: I just saw it in other file and assumed it was needed. I have removed it now.

imageio/plugins/rawpy.py
tahseenadit: Removed the redundant text.

imageio/plugins/rawpy.py
tahseenadit: Thanks for the clarification. I have updated the file according to your comment.

setup.py
tahseenadit: I did not have to do any additional installation steps. So, I think only installing the python package should be enough.

tests/test_rawpy.py
tahseenadit: Ok, thanks! I have updated it now.

tests/test_rawpy.py
tahseenadit: Updated the file accordingly. Thanks for the explanation.

imageio/config/extensions.py
tahseenadit: Thanks! I will make it a low priority when the all implementations are done. I have kept it so far in this way to complement tests.

imageio/plugins/rawpy.py
```diff
@@ -0,0 +1,94 @@
...
+        ----------
+        request: {Request}
+            A request object representing the resource to be operated on.
+        """
+
+        super().__init__(request)
+
+        self._image_file = None
+
+        if request.mode.io_mode == IOMode.read:
```
FirefoxMetzger, line 40: Since rawpy doesn't support writing, we should check if the plugin is called with `IOMode.write` and if so raise an `InitializationError` with the message that rawpy can't write images.

imageio/plugins/rawpy.py
```diff
@@ -0,0 +1,94 @@
...
+            try:
+                with rawpy.RawPy.open_buffer(request.get_file()):
+                    self._image_file = request.get_file()
+            except rawpy.NotSupportedError as ex:
+                raise InitializationError(
+                    f"RawPy can not read {request.raw_uri}."
+                ) from None
```
FirefoxMetzger, line 47: We need to catch an edge case here: If the user provides a raw bytes string to read, we don't want to blurt it out to the console.

Could you check here if the request type is `URI_BYTES` and if so choose an error message that doesn't include the raw URI?

imageio/plugins/rawpy.py
```diff
@@ -0,0 +1,94 @@
...
+        """
+
+        super().__init__(request)
+
+        self._image_file = None
+
+        if request.mode.io_mode == IOMode.read:
+            try:
+                with rawpy.RawPy.open_buffer(request.get_file()):
+                    self._image_file = request.get_file()
```
FirefoxMetzger, line 43: Seeing this now, don't forget to also implement the `def close()` method to close the file once we are done with it and avoid leaking a file handle.

imageio/plugins/rawpy.py
```diff
@@ -0,0 +1,94 @@
...
+            The image data
+        """
+
+        nd_image: np.ndarray
+        
+        try:
+            with rawpy.imread(self._image_file) as raw_image:
+                nd_image = raw_image.postprocess(**kwargs)
+        except Exception as ex:
+            print(ex)
```
FirefoxMetzger, line 66: No need to print an exception here. Unless you want to do one of the following:
- have an explicit idea how to correct what went wrong
- want to modify the error message
- want to chain an exception

It's better to just let exceptions "bubble up" and have the caller worry about handling it. This allows downstream code to implement more sophisticated exception handling and/or will give the user precise feedback on what went wrong so they can investigate and fix the issue.

imageio/plugins/rawpy.py
```diff
@@ -0,0 +1,94 @@
...
+
+        return nd_image
+    
+    def write(
+        self, 
+        ndimage: Union[ArrayLike, List[ArrayLike]]
+    ) -> bytes | None:
+        """RawPy implementation not found.
+        """
```
FirefoxMetzger, line 75:
```suggestion
        """RawPy does not support writing.
        """
```

imageio/plugins/rawpy.py
```diff
@@ -0,0 +1,94 @@
...
+        nd_image: np.ndarray
+        
+        try:
+            with rawpy.imread(self._image_file) as raw_image:
+                nd_image = raw_image.raw_image
+        except Exception as ex:
+            print(ex)
+
+        return iter(nd_image)
```
FirefoxMetzger, line 94: I don't think this will produce what we expect.

- You are never calling `postprocess` in this function and thus never converting from bayer to RGB.
- You are calling `iter` on a numpy array, which will iterate over the first dimension of the array. If a "normal" (2D/3D) image is passed, this means we will iterate over the column dimension and return the image one row at a time.

Until we know if/how rawpy supports multi-frame images, let's re-use our `read` method:

```python
yield self.read(...)  # TODO: add the correct signature 
```

This way `iter` becomes a generator that will yield a single image and then terminate. We'll also need to match the expected signature of `iter(...)` and handle some cases regarding `index`. However, we can take those later, once the main functions are implemented.

FirefoxMetzger (commented):
Nice progress 🚀 

I started to comment on edge cases, which is a sign that we are getting closer to the finish line for the functions that are currently implemented. 

Metadata is still missing and we should start thinking of docstrings next, but this is something we can take one step at a time.

Good work!

imageio/plugins/rawpy.py
tahseenadit: Thank you for your comment. I have updated it now accordingly. I also noticed that `rawpy.imread()` just calls `open_buffer()` without doing any additional thing. In addition to that, I did not use the `with` context manager because then I can not read the file again from the request object using `get_file()` in the read method. Therefore, I have changed my implementation a bit. Please check and let me know what do you think.

imageio/plugins/rawpy.py
tahseenadit: Thank you for your comment. I have updated it now accordingly.

imageio/plugins/rawpy.py
tahseenadit: Thank you for your comment. I have updated it now accordingly.

imageio/plugins/rawpy.py
tahseenadit: Thank you for your comment. I have updated it now accordingly. Let me know if it looks ok or not.

imageio/plugins/rawpy.py
tahseenadit: Thank you for your comment. I have updated it now accordingly.

imageio/plugins/rawpy.py
```diff
@@ -0,0 +1,187 @@
...
+                        f"RawPy can not read {request.raw_uri}."
+                    ) from None
+        elif request.mode.io_mode == IOMode.write:
+            raise InitializationError(
+                "RawPy does not support writing."
+            ) from None
+
+    def close(self) -> None:
+
+        if self._image_file:
```
FirefoxMetzger, line 59:
```suggestion
        if self._image_file:
```

imageio/plugins/rawpy.py
```diff
@@ -0,0 +1,187 @@
...
+        Returns
+        -------
+        nd_image: ndarray
+            The image data
+        """
+
+        try:
+            yield self.read()
+        except Exception as ex:
+            raise ex
```
FirefoxMetzger, line 102: I think this is rather redundant. If an exception gets raised here the default behavior is to forward it to the caller. Here, you are manually forwarding it to the caller. Is there any reason why?

imageio/plugins/rawpy.py
```diff
@@ -0,0 +1,187 @@
...
+
+        ImageSize = self._image_file.sizes
+
+        # TO DO: Find out what information needs to be included below and refine accordingly.
+        metadata["black_level_per_channel"] = self._image_file.black_level_per_channel
+        metadata["camera_white_level_per_channel"] = self._image_file.camera_white_level_per_channel
+        metadata["color_desc"] = self._image_file.color_desc
+        #metadata["color_matrix"] = self._image_file.color_matrix
+        metadata["daylight_whitebalance"] = self._image_file.daylight_whitebalance
+        metadata["dtype"] = self._image_file.raw_image.dtype.type
```
FirefoxMetzger, line 134:
```suggestion
        metadata["dtype"] = self._image_file.raw_image.dtype
```

Shouldn't this just be the numpy `dtype`?

imageio/plugins/rawpy.py
```diff
@@ -0,0 +1,187 @@
...
+        #metadata["raw_colors_visible"] = self._image_file.raw_colors_visible
+        #metadata["raw_image"] = self._image_file.raw_image
+        #metadata["raw_image_visible"] = self._image_file.raw_image_visible
+        #metadata["raw_pattern"] = self._image_file.raw_pattern
+        metadata["raw_type"] = self._image_file.raw_type.name
+        #metadata["rgb_xyz_matrix"] = self._image_file.rgb_xyz_matrix
+        #metadata["tone_curve"] = self._image_file.tone_curve
+        metadata["width"] = ImageSize.width
+        metadata["height"] = ImageSize.height
+        metadata["shape"] = (ImageSize.width, ImageSize.height)
```
FirefoxMetzger, line 147:
```suggestion
        metadata["width"] = ImageSize.width
        metadata["height"] = ImageSize.height
```

You don't need to build/combine metadata into more complex types here. It might be nice to in the future if we get a lot of feedback that this is expected and/or something users do themselves a lot. However, I would start with the simplest setup, which is to pass the fields as-is.

imageio/plugins/rawpy.py
```diff
@@ -0,0 +1,187 @@
...
+
+        """
+
+        ImageSize = self._image_file.sizes
+
+        width: int = ImageSize.width
+        height: int = ImageSize.height
+        shape: Tuple[int, ...] = (height, width)
+
+        dtype = self._image_file.raw_image.dtype.type
```
FirefoxMetzger, line 182: Hm, I need to think about this one. You are forwarding the dtype of the _raw_ image here. However, all the methods (read/iter) provide the postprocessed image, which may have changed its type and thus no longer match what the type of the raw image. That said, I don't know if this can happen in rawpy, so I need to look into if this is the right thing to do, or not.

tests/test_rawpy.py
```diff
@@ -0,0 +1,49 @@
...
+import rawpy
+import imageio.v3 as iio
+import numpy as np
+
+
+def test_nef_local(test_images):
+    """Test for reading .nef file from .test_images dir.
+    """
+    # Construct image path
+    im_path = "e:/Explorations/imageio-tahseen-new/.test_images/infrared.nef" #test_images / "infrared.nef"
```
FirefoxMetzger, line 13: How come you've replaced this with a path local to your filesystem?

FirefoxMetzger (commented):
Really nice work @tahseenadit 🚀 

Code-wise this looks pretty much ready to go except and there are now some action items for me:
- Provide you feedback on which metadata fields need exporting and which we should keep out of the dict.
- Check if forwarding the raw image's `dtype` is the correct thing to do in properties.
- Find some permissively-licensed test images you can use for unit testing.

For you, there are two things left (beyond minor changes after me completing above tasks):

1. Handling of `index=...` in `read`.
2. Getting test coverage to 100%.

Regarding the first point, we have this piece of docs in the `PluginV3` class:

> If a PNG file (single image format) is read with
        ``index=None`` the plugin does a very similar thing: It loads all
        ndimages in the file (here it's just one) and stacks them along a new
        first axis, effectively prepending an axis with size 1 to the image. If
        a plugin does not wish to support ``index=None`` it should set a more
        sensible default and raise a ``ValueError`` when requested to read using
        ``index=None``.

The docs are slightly outdated in that it should read `index=...` where it says `index=None` (I will update). The logic here is that some images may contain multiple frames/images and users should be able to conveniently request _all_ images as a batch.

For rawpy we currently think that it handle single-image formats only, so handling this is fairly easy: Whenever a user calls `file.read("some_image", index=...)` we read the image and then add an "empty" dimension at the front. This can be as simple as `image = image[None, ...]`.

Regarding the second point, you've already started adding some tests. Keep doing so and feel free to use images that you have the right to use for this purpose :) Once I find pictures we can use and distribute I will share them here and then we can keep the test logic but replace the image.

FirefoxMetzger:
I also just saw that we are getting some CI failures due to the use of `bytes | None` in your type hints.

We need to solve this a bit differently here, because we need to support all python versions (currently 3.8 - 3.12) and not all of them support this syntax yet. Instead you'll have to do something like:

```python
from typing import Optional

... # somewhere in the code
... -> Optional[bytes]
```

FirefoxMetzger:
Adding this here for later: I found that rawpy can raise a `LibRawFileUnsupportedError` error when reading an unsupported image. We should catch this during initialization and rase a `InitializationError` when that happens.

FirefoxMetzger:
> Find some permissively-licensed test images you can use for unit testing.

I found a source of images: https://raw.pixls.us/

The website has a _long_ list of CC0 licenced raw images that we can use (and distribute) for unit testing :rocket::heart: @tahseenadit feel free to have a look around and pick some images in different formats. Try to choose smaller files so that the initial image dataset download doesn't blow up too much 🙏

imageio/plugins/rawpy.py
tahseenadit: Thanks for the suggestion. I have updated the script accordingly.

imageio/plugins/rawpy.py
tahseenadit: Thank you for mentioning it. I was unaware about the default behavior. I have updated it now.

imageio/plugins/rawpy.py
tahseenadit: Are you suggesting here to remove the newline ? If yes, then I have now removed it.

imageio/plugins/rawpy.py
tahseenadit: I tried using only dtype in the beginning but I was getting some error. I tried again today and now it looks like it is working. May be the error was happening because of something else. I have updated the file now.

imageio/plugins/rawpy.py
tahseenadit: Thank you for the comment. I do not have much idea about the best practices here. So I started including them and thought I would later remove it if I receive any comment. I have updated the file accordingly now.

tests/test_rawpy.py
tahseenadit: I forgot to revert the change before the commit. I replaced this with a path local to my filesystem because when I use the `test_images` parameter, I get the following error:

`e:\Explorations\imageio-tahseen-new\tests\test_rawpy.py::test_nef_local failed: TypeError: bad argument type for built-in operation

The above exception was the direct cause of the following exception:

test_images = WindowsPath('e:/Explorations/imageio-tahseen-new/.test_images')

    def test_nef_local(test_images):
        """Test for reading .nef file from .test_images dir.
        """
        # Construct image path
        im_path = test_images / "infrared.nef"
    
        # Test if plugin's content mathces rawpy content
        actual = iio.imread(im_path, plugin="rawpy")
>       expected = rawpy.imread(im_path).postprocess()

tests\test_rawpy.py:17: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
.venv\Lib\site-packages\rawpy\__init__.py:20: in imread
    d.open_file(pathOrFile)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

>   ???
E   SystemError: <class 'RuntimeError'> returned a result with an exception set

rawpy\\_rawpy.pyx:404: SystemError`

So, I replaced the value with a local path for quick test. Do you know why I am seeing this error ? Is it because of my OS ?

imageio/plugins/rawpy.py
tahseenadit: One more thing, raw_image has raw_shape with raw_width and raw_height. The raw_height is slightly different than ImageSize.height of the raw image. In the documentation, there is ImageSize for width and height.  Do you think I should use raw_height instead of height from the ImageSize ? What will be the shape for both metadata and properties in that case ? Or should I keep them both as I have done now ?

imageio/plugins/rawpy.py
FirefoxMetzger: yes, that's what I was suggesting

tests/test_rawpy.py
FirefoxMetzger: Oh, that looks like a `String` vs `pathlib.Path` error. can you try again with

```python
im_path = test_images / "infrared.nef"
actual = iio.imread(str(im_path), plugin="rawpy")
```

If this works, then we have to update the plugin, because ImageIO also allows `pathlib.Path` inputs.

tahseenadit: Hi, the issue is not happening with imageio plugin. It happens when I try to directly read the image using rawpy in my test. The error is in the line `expected = rawpy.imread(im_path)` and not `iio.imread(im_path, plugin="rawpy")` . I have updated the code to `expected = rawpy.imread(str(im_path))` and now it is working.

imageio/plugins/rawpy.py
```diff
@@ -0,0 +1,186 @@
...
+        """
+
+        nd_image: np.ndarray
+
+        try:
+            nd_image = self._image_file.postprocess(**kwargs)
+        except Exception:
+            pass
+        
+        nd_image = nd_image[None, ...]
```
FirefoxMetzger, line 79:
```suggestion
        if index is Ellipsis:
            nd_image = nd_image[None, ...]
```

We should only append an initial dimension (the `batch`) dimension, when the user asks for `index=...`. Otherwise, we return the first (and only) image as-is.

FirefoxMetzger (commented):
Nice work! Only very minor changes from my side this time.

I noticed that the docs are not picking up the new plugin. I will check this evening why that is the case and will let you know on Wednesday.

FirefoxMetzger (commented):
A bit later than I wanted, but I am now done with another round of review. I also went through the source of rawpy to see which pieces of metadata should live where.

The one thing I haven't managed to work out yet is how the raw `dtype` relates to the actual `dtype` of the returned image. What I can say so far is that there are more raw types than returned types and, hence, we can't just forward the raw `dtype` like we do now. However, I'm still not 100% sure what the correct alternative is so lets leave it as-is until we know :)

imageio/plugins/rawpy.py
```diff
@@ -49,10 +49,11 @@ def __init__(self, request: Request) -> None:
                     raise InitializationError(
                         f"RawPy can not read {request.raw_uri}."
                     ) from None
-        elif request.mode.io_mode == IOMode.write:
-            raise InitializationError(
-                "RawPy does not support writing."
-            ) from None
+        # Why do we need this when we are raising NotImplementedError in the write method ?
```
FirefoxMetzger, line 52: Because the `InitializationError` is used during plugin selection. It allows ImageIO to try this plugin to fulfill a request (e.g. for an unknown file format), realize it won't work, and try a different one before returning control to the user.

tests/test_rawpy.py
```diff
@@ -1,48 +1,97 @@
...
     assert np.allclose(actual, expected)
 
 
+@pytest.mark.parametrize(
+    "im_in",
+    [
+       ("camera.png"),
+       ("bricks.jpg"), 
+    ],
+)
```
FirefoxMetzger, line 36: I don't think we need to parametrize a failure test. One image enough to hit the code-path I suspect?

tests/test_rawpy.py
```diff
@@ -1,48 +1,97 @@
...
+    """Test for writing .nef file from .test_images dir.
+    """
+    # Construct image path
+    im_path = test_images / "Blackmagic.dng"
+
+    # the base image as numpy array
+    im_actual = iio.imread(im_path, index=..., plugin="rawpy")
+    iio_file = tmp_path / "test.png"
+    with pytest.raises(NotImplementedError):
+        iio.imwrite(iio_file, im_actual, plugin="rawpy")
```
FirefoxMetzger, line 57: Writing is not supported, so we can actually skip this test and (if necessary) exclude the `raise NotImplementError` line from coverage.

FirefoxMetzger (commented):
Quick review before our meeting today. Looking good so far :)

imageio/plugins/rawpy.py
```diff
@@ -0,0 +1,186 @@
...
+        Returns
+        -------
+        metadata : dict
+            A dictionary of format-specific metadata.
+
+        """
+
+        metadata = {}
+
+        ImageSize = self._image_file.sizes
```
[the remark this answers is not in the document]
tahseenadit: Sorry about that. I did not notice earlier. It's been fixed now.

imageio/plugins/rawpy.py
tahseenadit: Done. Thanks for the review.

tests/test_rawpy.py
tahseenadit: The test is removed as per our last discussion. Thanks for the review

tests/test_rawpy.py
tahseenadit: The test has been removed as per our last discussion. Thanks for the review.

imageio/plugins/rawpy.py
```diff
@@ -0,0 +1,186 @@
...
+        metadata["flip"] = ImageSize.flip
+        metadata["num_colors"] = self._image_file.num_colors
+        #metadata["raw_colors"] = self._image_file.raw_colors
+        #metadata["raw_colors_visible"] = self._image_file.raw_colors_visible
+        #metadata["raw_image"] = self._image_file.raw_image
+        #metadata["raw_image_visible"] = self._image_file.raw_image_visible
+        #metadata["raw_pattern"] = self._image_file.raw_pattern
+        metadata["raw_type"] = self._image_file.raw_type.name
+        #metadata["rgb_xyz_matrix"] = self._image_file.rgb_xyz_matrix
+        #metadata["tone_curve"] = self._image_file.tone_curve
```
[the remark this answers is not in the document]
tahseenadit: Done. But in previous comment, you mentioned that raw_pattern, raw_type and rgb_xyz_matrix are config paramteres. Should we still include them ?

imageio/plugins/rawpy.py
```diff
@@ -0,0 +1,186 @@
...
+        metadata["dtype"] = self._image_file.raw_image.dtype
+        metadata["flip"] = ImageSize.flip
+        metadata["num_colors"] = self._image_file.num_colors
+        #metadata["raw_colors"] = self._image_file.raw_colors
+        #metadata["raw_colors_visible"] = self._image_file.raw_colors_visible
+        #metadata["raw_image"] = self._image_file.raw_image
+        #metadata["raw_image_visible"] = self._image_file.raw_image_visible
+        #metadata["raw_pattern"] = self._image_file.raw_pattern
+        metadata["raw_type"] = self._image_file.raw_type.name
+        #metadata["rgb_xyz_matrix"] = self._image_file.rgb_xyz_matrix
```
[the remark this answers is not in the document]
tahseenadit: Removed except raw_pattern, raw_type and rgb_xyz_matrix. Can you tell me how did you find out if these are metadata of the image or config paramteres of RawPy ? I would love to learn about it.

imageio/plugins/rawpy.py
```diff
@@ -0,0 +1,186 @@
...
+
+        # TO DO: Find out what information needs to be included below and refine accordingly.
+        metadata["black_level_per_channel"] = self._image_file.black_level_per_channel
+        metadata["camera_white_level_per_channel"] = self._image_file.camera_white_level_per_channel
+        metadata["color_desc"] = self._image_file.color_desc
+        #metadata["color_matrix"] = self._image_file.color_matrix
+        metadata["daylight_whitebalance"] = self._image_file.daylight_whitebalance
+        metadata["dtype"] = self._image_file.raw_image.dtype
+        metadata["flip"] = ImageSize.flip
+        metadata["num_colors"] = self._image_file.num_colors
```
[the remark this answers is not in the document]
tahseenadit: Done. Can you please also tell me why we should remove it if something is consumed by postprocess? Thanks!

imageio/plugins/rawpy.py
```diff
@@ -0,0 +1,186 @@
...
+        metadata = {}
+
+        ImageSize = self._image_file.sizes
+
+        # TO DO: Find out what information needs to be included below and refine accordingly.
+        metadata["black_level_per_channel"] = self._image_file.black_level_per_channel
+        metadata["camera_white_level_per_channel"] = self._image_file.camera_white_level_per_channel
+        metadata["color_desc"] = self._image_file.color_desc
+        #metadata["color_matrix"] = self._image_file.color_matrix
+        metadata["daylight_whitebalance"] = self._image_file.daylight_whitebalance
```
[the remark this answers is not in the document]
tahseenadit: Updated the function accordingly. Thanks for the review!

imageio/plugins/rawpy.py
```diff
@@ -0,0 +1,186 @@
...
+        metadata["raw_type"] = self._image_file.raw_type.name
+        #metadata["rgb_xyz_matrix"] = self._image_file.rgb_xyz_matrix
+        #metadata["tone_curve"] = self._image_file.tone_curve
+        metadata["width"] = ImageSize.width
+        metadata["height"] = ImageSize.height
+        metadata["raw_width"] = ImageSize.raw_width
+        metadata["raw_height"] = ImageSize.raw_height
+        metadata["raw_shape"] = self._image_file.raw_image.shape
+        metadata["iwidth"] = ImageSize.iwidth
+        metadata["iheight"] = ImageSize.iheight
```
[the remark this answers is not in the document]
tahseenadit: Updated the function accordingly. Thanks for the review!

imageio/plugins/rawpy.py
```diff
@@ -0,0 +1,186 @@
...
+        metadata["pixel_aspect"] = ImageSize.pixel_aspect
+        metadata["white_level"] = self._image_file.white_level
```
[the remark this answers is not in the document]
tahseenadit: Updated the function accordingly. Thanks for the review!

imageio/plugins/rawpy.py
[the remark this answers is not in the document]
tahseenadit: Also, after updating the metadata based on all the comments above, it seems only width, height and pixel_aspect are the ones which get included when exclude_applied is set to True. So, in the test function for metadata, I have only included these three since exclude_applied is set to True by default. Please let me know if that is ok. Thanks!

imageio/plugins/rawpy.py
tahseenadit: Makes sense. Thanks for the review. I have updated the function accordingly.

imageio/plugins/rawpy.py
```diff
@@ -0,0 +1,186 @@
...
+        properties : ImageProperties
+            A dataclass filled with standardized image metadata.
+
+        Notes
+        -----
+        This does not decode pixel data and is fast for large images.
+
+        """
+
+        ImageSize = self._image_file.sizes
```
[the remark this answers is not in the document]
tahseenadit: Done, thanks for the review.

imageio/plugins/rawpy.py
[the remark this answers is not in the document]
FirefoxMetzger: Because it is less confusing to the user. Take the `flip` metadata as an example. Suppose I do

```
img = iio.imread("foo.dng")
meta = iio.immeta("foo.dng")
```

What does the value of `meta["flip"]` mean here? Do I have to manually flip `img` to restore it to its original form, or should I ignore `flip` because ImageIO flipped it already? If both the user and ImageIO flip the image (or nobody flips) the processing is wrong, but there is no way of knowing without knowing the plugin's internal behavior.

I need to know how to restore the image to its original form including the application of any processing steps that apply after decoding (white balance, flip, rotation, gamma correction, ...). Some plugins apply these by default, others don't. So, to provide a feedback mechanism to the user on what has been applied and what hasn't we hide metadata fields that are applied by default - hence `exclude_applied=True`.

However, some programs need to inspect the _raw_ metadata in the file regardless of the processing applied by default. An example of such a use-case is a programs that coordinates an image processing pipelines. A program might decide that it wants to apply special correction to images that are over-exposed (too bright). In this scenario it needs to know the value of, say, `white_level` to make the correct decision and uses `exclude_applied=False` to get the metadata applicable to the raw file - which we would hide by default because the plugin applies white balance correction by default.

imageio/plugins/rawpy.py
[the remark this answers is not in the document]
FirefoxMetzger: I read the source code of rawpy 😅

imageio/plugins/rawpy.py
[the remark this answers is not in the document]
FirefoxMetzger: Oh we are being trolled by the Github UI 🤣

I left the comment on the last line `"tone_curve"`, but apparently GH always shows a minimum of 4 lines for a comment. I see how that becomes confusing. Please keep `tone_curve` and remove the others :)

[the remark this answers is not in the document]
FirefoxMetzger: It's perfectly fine to only assert values of the metadata that is returned by default.

imageio/plugins/rawpy.py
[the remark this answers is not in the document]
tahseenadit: Thank you so much for the explanation!

imageio/plugins/rawpy.py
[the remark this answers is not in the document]
tahseenadit: Thank you so much for the answer!

imageio/plugins/rawpy.py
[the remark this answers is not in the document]
tahseenadit: Thanks! I have removed others except tone_curve.

tahseenadit:
> I also just saw that we are getting some CI failures due to the use of `bytes | None` in your type hints.
> 
> We need to solve this a bit differently here, because we need to support all python versions (currently 3.8 - 3.12) and not all of them support this syntax yet. Instead you'll have to do something like:
> 
> ```python
> from typing import Optional
> 
> ... # somewhere in the code
> ... -> Optional[bytes]
> ```

Just leaving a reply here that it has been updated accordingly.

PR #1068, opened 2024-03-17
Fix incorrect DICOMSeries order

### LINKED ISSUE #1067: Only partial DICOM layers loading in Google Colab using imageio.volread
I'm encountering an issue while trying to load DICOM files using imageio.volread in Google Colab. I have a folder containing 27 DICOM layers, but when I try to load them using the following code snippet:

```python
import imageio

folder_path = 'Data/Brain/SE000001'
vol = imageio.volread(folder_path, 'DICOM')
print('Shape of image array:', vol.shape)
```

Only 5 out of the 27 layers are being loaded. So I get the output:

```python
Warning: missing file after '/content/ImageAnalysis/ImageAnalysis/Data/Brain/SE000001/MR000023'
Reading DICOM (examining files): 27/27 files (100.0%)
  Found 6 correct series.
Reading DICOM (loading data): 5/5  (100.0%)
Shape of image array: (5, 256, 256)
```

The volume (27 slices) is 3.8MB in size.

This code works fine in my local Jupyter Notebook within VS Code, but not in Google Colab. I've checked the file paths and they seem to be correct. Are there any known issues with imageio.volread in Google Colab or any potential workarounds for this problem?

I've also tried imageio.v2 and imageio.v3, the problem is still the same. I only get the warning on Colab. The version `imageio==2.34` is running on both systems. I imported the files into Google Colab at the beginning using `!git clone`. If I use pydicom, for example, I can also load the pixel data from all files, I don't get any errors.

Any help or insights would be greatly appreciated! Thanks!

Pandede:
Fixes #1067.
The root cause is the series order in `DicomSeries`.
The whole series shares the identical `InstanceNumber`. When `DicomSeries._sort` is invoked, the sorting is no longer sorted by `InstanceNumber`, but the **filename**:
```python
# File order in local
[
  './DICOM/DICOM/ST000000/SE000001/MR000000',
  './DICOM/DICOM/ST000000/SE000001/MR000001',
  './DICOM/DICOM/ST000000/SE000001/MR000002',
  './DICOM/DICOM/ST000000/SE000001/MR000003',
  './DICOM/DICOM/ST000000/SE000001/MR000004',
  './DICOM/DICOM/ST000000/SE000001/MR000005',
  ...
]
# File order in Google Colab
[
  '/content/DICOM/DICOM/ST000000/SE000001/MR000020',
  '/content/DICOM/DICOM/ST000000/SE000001/MR000014',
  '/content/DICOM/DICOM/ST000000/SE000001/MR000015',
  '/content/DICOM/DICOM/ST000000/SE000001/MR000001',
  '/content/DICOM/DICOM/ST000000/SE000001/MR000011',
  ...
]
```
This makes function `splitSerieIfRequired` splits into 1 series in **local** while 8 series in **Google Colab**.
In summary, the series should not be sorted by **filename**, it should be sorted by the attribute `ImagePositionPatient[2]` (The z-index of the image?) in the DICOM file.

### DIFF
--- a/imageio/plugins/_dicom.py
+++ b/imageio/plugins/_dicom.py
@@ -703,7 +703,10 @@ class DicomSeries(object):
         self._entries.append(dcm)
 
     def _sort(self):
-        self._entries.sort(key=lambda k: (k.InstanceNumber, k.ImagePositionPatient[2]))
+        self._entries.sort(key=lambda k: (
+            k.InstanceNumber, 
+            getattr(k, 'ImagePositionPatient[2]', None)
+        ))
 
     def _finish(self):
         """
--- a/imageio/plugins/_dicom.py
+++ b/imageio/plugins/_dicom.py
@@ -703,7 +703,7 @@ class DicomSeries(object):
         self._entries.append(dcm)
 
     def _sort(self):
-        self._entries.sort(key=lambda k: k.InstanceNumber)
+        self._entries.sort(key=lambda k: (k.InstanceNumber, k.ImagePositionPatient[2]))
 
     def _finish(self):
         """
--- a/imageio/plugins/_dicom.py
+++ b/imageio/plugins/_dicom.py
@@ -703,10 +703,12 @@ class DicomSeries(object):
         self._entries.append(dcm)
 
     def _sort(self):
-        self._entries.sort(key=lambda k: (
-            k.InstanceNumber, 
-            getattr(k, 'ImagePositionPatient[2]', None)
-        ))
+        self._entries.sort(
+            key=lambda k: (
+                k.InstanceNumber,
+                getattr(k, "ImagePositionPatient[2]", None),
+            )
+        )
 
     def _finish(self):
         """


FirefoxMetzger:
> the series should not be sorted by filename, it should be sorted by the attribute ImagePositionPatient[2] (The z-index of the image?) in the DICOM file

Yep, `ImagePositionPatient[2]` is similar to the z-order of the image stack. It's the z-component of a reference frame that is shared by all images in a series, so if you order by that you get consistent z-ordering and if you look at the values you get information on how far the scans are apart (the spacing along Z).

> The whole series shares the identical InstanceNumber. When DicomSeries._sort is invoked, the sorting is no longer sorted by InstanceNumber, but the filename

To clarify: When the `InstanceNumber` is identical then python will keep files in the order in which they were listed by the OS. Most OS make no guarantees about the order they return when listing a filesystem. While this typically is alphabetical (especially for single disks) it isn't always the case for distributed disks ... and apparently S3 is one example where this is _not_ the case.

Awesome detective work @Pandede - this would have taken me quite a while to find. 🚀

FirefoxMetzger (approved):
LGTM

PR #1069, opened 2024-03-19
Fix bug of #1068 that `getattr` cannot get attribute with index directly

Pandede:
Small fix on PR #1068.
I think I should add the testcase to prove the PR works. Is that possible to expand the test image set?

### DIFF
--- a/imageio/plugins/_dicom.py
+++ b/imageio/plugins/_dicom.py
@@ -706,7 +706,11 @@ class DicomSeries(object):
         self._entries.sort(
             key=lambda k: (
                 k.InstanceNumber,
-                getattr(k, "ImagePositionPatient[2]", None),
+                (
+                    k.ImagePositionPatient[2]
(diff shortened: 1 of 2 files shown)


FirefoxMetzger:
@Pandede Yes, we can add more test images. However, the condition is that the image is CC0 licensed or that we get written permission to redistribute the image for testing purposes.

Pandede:
Maybe I may use `requests` to download the dataset as temporary file for testing? This may avoid the license problem.

Pandede:
Thanks for the comprehensive explanation! I would like to contact the author and ask for the write permission of the dataset.

Pandede:
I have asked for the permission but it failed. I think this PR can be merged if you don't mind the missing of testcases.

FirefoxMetzger:
Great, iirc the test image was the last piece missing in this PR :) Thank you for adding it @Pandede !

PR #1070, opened 2024-03-20
Update pyav.py

cwood1967:
The documentation for init_video_stream suggests using "x264" but this results in in a no codec error. Replacing that with "libx264" works, so I think the documentation should be updated.


FirefoxMetzger:
That seems like a reasonable suggestion :) Thanks for fixing it 🚀

Issue #1073, opened 2024-04-04
How to add a pause between loops of a GIF?


Miguel-LlamasLanza:
I use `imageio.mimsave("fname.gif", frames, loop=4)` to save a GIF from python plots. I would like to add a pause between each loop. Is this possible?

Pandede:
Use optional argument `duration`, it accepts `float` for all frames or `list` for one value of each frame.
Pausing the GIF between loops can be done via this tricky method:
```python
import imageio.v2 as iio

frames = iio.mimread("imageio:newtonscradle.gif")
# The default duration of each frame
default_duration = 50
# The lagging
lag = 1000
# Pause the GIF by extending the duration of the last frame
duration = [default_duration] * (len(frames)-1) + [lag]
iio.mimwrite('paused.gif', frames, loop=4, duration=duration)
```
![paused](https://github.com/imageio/imageio/assets/38829000/2d86a6e5-833b-41f4-8918-2e7c27f4692f)

Note that, you may find out the frame duration with `imageio.v3.immeta`:
```python
import imageio.v3 as iio
meta = iio.immeta("imageio:newtonscradle.gif")
print(meta['duration'])  # 50
```

FirefoxMetzger:
I am closing this as answered - thank you again @Pandede 🚀

PR #1080, opened 2024-05-04
Fix wrong time unit of the argument `duration` in documentation

### LINKED ISSUE #1074: Wrong time unit in documentation of pillow_legacy
Hi,

I noticed that in the [documentation of the `pillow_legacy` plugin](https://imageio.readthedocs.io/en/v2.14.0/reference/_backends/imageio.plugins.pillow_legacy.html#parameters-for-saving), the unit for the duration is written to be seconds. However, after looking through the code, I noticed that this parameter is passed directly to PIL, which [expects milliseconds](https://pillow.readthedocs.io/en/stable/handbook/image-file-formats.html#gif). 

I believe the documentation should be updated.

Currently, this issue [affects MoviePy](https://github.com/Zulko/moviepy/issues/2151).

Best,
Tim

Pandede:
Fixes #1074.
The duration is actually in unit of milliseconds (here is the [proof](https://github.com/imageio/imageio/issues/1073)).

### DIFF
--- a/imageio/plugins/pillow_legacy.py
+++ b/imageio/plugins/pillow_legacy.py
@@ -137,10 +137,8 @@ loop : int
     The number of iterations. Default 0 (meaning loop indefinitely).
 duration : {float, list}
     (Only available in GIF-PIL)
-    The duration (in seconds) of each frame. Either specify one value
+    The duration (in milliseconds) of each frame. Either specify one value
     that is used for all frames, or one value for each frame.
-    Note that in the GIF format the duration/delay is expressed in
-    hundredths of a second, which limits the precision of the duration.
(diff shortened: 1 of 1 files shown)


FirefoxMetzger (approved):
Thanks @Pandede ! LGTM

Regarding the comment on the GIF side that you've deleted: I agree that we should remove the sentence here because it is confusing. 

That said, internally GIF can't represent ms-level durations and is capped at 100 FPS (100ms duration). Most readers, however, only support up to 50 FPS and setting an unsupported duration makes them fall back to a default duration. I.e., if you find that your GIF animates too slow chances are you set an incompatible duration.

PR #1081, opened 2024-05-04
Support using offline test image cache w/ `-m "not needs_internet"`

### LINKED ISSUE #1078: Please make it possible to run tests against `test_images` without Internet access
A fair number of tests currently require data from the `test_images` repository. It is currently technically possible to download the contents of the repository as a git snapshot and unpack it into `.test_images` directory but `conftest.py` unconditionally invokes `git pull`, which breaks that.

Furthermore, the `test_images` fixture currently implies `needs_internet` marker, which makes it impossible to distinguish between the tests that actually access Internet at testing time, and these that just need the `.test_images`.

Could you, please, make it possible to use a prefetched test image cache without using `git` at all, and to deselect just the tests that need Internet? I can make a pull request if you tell me which approach you'd like to employ for this.

mgorny:
Disable fetching/updating `.test_images` cache when `-m "not needs_internet"` is requested.  Instead, either reuse the existing cache if one is available, or skip the relevant tests.

This preserves the current behavior for the two scenarios: when `-m "not needs_internet"` is not passed, the cache is normally cloned and updated.  If it is passed and cache does not exist, the tests are skipped as before.  However, if one populates the cache first (either via running the tests or manually cloning it) and then passes `-m "not needs_internet"`, the tests run against the existing cache.

Fixes #1078

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -70,7 +70,7 @@ def working_directory(path):
 
 
 @pytest.fixture(scope="session")
-def test_images(request):
+def test_images(request, pytestconfig):
     """A collection of test images.
 
     Note: The images are cached persistently (across test runs) in
@@ -81,6 +81,7 @@ def test_images(request):
     # downloaded once in the beginning
 
     checkout_dir = request.config.cache.get("imageio_test_binaries", None)
+    use_internet = "not needs_internet" not in pytestconfig.getoption("-m")
 
     if checkout_dir is not None and not Path(checkout_dir).exists():
         # cache value set, but cache is gone :( ... recreate
@@ -90,6 +91,8 @@ def test_images(request):
         # download test images and other binaries
         checkout_dir = Path(__file__).parents[1] / ".test_images"
 
+        if not checkout_dir.exists() and not use_internet:
+            pytest.skip("Internet use disabled and `.test_images` not found")
         try:
             checkout_dir.mkdir()
         except FileExistsError:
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -90,6 +91,8 @@ def test_images(request):
         # download test images and other binaries
         checkout_dir = Path(__file__).parents[1] / ".test_images"
 
+        if not checkout_dir.exists() and not use_internet:
+            pytest.skip("Internet use disabled and `.test_images` not found")
```
FirefoxMetzger, line 95:
```suggestion
            pytest.skip("Internet use disabled and `.test_images` not found.")
```

Nit - but I like sentences ending with a period.

FirefoxMetzger (changes requested):
Pretty much good to go - thanks for the PR 🚀

I only have one nitpick ... I also merged a fix for the failing tests earlier - now the tests run correctly again. If they are happy and we add the `.` then I am happy, too :)

mgorny:
Done. Do you want me to rebase and squash the change?

FirefoxMetzger:
No that's fine. I will squash-merge :)

FirefoxMetzger (approved):
LGTM once tests are happy. 

GHA has a flaky pypy runner right now. One of the tests sporadically fails while setting up, so it is unrelated to this PR.

mgorny:
Thanks!

For the record, I don't think you need to test against pypy-3.8 anymore. Upstream is no longer maintaining that branch, and new releases include 2.7, 3.9 and 3.10 only.

FirefoxMetzger:
> For the record, I don't think you need to test against pypy-3.8 anymore. Upstream is no longer maintaining that branch, and new releases include 2.7, 3.9 and 3.10 only.

Thanks for the heads up :) I have also noticed this when adding tests for `pypy-3.10` recently. I was thinking of dropping tests for it alongside the CPython 3.8 deprecation this October.

PR #1085, opened 2024-05-20
DEV: Update actions/checkout and actions/setup-python

boidolr:
The current versions will run on soon to be unsupported versions of Node
(see https://github.blog/changelog/2024-05-17-updated-dates-for-actions-runner-using-node20-instead-of-node16-by-default/).

### DIFF
--- a/.github/workflows/build_docs.yml
+++ b/.github/workflows/build_docs.yml
@@ -8,8 +8,8 @@ jobs:
     name: Build Docs
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v3
-    - uses: actions/setup-python@v4
+    - uses: actions/checkout@v4
+    - uses: actions/setup-python@v5
       with:
         python-version: '3.x'
     - name: Install Dependencies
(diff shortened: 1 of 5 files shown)


FirefoxMetzger (approved):
This looks like a very good thing to do. Thanks for the PR 🚀

boidolr:
Not sure why the hash for requirements on macOS fails - I don't think it is related to the update of the workflow actions though, especially considering other platforms working without issues. Let me know what I can do to resolve this, if at all.

FirefoxMetzger:
@boidolr no it's not related. It's because Pypi 3.8 is phasing out ... I was hoping to remove it from our tests alongside the tests for cpython 3.8 (which will reach EoL after summer). However, this is breaking quite regularly now, so I will just remove it now instead.

FirefoxMetzger:
Fixed one error, found the next. I also don't think this one is related to your PR. It looks like our ffmpeg plugin is not loading the ffmpeg binary correctly anymore - i need to investigate.

Issue #1099, opened 2024-09-03
Bug with PyAV plugin


johnathanchiu:
Running the snippet below results in the following error. I am currently using version `imageio==2.34.2`. 

Error:
```
Traceback (most recent call last):
  File "example.py", line 6, in <module>
    for i, im in enumerate(reader):
  File "~/anaconda3/envs/yt-clipper/lib/python3.11/site-packages/imageio/v2.py", line 166, in iter_data
    metadata = self.instance.metadata(index=idx, exclude_applied=False)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "~/anaconda3/envs/yt-clipper/lib/python3.11/site-packages/imageio/plugins/pyav.py", line 771, in metadata
    {item.type.name: item.to_bytes() for item in desired_frame.side_data}
  File "~/anaconda3/envs/yt-clipper/lib/python3.11/site-packages/imageio/plugins/pyav.py", line 771, in <dictcomp>
    {item.type.name: item.to_bytes() for item in desired_frame.side_data}
                     ^^^^^^^^^^^^^
AttributeError: 'av.sidedata.sidedata.SideData' object has no attribute 'to_bytes'
```

Snippet:
```
import imageio as iio

video_path = 'test_video_file.ts'
reader = iio.get_reader(video_path)
for i, im in enumerate(reader):
    break
```

hmaarrfk:
can likely be closed right

FirefoxMetzger:
Yes, good find @hmaarrfk !

PR #1100, opened 2024-09-03
Use `bytes` instead of `.to_bytes` in PyAV plugin

johnathanchiu:
This PR resolves issue https://github.com/imageio/imageio/issues/1099. `av.sidedata.sidedata.SideData` doesn't have a `to_bytes` method. This results in the error being thrown `AttributeError: 'av.sidedata.sidedata.SideData' object has no attribute 'to_bytes'`. Changing to just `bytes` instead resolves this issue.


FirefoxMetzger:
Thanks for the PR @johnathanchiu ! 

I wonder if this is related to an interface change in PyAV where they deprecated `to_bytes` but it existed in an earlier version. I'll have a look later and if needed add logic to use the right method depending on the installed PyAV version.

PR #1101, opened 2024-10-01
Fix test_video_format_to_dtype for d3d12

K900:
New video format added in recent ffmpeg releases, should be handled the same way as other hardware accelerated ones.

### DIFF
--- a/tests/test_pyav.py
+++ b/tests/test_pyav.py
@@ -144,6 +144,7 @@ def test_video_format_to_dtype():
         "vaapi_vld",
         "d3d11",
         "d3d11va_vld",
+        "d3d12",
         "videotoolbox_vld",
         "vaapi",
         "vaapi_idct",


FirefoxMetzger (approved):
Open-and-shut case - LGTM

Thanks for the contribution! 🚀

K900:
The test failure seems unrelated.

K900:
So uh, how do I proceed here?

K900:
I'm sorry if I'm being annoying, but can I get a merge here?

FirefoxMetzger:
Sorry for being slow to respond. I was traveling for work this week. This tends to include "social activities" after work hours, so I end up with little time to spend on other stuff.

I could merge, but it will - unfortunately - just break again during CD later. The problem is that we need to deprecate python 3.8 support and tests. I was planning it for this weekend so we should be able to merge after that.

K900:
Understood, thanks! We'll just vendor the patch in nixpkgs for now, then.

FirefoxMetzger:
I encountered this one while removing 3.8 support. 😅 

I will merge and then rebase my PR on yours. Thanks for the contribution :)

PR #1108, opened 2024-11-08
Fix `ffmpeg` can only write 8-bits image/video

### LINKED ISSUE #1107: Cannot Write 16-Bit Monochrome Lossless Videos
Hi,

I'm trying to run a simple ChatGPT code example that will save 16-bit monochrome videos with lossless encoding, but every time I try to do so, I get a warning message `Lossy conversion from uint16 to uint8. Losing 8 bits of resolution. Convert image to uint8 prior to saving to suppress this warning.`

I don't know how to get around this issue, but I'm pretty sure FFV1 supports 16-bit encoding/decoding when I ran imageio's ffmpeg executable using the `-codecs` flag. I'm using python 3.12.0 with imageio 2.34.0. I installed it with `pip install imageio[ffmpeg]`.

I just want to write 16-bit monochrome lossless videos for data storage purposes. I'm intentionally moving away from splitting the 16-bit frames into an 8-bit RGB frames.

```python
import numpy as np
import imageio

# Define video parameters
width, height, fps = 640, 480, 30
num_frames = fps*2

# Create a video writer object with the desired parameters
output_file = "output_video.avi"

with imageio.get_writer(
    output_file,
    format='FFmpeg',
    fps=fps,
    codec='ffv1',  # Use FFV1 for lossless compression
    ffmpeg_params=['-pix_fmt', 'gray16le']  # Set pixel format to 16-bit grayscale
) as writer:
    for _ in range(num_frames):
        # Generate a random 16-bit frame
        frame = np.random.randint(0, 65536, (height, width), dtype=np.uint16)
        
        # Append the frame to the video writer
        writer.append_data(frame)

print(f"Video saved as {output_file}")
```

Pandede:
Resolves #1107.
The plugin `ffmpeg` can only write 8-bits image/video because of the fixed bitdepth conversion, relax the condition may solve the problem.

### DIFF
--- a/imageio/plugins/ffmpeg.py
+++ b/imageio/plugins/ffmpeg.py
@@ -566,8 +566,17 @@ class FfmpegFormat(Format):
             size = w, h
             depth = 1 if im.ndim == 2 else im.shape[2]
 
-            # Ensure that image is in uint8
-            im = image_as_uint(im, bitdepth=8)
+            # Resolve issue #1107
+            # Determine `bitdepth` of the image
+            if im.dtype == np.uint8:
+                bitdepth = 8
+            elif im.dtype == np.uint16:
+                bitdepth = 16
+            else:
+                # `image_as_uint` returns uint8 image if `bitdepth` is not provided
+                bitdepth = None
+
+            im = image_as_uint(im, bitdepth=bitdepth)
             # To be written efficiently, ie. without creating an immutable
             # buffer, by calling im.tobytes() the array must be contiguous.
             if not im.flags.c_contiguous:


Pandede:
Please don't merge it before the regression test is pushed.

Pandede:
I cannot do the regression test because the plugin `ffmpeg` seems to be messed up on `pix_fmt` and `dtype` when reading the image/video with `dtype=uint16`.
I may suggest the user to use plugin `pyav` to do the same thing instead.

FirefoxMetzger:
Yes, the goal is to eventually phase out the old FFMPEG plugin (it's written for the v2 API) and fully switch to PyAV instead.

I need to look into where we are in terms of feature parity again, but if memory serves the only reason FFMPEG is the default plugin instead of pyav is because pyav was too new at the time and we wanted to have a longer period where we test it before switching defaults.

PR #1109, opened 2024-11-11
Fix minimum Python version in docs

dstansby:
I noticed this was out of date, so here's a fix! Alternatively, I could get rid of the specific version completely to avoid this issue in the future?


FirefoxMetzger:
Oh, nice find. I have totally missed this one when upgrading the past couple python versions :)

PR #1112, opened 2024-12-04
Avoid pyav v14 to prevent AttributeError

### LINKED ISSUE #1111: latest pyav `av.video.codeccontext.VideoCodecContext` does not have a `close()` method
The new release of `pyav` v14.0.0 just a few hours ago removes the `CodecContext.close()` method, which prevents `iio.imread()` from being used for reading video files.

```python
examples/image_widget/image_widget_videos.py:17: in <module>
    cockatoo = iio.imread("imageio:cockatoo.mp4")
/opt/hostedtoolcache/Python/3.12.7/x64/lib/python3.12/site-packages/imageio/v3.py:53: in imread
    with imopen(uri, "r", **plugin_kwargs) as img_file:
/opt/hostedtoolcache/Python/3.12.7/x64/lib/python3.12/site-packages/imageio/core/v3_plugin_api.py:367: in __exit__
    self.close()
/opt/hostedtoolcache/Python/3.12.7/x64/lib/python3.12/site-packages/imageio/plugins/pyav.py:785: in close
    self._video_stream.close()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   AttributeError: 'av.video.codeccontext.VideoCodecContext' object has no attribute 'close'
```

shaform:
Closes: #1111.

Work around the issue described in #1111 by restricting pyav version.

### DIFF
--- a/imageio/plugins/pyav.py
+++ b/imageio/plugins/pyav.py
@@ -272,8 +272,13 @@ def _get_frame_type(picture_type: int) -> str:
         A human readable name of the picture type
 
     """
+
+    if not isinstance(picture_type, int):
+        # old pyAV versions send an enum, not an int
+        return picture_type.name
+
     picture_types = [
-        "None",
+        "NONE",
         "I",
         "P",
         "B",
--- a/imageio/plugins/pyav.py
+++ b/imageio/plugins/pyav.py
@@ -458,7 +458,10 @@ class PyAVPlugin(PluginV3):
 
             # reset stream container, because threading model can't change after
             # first access
-            self._video_stream.close()
+            try:
+                self._video_stream.close()
+            except AttributeError:
+                pass  # close() was removed in newer versions
             self._video_stream = self._container.streams.video[0]
 
             return frames
@@ -785,6 +788,8 @@ class PyAVPlugin(PluginV3):
                 self._video_stream.close()
             except ValueError:
                 pass  # stream already closed
+            except AttributeError:
+                pass  # close() was removed in newer versions
 
         if self._container is not None:
             self._container.close()
--- a/imageio/plugins/pyav.py
+++ b/imageio/plugins/pyav.py
@@ -495,12 +495,9 @@ class PyAVPlugin(PluginV3):
 
             return frames
 
-        if (
-            thread_type is not None
-            and not (
-                self._video_stream.thread_type == thread_type
-                or self._video_stream.thread_type.name == thread_type  
-            )
+        if thread_type is not None and not (
+            self._video_stream.thread_type == thread_type
+            or self._video_stream.thread_type.name == thread_type
         ):
             self._video_stream.thread_type = thread_type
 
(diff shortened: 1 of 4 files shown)


imageio/plugins/pyav.py
```diff
@@ -785,6 +788,8 @@ def close(self) -> None:
                 self._video_stream.close()
             except ValueError:
                 pass  # stream already closed
+            except AttributeError:
+                pass  # close() was removed in newer versions
```
FirefoxMetzger, line 792: If `CodecContext` no longer has a `.close()` method, how is it disposed of or finalized once we are done writing?

shaform: That's a good question. It seems the resource would be de-allocated when `__dealloc__` is called:
https://github.com/PyAV-Org/PyAV/blob/e0f0ef415b0ec7dfd48e582cf1735eb3bf04829c/av/codec/context.pyx#L227-L228

More reference: https://patchwork.ffmpeg.org/project/ffmpeg/patch/20240201082938.16687-4-anton@khirnov.net/

kushalkolar: They say to use `del` https://github.com/PyAV-Org/PyAV/pull/1592

Not sure whats the best way to implement that here to account for all the different ways `_video_stream` can be used by `PyAVPlugin`.

shaform: I see. Indeed if we want to account for all the different usage, it would require more thoughts. Perhaps we could instead restrict the version of pyav that imageio supports for now to fix the current issue, so that people will have more time to work on a more general solution? If so I could modify the PR to just restrict the version of pyav. Thanks!

FirefoxMetzger: We can't merge a PR that pins pyAV below some version-major without also working on a PR that removes the constraint. Pinning is more of a hotfix.

I looked at the changes and, as it turns out, we can simply stop calling `.close` on the video stream without replacement. If it is used for reading it won't hold any resources (the file pointer is managed by the container) and the state of the codec is okay to lose once we no longer need it. If we use the stream for writing, we want to keep it around until we are done writing and finalize the stream, which is done when we close/finalize the container. All we need to do is make sure we don't hold any references to the streams anymore so that GC can deallocate them once the container gets deallocated.

FirefoxMetzger:
[Sorry for ghosting this PR. I've had a neat sequence of events that, unfortunately, kept me away from open-source (work deadline, vacation in China, being sick). Now I am back.

I took the liberty to not just go over the PR but also checked out other changes that come with av v14, which are mostly removals of various convenience methods that used to exist before. I added changes for these here.

FirefoxMetzger (approved):
I added all the changes I would make to this PR, and then I will merge once I#ve updated CI.

PR #1114, opened 2024-12-25
Support JPEG-XL image file

### LINKED ISSUE #1113: Support for jpeg-xl (jxl)
Is there plan to support jpeg-xl image format?

Pandede:
Resolves #1113.
Enable to read/write operation on JPEG-XL (jxl) image file by registering the Pillow plugin [pillow-jxl-plugin](https://pypi.org/project/pillow-jxl-plugin/).

### DIFF
--- a/tests/test_pillow.py
+++ b/tests/test_pillow.py
@@ -732,3 +732,18 @@ def test_webp_remote():
     url = "https://github.com/python-pillow/Pillow/raw/main/Tests/images/hopper_orientation_2.webp"
     im = iio.imread(url, plugin="pillow")
     assert im.shape == (128, 128, 3)
+
+
+@pytest.mark.needs_internet
+def test_jxl_read_remote():
+    url = "https://github.com/libjxl/conformance/raw/refs/heads/master/testcases/cafe/input.jxl"
+    im = iio.imread(url, plugin="pillow")
+    assert im.shape == (1600, 1280, 3)
+
+
+def test_jxl_write(test_images, tmp_path):
+    actual = iio.imread(test_images / "chelsea.png")
+    iio.imwrite(tmp_path / "chelsea.jxl", actual, lossless=True)
+
+    target = iio.imread(tmp_path / "chelsea.jxl")
+    assert np.allclose(actual, target)
--- a/imageio/config/extensions.py
+++ b/imageio/config/extensions.py
@@ -1865,6 +1865,11 @@ extension_list = [
         extension=".avif",
         priority=["pillow"],
     ),
+    FileExtension(
+        name="JPEG XL Image File Format",
+        extension=".jxl",
+        priority=["pillow"],
+    ),
 ]
 extension_list.sort(key=lambda x: x.extension)
 
--- a/imageio/plugins/pillow.py
+++ b/imageio/plugins/pillow.py
@@ -96,6 +96,12 @@ class PillowPlugin(PluginV3):
         else:
             register_avif_opener()
 
+        # Initialize JXL plugin for Pillow
+        try:
+            import pillow_jxl as _
+        except ImportError:
+            pass
+
         self._image: Image = None
         self.images_to_write = []
 
--- a/setup.py
+++ b/setup.py
@@ -100,6 +100,7 @@ plugins = {
     "spe": [],
     "swf": [],
     "tifffile": ["tifffile"],
+    "pillow-jxl": ["pillow-jxl-plugin"],
 }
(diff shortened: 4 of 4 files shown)


tests/test_pillow.py
```diff
@@ -732,3 +732,18 @@ def test_webp_remote():
     url = "https://github.com/python-pillow/Pillow/raw/main/Tests/images/hopper_orientation_2.webp"
     im = iio.imread(url, plugin="pillow")
     assert im.shape == (128, 128, 3)
+
+
+@pytest.mark.needs_internet
+def test_jxl_read_remote():
+    url = "https://github.com/libjxl/conformance/raw/refs/heads/master/testcases/cafe/input.jxl"
```
FirefoxMetzger, line 739: I checked the repo and we could use this image for testing. 

However, I still want it to be part of our test images. The reason is that a URL will not be cached locally, so whenever anyone re-runs the suite (or even just this test) that person will re-download the image. Let's not slow down our tests unnecessarily.

FirefoxMetzger: On second thought, we might not be allowed to simply copy the image into our test suite. It says that the ITU retains the copyright for this image and that it allows everybody to share and use it for testing purposes:
> By the present licence, ITU grants users of the software a non-exclusive right
to freely reproduce the information contained therein, so that it may be used
for intended purposes only. ([source](https://github.com/libjxl/conformance/blob/master/testcases/cafe/itu-t.24_ReadMe.txt
))

The problem with the above is that images in our test suite can be used for any purpose, not just for testing. So to include the image in our suite we'll once again need an OK from them that we may distribute it for any purpose.

Pandede: To address the concerns regarding copyright and distribution, I propose an alternative solution: we can create our own JPEG-XL test image by converting one of our existing test images into JPEG-XL format. This way, we avoid any potential licensing issues while ensuring the test suite remains self-contained and efficient.

FirefoxMetzger: Yes, that would work perfectly fine :)

Pandede: Thank you! Here is the PR to imageio/test_images: https://github.com/imageio/test_images/pull/4.
I’ll proceed with the implementation once this is merged.

FirefoxMetzger: merged :)

FirefoxMetzger (commented):
Looks like we need to install the JPEG-XL dependency in CI so that we can load the plugin and open the file. I.e., Modify our `cpython_tests.yml`.

Is that something you know how to do (or want to learn), @Pandede, or should I fix that?

Pandede:
Thank you for your guidance. I know how to modify the CI/CD pipeline to install the JPEG-XL dependency. However, I'm currently unable to determine why the CI/CD is failing despite the pillow-jxl-plugin being installed successfully according to the logs. Could you please assist me in identifying the root cause of this issue?

Pandede:
@FirefoxMetzger I've contacted with the author of `pillow-jxl-plugin` and it's fixed now.

Pandede:
@FirefoxMetzger This PR is ready to be reviewed.

.github/workflows/cpython_tests.yml
```diff
@@ -30,6 +30,15 @@ jobs:
...
           fetch-depth: 2
+      - name: Shorten temp & cargo dirs on Windows
+        if: matrix.os == 'windows-latest'
+        shell: pwsh
+        run: |
+            New-Item -ItemType Directory -Force -Path C:\\tmp
+            New-Item -ItemType Directory -Force -Path C:\\tmp\\cargo-target
+            Add-Content -Path $env:GITHUB_ENV -Value "TMP=C:\\tmp"
+            Add-Content -Path $env:GITHUB_ENV -Value "TEMP=C:\\tmp"
+            Add-Content -Path $env:GITHUB_ENV -Value "CARGO_TARGET_DIR=C:\\tmp\\cargo-target"
```
FirefoxMetzger, line 41: This step looks a bit odd. Cargo is a build system for Rust ... how come this shows up here?

pyproject.toml
```diff
@@ -84,6 +84,7 @@ simpleitk = []
 spe = []
 swf = []
 tifffile = ["tifffile"]
+pillow-jxl = ["pillow-jxl-plugin>=1.3.2"]
```
FirefoxMetzger, line 87: What feature are we using that requires us to pin the version to `>=1.3.2`?

PR #1116, opened 2025-01-16
bugfix arr.newbyteorder deprecation in numpy>2.0

jasonkena:
Error addressed:
image
Unfortunately, I couldn't manage to create a test case which covers this behavior, because it's hard to trigger https://github.com/jasonkena/imageio/blob/098bf26280d3ef513f21f20fb7bdf1693c642ae5/imageio/plugins/_tifffile.py#L5074


FirefoxMetzger:
Looks good; thanks for the contribution :)

PR #1117, opened 2025-01-18
TEST: remove 3.9 tests and test on Windows

FirefoxMetzger:
This PR removes pypy3.9 from the test suite. It is no longer officially supported and pillow has stopped releasing prebuilt binaries for it.

The PR also re-introduces windows as an OS for which we run pypy tests. I remember that I removed it because a dependency was not supported there, but I will try to reintroduce it because time has passed. If it doesn't work we will keep running without.

### DIFF
--- a/.github/workflows/pypy_tests.yml
+++ b/.github/workflows/pypy_tests.yml
@@ -10,8 +10,8 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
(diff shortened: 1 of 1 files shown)


FirefoxMetzger:
Closing, because this change had to be done as part of the previous PR (circular "dependency" where the CI was broken by two things and both need to be fixed at once to ensure that we get everything to pass)

PR #1121, opened 2025-02-19
Fix `OverflowError` may be raised when `init_video_stream` with FPS in float

### LINKED ISSUE #1120: [PyAV Plugin] Setting Float `fps` Causes `OverflowError`
### Description

Currently, the `init_video_stream` method in the PyAV plugin converts float `fps` to a `Fraction`. However, for some common fps numbers, this results in an `OverflowError: value too large to convert to int`.

### Steps to Reproduce

The full code to reproduce the issue is as follows:
```python
writer = iio.imopen("<bytes>", 'w', extension='.mp4', plugin='pyav')
writer.init_video_stream('h264', fps=29.97)
```

```
---------------------------------------------------------------------------
OverflowError                             Traceback (most recent call last)
Cell In[36], line 1
----> 1 writer.init_video_stream('h264', fps=29.97)

File ~/miniconda3/envs/test/lib/python3.9/site-packages/imageio/plugins/pyav.py:886, in PyAVPlugin.init_video_stream(self, codec, fps, pixel_format, max_keyframe_interval, force_keyframes)
    850 """Initialize a new video stream.
    851 
    852 This function adds a new video stream to the ImageResource using the
   (...)
    882 
    883 """
    885 fps = Fraction.from_float(fps)
--> 886 stream = self._container.add_stream(codec, fps)
    887 stream.time_base = Fraction(1 / fps).limit_denominator(int(2**16 - 1))
    888 if pixel_format is not None:

File av/container/output.pyx:93, in av.container.output.OutputContainer.add_stream()

File av/utils.pyx:51, in av.utils.to_avrational()

OverflowError: value too large to convert to int
```

### Cause

This issue arises because float numbers are converted to a `Fraction` with very large numerator and denominator. For example, `Fraction.from_float(29.97)` returns `Fraction(1054475631502295, 35184372088832)`.

### Impact

This is difficult for me to work around since I am trying to maintain the original fps of the input file, as in 
```python
fps = reader.metadata()["fps"]
```

Pandede:
Resolves #1120.
This PR addresses a [long-standing issue](https://github.com/PyAV-Org/PyAV/issues/242) in `pyav` that has existed since version `0.4.0`.


FirefoxMetzger (approved):
Looking good :) Thanks for taking the time to fix this.

ameraner:
Anything blocking this PR? It is currently causing unavoidable failures in our dowstream application, would be great to have the fix! Thanks.

PR #1129, opened 2025-03-17
Use Packaging

### LINKED ISSUE #1128: Pillow Version Parsing
The current method used to parse the version of the pillow package cannot properly handle all versions specified by [PEP 440](https://peps.python.org/pep-0440/). Use the version handling mechanisms of [packaging](https://packaging.pypa.io/en/stable/version.html) or if you prefer to not add an additional dependency, remove support for legacy pillow versions.

dcb-imvaria:
Resolves #1128

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,8 +55,8 @@ jobs:
           python-version: '3.12'
       - name: Install Dependencies
         run: |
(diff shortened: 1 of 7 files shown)


FirefoxMetzger (approved):
Looking good. Thanks for the contribution :)

PR #1142, opened 2025-07-23
tests: mark more tests as needing the internet

philiptaron:
These tests use the internet more than others that accept `test_images`, since they use the `imageio:` syntax or a specific URL in the case of `test_uri_reading`.


philiptaron:
No worries! Thank you for the merge.

PR #1149, opened 2025-10-05
Avoid calling PIL.Image twice to open a file

hmaarrfk:
Sometimes reading metadata is just slow.... lets avoid it...

### DIFF
--- a/.github/workflows/pypy_tests.yml
+++ b/.github/workflows/pypy_tests.yml
@@ -5,7 +5,7 @@ on:
 
 jobs:
   pypy:
-    name: "${{ matrix.os }} / pypy-3.10"
+    name: "${{ matrix.os }} / pypy-3.11"
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: true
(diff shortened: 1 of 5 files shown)


imageio/plugins/pillow.py
```diff
@@ -101,12 +101,12 @@ def __init__(self, request: Request) -> None:
...
-                    # Check if it is generally possible to read the image.
-                    # This will not read any data and merely try to find a
-                    # compatible pillow plugin (ref: the pillow docs).
-                    pass
+                # Check if it is generally possible to read the image.
+                # This will not read any data and merely try to find a
+                # compatible pillow plugin (ref: the pillow docs).
+                image = Image.open(request.get_file())
             except UnidentifiedImageError:
+                image.close()
```
hmaarrfk, line 109:
```suggestion
```

hmaarrfk:
seems like something is off in the windows ffmpeg build?

FirefoxMetzger:
Okay, I got the logic now. Pillow's `Image.open` actually doesn't close the file if you pass a file handle and, as a result, we don't get two OS calls to open the file twice. We simply pass around file handles.

It's safe to merge the two calls into one, but we won't get a big performance boost out of it I fear.

hmaarrfk:
not sure i agree.

THe old way

1. Open call when entering the context manager (read metadata and whatnot)
2. CLose call when exiting the context manager
3. Open call again (read metadata again and whatnot)


Alot of time is "wasted' reading metadata from files. as that metadata grows, it becomes an increased cost. 

Its just a tiny optimization, but also improves readability IMO.

Thx for merging.

Issue #624, opened 2021-04-09
prefer_uint8=False upcast uint8 to int16


Conchylicultor:
From: https://imageio.readthedocs.io/en/stable/format_png-pil.html

> prefer_uint8: bool : Let the PNG writer truncate uint16 image arrays to uint8 if their values fall within the range [0, 255]. Defaults to true for legacy compatibility, however it is recommended to set this to false to avoid unexpected behavior when saving e.g. weakly saturated images.

I would expect that `prefer_uint8=False` do not modify the dtype of the image. So save `np.uint8` images as `np.uint8`, and   save `np.uint16` images as `np.uint16`. However, setting this param save all images as `np.uint16`:

Reproductible example:

```python
import imageio
import numpy as np

print(imageio.__version__)  # 2.9.0

img_path = 'x.png'
arr = np.array([[0, 1, 2], [0, 255, 200]], dtype=np.uint8)
print(arr)

imageio.imwrite(img_path, arr, prefer_uint8=False)
new_arr = imageio.imread(img_path)
print(new_arr)
```
Output:
```
[[  0   1   2]
 [  0 255 200]]
Lossy conversion from uint8 to uint16. Range [0, 255]. Convert image to uint16 prior to saving to suppress this warning.
[[    0   257   514]
 [    0 65535 51400]]
```
I find this behavior quite confusing.

FirefoxMetzger:
I saw this issue yesterday and gave it a day of thought. I'm actually not sure how to address it, i.e., if it would be a good idea to change the behavior, or if the keyword argument should be renamed. Mainly because I don't know which path will be more backward compatible.

@almarklein Do you have a good solution here?

almarklein:
I think the solution depends on whether we consider this behavior a bug. I would argue that it is. If you agree, we can simply change the behavior to stick to the dtype of the data (or closest to it).

FirefoxMetzger:
Sounds good. I guess the solution then is to update this function

https://github.com/imageio/imageio/blob/dc2d06b2358b6451164961b42a6d2f566fa5169e/imageio/plugins/pillow.py#L803

to handle `prefer_uint8=False` in a way that matches the expected behavior described by @Conchylicultor . This doesn't sound like a complicated change; I'll mark it as a good first issue.

@Conchylicultor would you be interested in submitting a PR for this?

PR #1186, opened 2026-05-13
Fix png dtype at function ndarray_to_pil

redhecker:
this pr is related to #624


fallenmi (approved):
Disclosure: I used OpenAI Codex to assist with this review and verified the result against exact commit `f0f306da0eed84b86dca0446c7a04a38f047a3dd`.

I reproduced the reported `prefer_uint8=False` path against base `c7f8249`: a uint8 PNG is written as `I;16` and byte values are scaled into uint16. On this head and current GitHub merge `03c9813`, uint8 remains mode `L` with exact values.

A deterministic 1,570-case matrix covered fixed, randomized, strided, and Fortran-order arrays, both PNG spellings, both preference values, and uint16 controls; base failed the 783 intended uint8/preference cases, while head and merge had zero failures and identical digests. The owning Pillow legacy test file passed on head and merge with 11 passed, 2 expected xfails, and 1 deselected; Black 26.3.0, Flake8, and diff checks also passed.

GitHub currently shows only a green Read the Docs status and no upstream code-test matrix. I found no blocking issue.

PR #1203, opened 2026-08-23
Make fuzz_read robust against short and invalid image inputs

xingyaner:
Update the fuzz target to use the `ImageIO v3 API`, skip inputs that are too short to represent a valid image, and catch all parsing exceptions raised by ImageIO and its plugins.

This prevents malformed fuzzing inputs from terminating the fuzz target, including `struct.error` exceptions raised while parsing incomplete image headers.


fallenmi (changes requested):
Please preserve the fuzzer's failure signal. On exact head `de90f303`, I injected an `iio.imread` sentinel that raises `AssertionError`; `TestOneInput(b"0123456789")` returned normally. The same unexpected exception escapes on exact base `3193352`, while the intended `ValueError` and `RuntimeError` cases remain filtered in both versions. Catching `Exception` here therefore hides invariant and plugin failures from Atheris—the failures this target needs to report. Please catch only the known invalid-input exceptions (or re-raise unexpected exceptions) and add a regression proving an unexpected sentinel escapes.

Reviewed with OpenAI Codex assistance.

xingyaner:
Thanks for the detailed feedback. I narrowed the exception handling to `ValueError`, `RuntimeError`, and `struct.error`, so unexpected exceptions such as `AssertionError` still escape to Atheris. I also added a regression test confirming that an unexpected sentinel exception is not suppressed. The updated fuzz target and regression test have been validated successfully.

